PR #13, opened 2021-06-15
Sys calls add calls 1

TristanB22:
Please check this code -- I was not able to test it very easily.


src/safeposix/dispatcher.rs
```diff
@@ -74,6 +74,8 @@ pub union Arg {
   ulong: u64,
   cstr: *const u8,
   cstrarr: *const *const u8,
+  void: *const c_void,
```
rennergade, line 77: @jesings we're going to have some different fields that are basically pointers, I'm not sure we need different types for all of them?

src/safeposix/syscalls/sys_constants.rs
```diff
@@ -3,4 +3,30 @@
...
+
+STACK_CUR : u64 = 8192*1024;
+STACK_MAX : u64 = 2**32;
+
+RLIMIT_STACK = 0;
+RLIMIT_NOFILE = 1;
+
+
+//R Limit for getrlimit system call
+#[repr(C)]
```
rennergade, line 28: I don't think we need to make this C typed?

src/safeposix/syscalls/sys_constants.rs
```diff
@@ -3,4 +3,30 @@
...
 
 // Define constants using static or const
-// Imported into fs_calls file
\ No newline at end of file
+// Imported into fs_calls file
+
+
+//GID AND UID DEFAULT VALUES
+
+DEFAULT_UID : u64 = 1000;
```
rennergade, line 11: I think these need to start with 'const'

rennergade (commented):
This is mostly exactly what we're looking for. There's a few things we still need to iron out as a group since the overall structure is a work in progress.

src/safeposix/syscalls/sys_constants.rs
jesings: We absolutely do because the struct is passed in from C, and we populate its fields in rust

src/safeposix/dispatcher.rs
jesings: I mean casting them is unsafe and verbose so I was thinking that it's less hassle to have a separate type for each

src/safeposix/syscalls/sys_constants.rs
rennergade: Ah ok, I was still in the mindset that we needed to serialize structs like we were doing from Python. I guess this will work much better/easier for structs!

src/safeposix/dispatcher.rs
rennergade: Ok, yeah I'm fine with that but we should probably eventually have this organized in a way thats not confusing if we start having a long list of arg types.

rennergade (approved):
This is a great start. Good work Tristan!

PR #18, opened 2021-06-24
Implemented read and write system calls

jesings:


src/interface/file.rs
```diff
@@ -49,6 +49,10 @@ pub fn removefile(filename: String) -> std::io::Result<()> {
     Ok(())
 }
 
+fn is_allowed_char(c: char) -> bool{
+    char::is_alphanumeric(c) || c == '.'
```
rennergade, line 53: This is a good change. I simplified it to just alphanumeric on my first iteration. Obviously we need periods for usual filenames. There's probably a few more characters we can allow. Not sure if its worth doing now.

src/interface/file.rs
```diff
@@ -175,6 +180,30 @@ impl EmulatedFile {
 
         Ok(bytes_written)
     }
+
+    pub fn zero(&mut self, count: usize, offset: usize) -> std::io::Result<usize> {
```
rennergade, line 184: Maybe a more descriptive name if it's not going to be commented - fill_with_zerosat or something.

src/interface/misc.rs
```diff
@@ -10,10 +10,14 @@ pub use std::collections::HashMap as RustHashMap;
 
 pub use std::sync::RwLock as RustLock;
 pub use std::sync::Arc as RustRfc;
-pub use std::thread::{current as rust_gettid, ThreadId as RustThreadId};
```
rennergade, line 13: was this removed because were not doing that with the dispatcher anymore?

src/safeposix/dispatcher.rs
```diff
@@ -75,6 +75,10 @@ pub union Arg {
   int: i32,
   uint: u32,
   ulong: u64,
+  usz: usize,
```
rennergade, line 78: maybe c-usize.

In general I think these args should be probably all of the format c-(typename) and maybe with comments after them saying which syscalls they're for

rennergade (commented):
Most of this is well commented, very readable and well implemented! 

It sounds like theres a few updates coming so Ill leave this review as is for now.

rennergade (approved):
Looks great!

PR #19, opened 2021-06-25
Extricated fileobjecttable

jesings:


rennergade (approved):
looks good! seems like this will make things generally easier as well

PR #22, opened 2021-06-28
Made tests not disgusting, added special file creator and mkdir

jesings:


src/tests/fs_tests.rs
```diff
@@ -0,0 +1,67 @@
+#[cfg(test)]
```
rennergade, line 1: Agree that moving these tests here makes sense

src/safeposix/filesystem.rs
```diff
@@ -70,21 +70,53 @@ pub struct FilesystemMetadata {
     pub inodetable: interface::RustHashMap<usize, Inode>,
 }
 
+pub fn init_filename_to_inode_dict(parentinode: usize) -> interface::RustHashMap<std::ffi::OsString, usize> {
+    let mut retval = interface::RustHashMap::new();
+    retval.insert(std::ffi::OsString::from("."), ROOTDIRECTORYINODE);
```
rennergade, line 75: this is a non-interface call in main safeposix code *wags finger*

src/safeposix/filesystem.rs
```diff
@@ -70,21 +70,53 @@ pub struct FilesystemMetadata {
...
+        let dirinode = DirectoryInode {size: 0, uid: DEFAULT_UID, gid: DEFAULT_GID,
+            mode: S_IFDIR as u32 | S_IRWXA, linkcount: 3, refcount: 0, 
+            atime: time, ctime: time, mtime: time,
+            filename_to_inode_dict: init_filename_to_inode_dict(ROOTDIRECTORYINODE)};
+        retval.inodetable.insert(ROOTDIRECTORYINODE, Inode::Dir(dirinode));
         retval
     }
 }
 
+pub fn load_fs_special_files() {
```
rennergade, line 94: Is this where we're setting up the initial cage now?

rennergade (commented):
Mostly good. Had a question about how we're going to do init, and flagged some non-interface calls.

src/safeposix/filesystem.rs
jesings: No I haven't really settled on how/if the initial cage is going to work

rennergade: OK this is fine for now, we'll probably move this around once I do my PR.

PR #26, opened 2021-06-30
Persistmetadata

rennergade:
Implements serialization:

Adds persist_metadata, restore_metadata, and load_fs


src/interface/file.rs
```diff
@@ -182,6 +178,51 @@ impl EmulatedFile {
         Ok(bytes_written)
     }
 
+    // Reads entire file into provided C-buffer
+    pub fn readfile_to_new_string(&self, offset: usize) -> std::io::Result<String> {
+
+        let mut buf = String::new();
```
jesings, line 184: If you want, putting this closer to where it's used (inside the Some match arm) will make it clearer what this is/does.
Also, change the name from buf, it's a string not a [u8]. stringbuf would be fine probably.

src/safeposix/filesystem.rs
```diff
@@ -87,15 +91,36 @@ impl FilesystemMetadata {
...
+        filedescriptortable: interface::RustLock::new(interface::RustHashMap::new())};
+
+    let metadatafo = interface::openfile(METADATAFILENAME.to_string(), true);
+    
+    // If the metadata file exists, just close the file for later restore
+    // If it doesn't, lets create a new one and persist it.
+    match metadatafo {
+        Ok(file) => {file.close().unwrap();}
+        Err(_e) => {FilesystemMetadata::blank_fs_init(); persist_metadata(None);}
+    };
```
jesings, line 114: I think this means metadata file object but maybe making it metadata_fileobj would make that clearer than fo.
Also, is the persist necessary right after the blank_fs_init? I thought we discussed it and said it wasn't.

src/safeposix/filesystem.rs
```diff
@@ -117,16 +142,44 @@ pub fn load_fs_special_files() {
...
+
+    // write to file
+    let mut metadatafo = interface::openfile(METADATAFILENAME.to_string(), true).unwrap();
+    metadatafo.writefile_from_string(metadatastring, 0).unwrap();
+    metadatafo.close().unwrap();
 }
 
-pub fn restore_metadata() {
+// Read file, and deserialize json to FS METADATA
+pub fn restore_metadata() -> FilesystemMetadata {
```
jesings, line 167: I think the assymetry in persist_metadata and restore_metadata bears explaining in a comment. One directly uses the FSMETADATA value and the other does not. Maybe restore_metadata isn't the right name anymore and something like reconstruct_metadata would make that clearer as well?

src/tests/fs_tests.rs
```diff
@@ -3,6 +3,18 @@ mod fs_tests {
     use crate::interface;
     use crate::safeposix::{cage::*, filesystem};
     use super::super::*;
+
+    #[test]
+    pub fn persistancetest() {
```
jesings, line 8: misspelling

src/tests/fs_tests.rs
```diff
@@ -3,6 +3,18 @@ mod fs_tests {
...
+        let cage = init_cage();
+
+        let fd = cage.open_syscall("/testfile", O_CREAT | O_EXCL | O_RDWR, S_IRWXA);
+        assert!(fd >= 0);
+
+        filesystem::persist_metadata();
+        filesystem::restore_metadata();
+    }
```
jesings, line 16: Maybe do the close and unlink and then the open/access to check if it's still there in your test to really test it's deserializing right?

jesings (commented):
Overall it's great progress and we're getting closer and closer to a usable state.
I had a few nitpicks though.

src/safeposix/filesystem.rs
rennergade: I think persist here does make sense, we basically init the metadata file and also give us something to do the "restore" from at the next step that is consistent whether or not we did a blank init.

src/tests/fs_tests.rs
rennergade: I do think we need more tests here but we I think we need more syscalls implemented to do that fully.

PR #27, opened 2021-06-30
Fixed some fcntl things

jesings:


TristanB22 (approved):
I think I understand -- thank you!

PR #28, opened 2021-06-30
Implemented linkcount for cwd in chdir/exit/fork to prevent corrupting state

jesings:


src/safeposix/filesystem.rs
```diff
@@ -87,7 +87,7 @@ impl FilesystemMetadata {
         let mut retval = FilesystemMetadata {nextinode: STREAMINODE + 1, dev_id: 20, inodetable: interface::RustHashMap::new()};
         let time = interface::timestamp(); //We do a real timestamp now
         let dirinode = DirectoryInode {size: 0, uid: DEFAULT_UID, gid: DEFAULT_GID,
-            mode: S_IFDIR as u32 | S_IRWXA, linkcount: 3, refcount: 0, 
+            mode: S_IFDIR as u32 | S_IRWXA, linkcount: 2, refcount: 0,
```
rennergade, line 90: Needs a comment as to why these numbers.  We probably shouldn't have numbers floating around. That probably goes for dev_id as 20 to which should probably be a const or commented.

rennergade (commented):
Just a minor comment. I think you were also going to make the mmap stuff not 'pub' here too.

rennergade (approved):

PR #32, opened 2021-07-02
adding close / fcntl / dup / dup2 / chmod 

TristanB22:
This is going to need some merge-handling so it plays nicely with main.


src/safeposix/cage.rs
```diff
@@ -49,7 +49,8 @@ pub struct PipeDesc {
     pub flags: i32
 }
 
-type FdTable = interface::RustHashMap<i32, interface::RustRfc<interface::RustLock<FileDescriptor>>>;
+pub type FdTable = interface::RustHashMap<i32, interface::RustRfc<interface::RustLock<FileDescriptor>>>;
```
rennergade, line 52: why does this need to be pub?

src/tests/ut_lind_fs_broken_close.rs
```diff
@@ -0,0 +1,10 @@
+
+#[cfg(tests)]
```
rennergade, line 2: not sure whats going on here

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                if 0 == operation & LOCK_SH {
+                    //TO DO: raise unimplemented error
+                    return syscall_error(Errno::ENOSYS, "flock", "this operation has not been implemented yet");
+                }
+                //check whether the lock is blocking or not
+                if operation & LOCK_EX == 0 && operation & LOCK_NB == 0 {
+                    //if lock.acquire(False) then return 0, else syscall error
+                    return 0;
+                } else if 0 == operation & LOCK_EX {
+                    //fd lock acquire(true)
```
rennergade, line 1314: still needs this it looks like

rennergade (commented):
Looks very good in general. Seems like you still need to do a few things, and you definitely have a pre-merge to do!

src/tests/ut_lind_fs_broken_close.rs
TristanB22: This was where I was going to start writing tests. It does not have to be implemented, though.

src/safeposix/cage.rs
TristanB22: I think that was an accident on my part -- probably was matching wrong

src/safeposix/syscalls/fs_calls.rs
TristanB22: right -- I am not sure how to implement the flag and don't want to screw it up.

jesings: Our idea is that you add a field to every filedesc type (i.e. File, Pipe, etc.) which contains a interface::RustLock<()>, and for ex lock it for writing, for sh lock it for reading.
For nb use try_read/try_write instead of read/write locking, and you'll need to deal with the result. of the try, returning an errno if the try result is an Err (maybe using an if_let).
If you have any further questions, feel free to ask!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+        };
+
+        //checking whether the fd exists in the file table and is higher than the starting file descriptor
+        if let Some(_) = fdtable.get(&fd) {
+            if fd >= STARTINGFD {
+                return syscall_error(Errno::EBADF, "dup", "provided file descriptor is out of range");
+            }
+
+            //error below may need to be changed -- called if error getting file descriptor
+            let nextfd = if let Some(fd) = self.get_next_fd(Some(start_fd), Some(&fdtable)) {fd} else {return syscall_error(Errno::ENFILE, "dup_syscall", "no available file descriptor number could be found");};
```
jesings, line 930: This line is a bit too long, putting the else arm on a different line would be fine

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+            return newfd;
+        }
+
+        //need to add close helper and reference
+        match fdtable.get(&newfd) {
+            Some(_) => {return Self::_close_helper(&self, newfd, Some(&fdtable), false);},
+            None => {
+                fdtable.insert(newfd, fdtable.get(&oldfd).unwrap().clone());
+            } 
+        }
```
jesings, line 970: The logic here seems wrong to me, even if you have to close a file descriptor previously present at the new fd, you should still later insert the duplicated fd.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                            }
+                        },
+                        Inode::Pipe(_) | Inode::Socket(_) => {panic!("How did you get by the first filter?");},
+                    }
+                },
+            }
+            // if this is a part of the close syscall, then the fd table instance of the fd is removed
+            if is_close_syscall {
+                self.filedescriptortable.write().unwrap().remove(&fd);
+            }
```
jesings, line 1098: The file descriptor should be removed whether it is a stream or socket or pipe as well, put it outside the if let

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                            //if there are still references to the file, then do nothing to the file
+                            if normalfile_inode_obj.refcount != 0 {
+                                return 0;
+                            }
+                            if fobjtable.contains_key(&inodenum) {
+                                //close and delete the file in the object table
+                                fobjtable.get(&inodenum).unwrap().close().unwrap();
+                                fobjtable.remove(&inodenum);
+                            }
+                            if normalfile_inode_obj.refcount == 0 {
```
jesings, line 1037: refcount by this point is already assumed to be 0, you have returned if it was non-zero, this block is unnecessary

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+
+                            //if there are still references to the file, then do nothing to the file
+                            if dir_inode_obj.refcount != 0 {
+                                return 0;
+                            }
+                            if fobjtable.contains_key(&inodenum) {
+                                //close and delete the file in the object table
+                                fobjtable.get(&inodenum).unwrap().close().unwrap();
+                                fobjtable.remove(&inodenum);
+                            }
```
jesings, line 1059: Directories don't have associated file objects

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+
+                            //if there are still references to the file, then do nothing to the file
+                            if normalfile_inode_obj.refcount != 0 {
+                                return 0;
+                            }
+                            if fobjtable.contains_key(&inodenum) {
+                                //close and delete the file in the object table
+                                fobjtable.get(&inodenum).unwrap().close().unwrap();
+                                fobjtable.remove(&inodenum);
+                            }
```
jesings, line 1036: I think every file is supposed to have an entry in the fdtable, so the contains_key check is not necessary

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                            //if there are still references to the file, then do nothing to the file
+                            if dir_inode_obj.refcount != 0 {
+                                return 0;
+                            }
+                            if fobjtable.contains_key(&inodenum) {
+                                //close and delete the file in the object table
+                                fobjtable.get(&inodenum).unwrap().close().unwrap();
+                                fobjtable.remove(&inodenum);
+                            }
+                            if dir_inode_obj.refcount == 0 {
```
jesings, line 1060: refcount is assumed to be 0 by this point too

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                            //if there are still references to the file, then do nothing to the file
+                            if char_inode_obj.refcount != 0 {
+                                return 0;
+                            }
+
+                            if fobjtable.contains_key(&inodenum) {
+                                //close and delete the file in the object table
+                                fobjtable.get(&inodenum).unwrap().close().unwrap();
+                                fobjtable.remove(&inodenum);
+                            }
```
jesings, line 1081: Character devices don't have associated file objects, this code should be removed

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                            if char_inode_obj.refcount == 0 {
```
jesings, line 1082: refcount again known to be 0 by this point

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                    let mut fobjtable = FILEOBJECTTABLE.write().unwrap();
+
+                    match inodeobj {
+                        Inode::File(ref mut normalfile_inode_obj) => {
+                            normalfile_inode_obj.refcount -= 1;
+
+                            //if there are still references to the file, then do nothing to the file
+                            if normalfile_inode_obj.refcount != 0 {
+                                return 0;
+                            }
```
jesings, line 1031: You still want to go through with removing the file descriptor from the fdtable even if the refcount is non-zero, the code will need to be refactored to allow that

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                                } 
+                            }
+                        },
+                        Inode::Dir(ref mut dir_inode_obj) => {
+                            dir_inode_obj.refcount -= 1;
+
+                            //if there are still references to the file, then do nothing to the file
+                            if dir_inode_obj.refcount != 0 {
+                                return 0;
+                            }
```
jesings, line 1054: You still want to go through with removing the file descriptor from the fdtable even if the refcount is non-zero, the code will need to be refactored to allow that

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                                } 
+                            }
+                        },
+                        Inode::CharDev(ref mut char_inode_obj) => {
+                            char_inode_obj.refcount -= 1;
+
+                            //if there are still references to the file, then do nothing to the file
+                            if char_inode_obj.refcount != 0 {
+                                return 0;
+                            }
```
jesings, line 1075: You still want to go through with removing the file descriptor from the fdtable even if the refcount is non-zero, the code will need to be refactored to allow that

src/safeposix/syscalls/fs_calls.rs
jesings: Also whether refcount is 0 or not

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -910,6 +909,421 @@ impl Cage {
...
+                    if mode & (S_IRWXA|(S_FILETYPEFLAGS as u32)) == mode {
+                        general_inode.mode = (general_inode.mode &!S_IRWXA) | mode
+                    } else {
+                        //there doesn't seem to be a good syscall error errno for this
+                        return syscall_error(Errno::EACCES, "chmod", "provided file mode is not valid");
+                    }
+                    //persist metadata
+                    return 0;
+                },
+            }
```
jesings, line 1288: If you persist metadata at the end of every arm, then you should just persist at the end instead

jesings: Also return at the end, and error out on invalid mode  before this match, we want to duplicate as little code as possible

jesings (commented):
Some pretty substantial changes should be made to close, and there are a smattering of other things to do but overall this looks like a ton of great progress!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -133,7 +133,60 @@ impl Cage {
...
+            //If neither the file nor parent exists
+            (None, None) => {
+                syscall_error(Errno::ENOENT, "mkdir", "a directory component in pathname does not exist or is a dangling symbolic link")
+            }
+
+            //If the file doesn't exist but the parent does
+            (None, Some(pardirinode)) => {
+                let filename = truepath.file_name().unwrap().to_str().unwrap().to_string(); //for now we assume this is sane, but maybe this should be checked later
+
+                let effective_mode = S_IFREG as u32 | mode;
```
jesings: Mkdir does not make a regular file, it makes a directory, replace this with S_IFDIR

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -961,13 +1014,8 @@ impl Cage {
...
         }
 
-        //need to add close helper and reference
-        match fdtable.get(&newfd) {
-            Some(_) => {return Self::_close_helper(&self, newfd, Some(&fdtable), false);},
-            None => {
-                fdtable.insert(newfd, fdtable.get(&oldfd).unwrap().clone());
-            } 
-        }
+        if fdtable.contains_key(&newfd) {return Self::_close_helper(&self, newfd, Some(&fdtable), false);}
```
jesings, line 1017: The close result shouldn't be returned, it should still add to the fdtable even if it has to close what was there previously

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
-                            }
-                            if fobjtable.contains_key(&inodenum) {
-                                //close and delete the file in the object table
-                                fobjtable.get(&inodenum).unwrap().close().unwrap();
-                                fobjtable.remove(&inodenum);
+                        //if it's not a reg file, then we have nothing to close
+                        if !is_reg(normalfile_inode_obj.mode) {
+                            match fobjtable.get(&inodenum) {
+                                Some(_) => {syscall_error(Errno::ENOEXEC, "close or dup", "Non-regular file in file object table");},
+                                None => {return 0;}
```
jesings, line 1079: You still want to do the rest of the close stuff for non-regular files--like removing from fdtable and removing the inode if the refcount and linkcount are 0

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
+                        fobjtable.get(&inodenum).unwrap().close().unwrap();
+                        fobjtable.remove(&inodenum);
 
-                            //if there are still references to the file, then do nothing to the file
-                            if char_inode_obj.refcount != 0 {
-                                return 0;
+                        //if there are still references to the file, then do nothing to the file
+                        if normalfile_inode_obj.refcount != 0 {
+                            return 0;
+                        }
```
jesings, line 1089: You still want to remove from the fdtable (at the end of this function) if refcount is non-zero, you should check
```
if normalfile_inode_obj.refcount == 0 && if normalfile_inode_obj.linkcount == 0
    //remove from fobjtable
    //remove from inodetable
}
```
As the check for the inodetable removal

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
-                    let pipenumber = pipe_filedesc_obj.pipe;
-                    let read_references = Self::_lookup_refs_by_pipe_end(self, pipenumber, O_RDONLY);
-                    let write_references = Self::_lookup_refs_by_pipe_end(self, pipenumber, O_WRONLY);
+        //Decide how to proceed depending on the fd type.
+        //First we check in the file descriptor to handle sockets, streams, and pipes,
+        //and if it is a normal file descriptor we decrement the refcount to reflect
+        //one less reference to the file.
+        match &*filedesc_enum {
+            //if we are a socket, we dont change disk metadata
+            Stream(_) => {return 0;},
```
jesings, line 1048: Evn if you don't chang disk metadata, don't return here because you still need to remove from fdtable

src/safeposix/syscalls/fs_calls.rs
jesings: And anyway this whole block is unnecessary because all files that have inodes that are Inode::File are assumed to be reg

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
+                    Inode::Dir(ref mut dir_inode_obj) => {
+                        dir_inode_obj.refcount -= 1;
+
+                        //if it's not a reg file, then we have nothing to close
+                        if !is_reg(dir_inode_obj.mode) {
+                            match fobjtable.get(&inodenum) {
+                                Some(_) => {syscall_error(Errno::ENOEXEC, "close or dup", "Non-regular file in file object table");},
+                                None => {return 0;}
                             }
+                        }
```
jesings: No dirs are regular files, so rather than this whole statement, as mentioned earlier, don't return, just return a syscall error if there is a corresponding entry in the fobj table

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
+                        
+                        if dir_inode_obj.linkcount == 0  {
+                            
+                            //removing the file from the entire filesystem
+                            mutmetadata.inodetable.remove(&inodenum);
+                            let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
+                            interface::removefile(sysfilename).unwrap();
+    
+                        }
```
jesings, line 1123: Same comment abut don't return on refcount as above, but no need to remove the file object in that case

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
-                                    let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
-                                    interface::removefile(sysfilename).unwrap();
-            
-                                } 
-                            }
-                        },
-                        Inode::CharDev(ref mut char_inode_obj) => {
-                            char_inode_obj.refcount -= 1;
+                        fobjtable.get(&inodenum).unwrap().close().unwrap();
+                        fobjtable.remove(&inodenum);
```
jesings, line 1084: Only remove from the fobjtable if the refcount is 0 and the linkcount is 0

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
-                },
-            }
-            // if this is a part of the close syscall, then the fd table instance of the fd is removed
-            if is_close_syscall {
-                self.filedescriptortable.write().unwrap().remove(&fd);
-            }
-            return 0; //_close_helper has succeeded!
-        } else {
-            return syscall_error(Errno::ENOENT, "close or dup", "invalid file descriptor");
+                        }
```
jesings: No chardevs are regular files so the same comment as for dirs applies

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
+                        
+                        if char_inode_obj.linkcount == 0  {
+                            
+                            //removing the file from the entire filesystem
+                            mutmetadata.inodetable.remove(&inodenum);
+                            let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
+                            interface::removefile(sysfilename).unwrap();
+    
+                        } 
+                    },
```
jesings, line 1149: Again, don't return and check differently, the same way as you do for dirs

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -988,118 +1036,122 @@ impl Cage {
...
+                    Inode::Pipe(_) | Inode::Socket(_) => {panic!("How did you get by the first filter?");},
+                }
+            },
         }
+        return 0; //_close_helper has succeeded!
```
jesings, line 1154: remove from the fdtable before returning

jesings (commented):
Comments on the updated close and on mkdir, great progress!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1073,28 +1073,17 @@ impl Cage {
...
+                                let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
+                                interface::removefile(sysfilename).unwrap();
+                            }   
                             //removing the file from the entire filesystem
                             mutmetadata.inodetable.remove(&inodenum);
-                            let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
-                            interface::removefile(sysfilename).unwrap();
-    
+                            fobjtable.get(&inodenum).unwrap().close().unwrap();
+                            fobjtable.remove(&inodenum);
```
jesings: should only be removed if both linkcount and refcount are 0

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1107,19 +1096,13 @@ impl Cage {
...
                             //removing the file from the entire filesystem
-                            mutmetadata.inodetable.remove(&inodenum);
-                            let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
-                            interface::removefile(sysfilename).unwrap();
-    
+                            if dir_inode_obj.refcount == 0 {
+                                let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
+                                interface::removefile(sysfilename).unwrap();
+                            }
+                            mutmetadata.inodetable.remove(&inodenum);
```
jesings, line 1105: Should only be removed if both refcount and linkcount are 0

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1134,23 +1117,20 @@ impl Cage {
...
-                        
-                        if char_inode_obj.linkcount == 0  {
-                            
+                        if char_inode_obj.linkcount == 0 {
                             //removing the file from the entire filesystem
+                            if char_inode_obj.refcount == 0 {
+                                let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
+                                interface::removefile(sysfilename).unwrap();
+                            }
                             mutmetadata.inodetable.remove(&inodenum);
```
jesings, line 1126: Should only be removed if refcount and linkcount are 0

jesings (commented):
It's getting a lot closer to how we want it!
Other than that you commit a .DS_STORE vscode editor file so you're gonna want to remove that from the repo.

TristanB22 (commented):
I think this is ready to be merged with main, now

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+        if newfd >= MAXFD || newfd <= STARTINGFD {
+            return syscall_error(Errno::EBADF, "dup or dup2", "provided file descriptor is out of range");
+        }
+
+        //if the file descriptors are equal, return the new one
+        if newfd == oldfd {
+            return newfd;
+        }
+
+        if fdtable.contains_key(&newfd) {Self::_close_helper(&self, newfd, Some(fdtable));}
```
jesings, line 1017: Return the ERRNO result if close_helper fails

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+    }
+
+    pub fn _close_helper(&self, fd: i32, fdtable_lock: Option<&mut FdTable>) -> i32 {
+        let mut writer;
+        let fdtable = if let Some(rl) = fdtable_lock {rl} else {
+            writer = self.filedescriptortable.write().unwrap(); 
+            &mut writer
+        };
+
+        let filedesc = fdtable.get(&fd).unwrap();
```
jesings, line 1039: This would be better named locked_filedesc or something like that

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+        let filedesc_enum = filedesc.write().unwrap();
```
jesings, line 1040: This probably can be a read rather than a write

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                        //if it's not a reg file, then we have nothing to close
+                        //Inode::File is a regular file by default
+                        
+                        if normalfile_inode_obj.linkcount == 0 && normalfile_inode_obj.refcount == 0 {
+                            let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
+                            interface::removefile(sysfilename).unwrap();
+                            //removing the file from the entire filesystem
+                            mutmetadata.inodetable.remove(&inodenum);
+                            fobjtable.get(&inodenum).unwrap().close().unwrap();
+                            fobjtable.remove(&inodenum);
```
jesings, line 1085: No need to get and remove, just remove

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                    },
+                    Inode::Dir(ref mut dir_inode_obj) => {
+                        dir_inode_obj.refcount -= 1;
+
+                        //if it's not a reg file, then we have nothing to close
+                        match fobjtable.get(&inodenum) {
+                            Some(_) => {syscall_error(Errno::ENOEXEC, "close or dup", "Non-regular file in file object table");},
+                            None => {}
+                        }
+                        if dir_inode_obj.linkcount == 0 && dir_inode_obj.refcount == 0 {
```
jesings, line 1096: The minimum linkcount for dirs should be 2, so you should check it against 2 rather than 0

jesings: The 2 links being .. and .

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                            fobjtable.get(&inodenum).unwrap().close().unwrap();
+                            fobjtable.remove(&inodenum);  
+                        } 
+                    },
+                    Inode::Dir(ref mut dir_inode_obj) => {
+                        dir_inode_obj.refcount -= 1;
+
+                        //if it's not a reg file, then we have nothing to close
+                        match fobjtable.get(&inodenum) {
+                            Some(_) => {syscall_error(Errno::ENOEXEC, "close or dup", "Non-regular file in file object table");},
```
jesings, line 1093: return this error, otherwise it's dangling

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                            //removing the file from the metadata 
+                            mutmetadata.inodetable.remove(&inodenum);
+                        } 
+                    },
+                    Inode::CharDev(ref mut char_inode_obj) => {
+                        char_inode_obj.refcount -= 1;
+
+                        //if it's not a reg file, then we have nothing to close
+                        match fobjtable.get(&inodenum) {
+                            Some(_) => {syscall_error(Errno::ENOEXEC, "close or dup", "Non-regular file in file object table");},
```
jesings, line 1106: return this error, otherwise it's dangling

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                    }
+                },
+                // set the flags but make sure that the flags are valid
+                (F_SETFD, arg) if arg >= 0 => {
+                    match &mut *filedesc_enum {
+                        //Match type (since each has a different file descriptor type)
+                        Pipe(ref mut obj) => {
+                            //bitwise or on the flags
+                            obj.flags = obj.flags | O_CLOEXEC;
+                            return 0;
```
jesings, line 1160: One return 0 at the end would minimize duplicate code

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                        File(obj) => {return obj.flags;},
+                    }
+                },
+
+                (F_SETFL, arg) if arg >= 0 => {
+                    //TO DO: check that the type of x is an int or a long 
+                    match &mut *filedesc_enum {
+                        Pipe(ref mut obj) => {
+                            obj.flags = arg;
+                            return 0;
```
jesings, line 1191: One return 0 at the end would minimize duplication

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                        },
+                        File(ref mut obj) => {
+                            obj.flags = arg;
+                            return 0;
+                        },
+                    }
+                },
+                (F_DUPFD, arg) if arg >= 0 => {
+                    //TO DO: check that the type of x is an int or a long 
+                    Self::dup_syscall(self, fd, Some(arg));
```
jesings, line 1209: You should just return the result of this

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                    return 0;
+                },
+                //TO DO: implement. this one is saying get the signals
+                (F_GETOWN, ..) => {
+                    match &*filedesc_enum {
+                        Pipe(_) => {
+                            //TO DO: traditional SIGIO behavior
+                            return 0;
```
jesings, line 1217: Again return 0 after match instead of in each arm

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+        //check if there is a valid path or not there to an inode
+        if let Some(inodenum) = metawalk(truepath.as_path(), Some(&metadata)) {
+            let thisinode = metadata.inodetable.get_mut(&inodenum).unwrap();
+            if mode & (S_IRWXA|(S_FILETYPEFLAGS as u32)) == mode {
+                match thisinode {
+                    Inode::File(ref mut general_inode) => {
+                        general_inode.mode = (general_inode.mode &!S_IRWXA) | mode
+                    },
+                    Inode::CharDev(ref mut dev_inode) => {
+                            dev_inode.mode = (dev_inode.mode &!S_IRWXA) | mode;
```
jesings, line 1260: wonky indent

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +963,454 @@ impl Cage {
...
+                    //         return 0;
+                    //     }
+                    //     if let guard = filedesc.access_lock.try_write().unwrap() {return 0;} 
+                    //     else {return syscall_error(Errno::EAGAIN, "flock", "exclusive lock couldn't be acquired");}
+                    // } else if operation & LOCK_UN == 0 {
+                    //     if let Some(_) = filedesc.readLock.pop() {return 0;}
+                    //     else {return syscall_error(Errno::EAGAIN, "flock", "file is already unlocked");} 
+                    // } 
+                    // panic!("Should not be possible to hit this code");
+                    return 0;
```
jesings, line 1406: even for this temporary one, just return 0 once at the end

jesings (commented):
Looking great! All it needs is some reorganization

rennergade (commented):
Mostly looks really good! I agree with Jonathans comments. In addition:

- I think its worth removing flock from this PR completely at this point, since it may take awhile
- I think the times you pass a read/write guard into a function and have to reassign it with writer/reader etc its probably worth a short comment to whats going on. I know its straight from the example but helpful for people to not have to think too much about it.

TristanB22:
Made some final changes -- I think that it is ready for merge now. May I merge it?

TristanB22:
Oops -- that's not true. Still need to implement the refcount changes. After that it should be ready.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -968,17 +967,14 @@ impl Cage {
...
         let mut fdtable = self.filedescriptortable.write().unwrap();
 
+        //if a starting fd was passed, then use that as the starting point, but otherwise, use the designated minimum of STARTINGFD
         let start_fd = match start_desc {
             Some(start_desc) => start_desc,
             None => STARTINGFD,
         };
 
-        //checking whether the fd exists in the file table and is higher than the starting file descriptor
+        //checking whether the fd exists in the file table and is higher than the starting file descriptor or not
```
jesings, line 976: The higher than the starting file descriptor or not part is no longer relevant to the current code

TristanB22:
I think that this is ready to merge -- may I do so?

src/safeposix/cage.rs
```diff
@@ -19,14 +19,20 @@ pub enum FileDescriptor {
 pub struct FileDesc {
     pub position: usize,
     pub inode: usize,
-    pub flags: i32
+    pub flags: i32,
+    // pub access_lock: interface::RustLock<()>,
```
rennergade, line 23: remove

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +962,354 @@ impl Cage {
...
+                    }
+                },
+            }
+        }
+        //removing inode from fd table
+        fdtable.remove(&fd);
+        return 0; //_close_helper has succeeded!
+    }
+
+    pub fn _lookup_refs_by_pipe_end(&self, pipenumber: usize, flags: i32) -> i32 {
```
rennergade, line 1161: unused

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -964,6 +962,354 @@ impl Cage {
...
+        let pipe_references = 0;
+        //NOT COMPLETE
+        return 0;
+    }
+    
+    pub fn _cleanup_socket(&self, fd: &i32, partial: bool) -> i32 {
```
rennergade, line 1167: unused

src/safeposix/syscalls/fs_constants.rs
```diff
@@ -65,6 +85,11 @@ pub const S_IFREG: i32 = 0o100000;
 pub const S_IFSOCK: i32 = 0o140000;
 pub const S_FILETYPEFLAGS: i32 = 0o170000;
 
+//for flock syscall
+pub const LOCK_SH: i32 = 1;
```
rennergade, line 89: adding these are probably fine

rennergade (commented):
Logic seems sound, but theres a bunch of unused/commented code that I think should be taken out still. 

Also are there any tests with this?

rennergade (approved):
Approved. Congrats!

PR #33, opened 2021-07-11
Added rmdir() and rename()

kliu513:
I include both the syscalls in one PR because they are similar. I don't know why they can't merge with the main branch, though.

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1386,18 +1386,18 @@ impl Cage {
                 match inodeobj {
                     // make sure inode matches a directory
                     Inode::Dir(dir_obj) => {
-                        if dir_obj.linkcount > 2 {return syscall_error(Errno::EEXIST, "rmdir", "Directory is not empty");}
-                        if !is_dir(dir_obj.mode) {return syscall_error(Errno::ENOTDIR, "rmdir", "Mode is not set to directory");}
+                        if dir_obj.linkcount > 2 {return syscall_error(Errno::ENOTEMPTY, "rmdir", "Directory is not empty");}
+                        if !is_dir(dir_obj.mode) {panic!("The dir does not have its mode set to dir");}
 
                         // check if dir has write permission
-                        if dir_obj.mode as u32 & S_IWOTH == 0 {return syscall_error(Errno::EPERM, "rmdir", "Directory does not have write permission")}
+                        if dir_obj.mode as u32 & (S_IWOTH | S_IWGRP | S_IWUSR) == 0 {return syscall_error(Errno::EPERM, "rmdir", "Directory does not have write permission")}
                         
                         // remove entry of corresponding inodenum from inodetable
                         metadata.inodetable.remove(&inodenum);
                         
                         if let Inode::Dir(parent_dir) = metadata.inodetable.get_mut(&parent_inodenum).unwrap() {
                             // check if parent dir has write permission
-                            if parent_dir.mode as u32 & S_IWOTH == 0 {return syscall_error(Errno::EPERM, "rmdir", "Parent directory does not have write permission")}
+                            if parent_dir.mode as u32 & (S_IWOTH | S_IWGRP | S_IWUSR) == 0 {return syscall_error(Errno::EPERM, "rmdir", "Parent directory does not have write permission")}
                             
                             // remove entry of corresponding filename from filename-inode dict
                             parent_dir.filename_to_inode_dict.remove(&truepath.file_name().unwrap().to_str().unwrap().to_string());
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1387,7 +1387,7 @@ impl Cage {
                     // make sure inode matches a directory
                     Inode::Dir(dir_obj) => {
                         if dir_obj.linkcount > 2 {return syscall_error(Errno::ENOTEMPTY, "rmdir", "Directory is not empty");}
-                        if !is_dir(dir_obj.mode) {panic!("The dir does not have its mode set to dir");}
+                        if !is_dir(dir_obj.mode) {panic!("This directory does not have its mode set to S_IFDIR");}
 
                         // check if dir has write permission
                         if dir_obj.mode as u32 & (S_IWOTH | S_IWGRP | S_IWUSR) == 0 {return syscall_error(Errno::EPERM, "rmdir", "Directory does not have write permission")}
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1376,7 +1376,7 @@ impl Cage {
             (None, ..) => {
                 syscall_error(Errno::EEXIST, "rmdir", "Path does not exist")
             }
-            (Some(_), None) {
+            (Some(_), None) => {
                 syscall_error(Errno::EBUSY, "rmdir", "Cannot remove root directory")
             }
             (Some(inodenum), Some(parent_inodenum)) => {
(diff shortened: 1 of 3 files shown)


rennergade (commented):
Looks very good! A couple general comments.

In rmdir, it doesn't look like you check to make sure the directory is empty, which is important. It also might make sense to make sure the mode is set to directory (even if we are getting a Dir inode type here for the enum). Even further, there's a note in the Repy version about checking permissions. I think its worth putting that in! We just want to make sure both the directory, and its parent have the 'w' permission. (I think that's right?).

Would also like a few comments so its easier to follow along.

You also need to link this to the dispatcher, and adding a test is good and pretty easy to do in Rust!

jesings (commented):
Overall really good! I just had some tiny comments on rmdir, and rename should be relatively easy to fix as well.

kliu513:
Please re-review this PR! I have fixed everything regarding implementation and added tests for the syscalls respectively.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1364,4 +1364,90 @@ impl Cage {
...
+            (Some(_), None) => { // path exists but parent does not => path is root dir
+                syscall_error(Errno::EBUSY, "rmdir", "Cannot remove root directory")
+            }
+            (Some(inodenum), Some(parent_inodenum)) => {
+                let inodeobj = metadata.inodetable.get_mut(&inodenum).unwrap();
+
+                match inodeobj {
+                    // make sure inode matches a directory
+                    Inode::Dir(dir_obj) => {
+                        if dir_obj.linkcount > 2 {return syscall_error(Errno::EEXIST, "rmdir", "Directory is not empty");}
```
jesings, line 1389: The right error code here is ENOTEMPTY

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1364,4 +1364,90 @@ impl Cage {
...
+                        if !is_dir(dir_obj.mode) {return syscall_error(Errno::ENOTDIR, "rmdir", "Mode is not set to directory");}
```
jesings, line 1390: This maybe should be a panic, as Inode::Dir must be a directory. @rennergade thoughts?

jesings: Either that or don't have this check at all possibly

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1364,4 +1364,90 @@ impl Cage {
...
+
+                        // check if dir has write permission
+                        if dir_obj.mode as u32 & S_IWOTH == 0 {return syscall_error(Errno::EPERM, "rmdir", "Directory does not have write permission")}
```
jesings, line 1393: I think you need to replace `S_IWOTH` with `(S_IWOTH | S_IWGRP | S_IWUSR)`

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1364,4 +1364,90 @@ impl Cage {
...
+                        
+                        // remove entry of corresponding inodenum from inodetable
+                        metadata.inodetable.remove(&inodenum);
+                        
+                        if let Inode::Dir(parent_dir) = metadata.inodetable.get_mut(&parent_inodenum).unwrap() {
+                            // check if parent dir has write permission
+                            if parent_dir.mode as u32 & S_IWOTH == 0 {return syscall_error(Errno::EPERM, "rmdir", "Parent directory does not have write permission")}
```
jesings, line 1400: Replace `S_IWOTH` with `(S_IWOTH | S_IWGRP | S_IWUSR)` again

src/tests/fs_tests.rs
```diff
@@ -17,7 +17,9 @@ mod fs_tests {
         // ut_lind_fs_fdflags();
         // ut_lind_fs_file_link_unlink();
         // ut_lind_fs_file_lseek_past_end();
-        ut_lind_fs_fstat_complex();
+        // ut_lind_fs_fstat_complex();
+        // ut_lind_fs_rmdir();
```
jesings, line 21: Why are these commented out?

src/tests/fs_tests.rs
```diff
@@ -582,4 +584,29 @@ mod fs_tests {
         assert_eq!(cage.exit_syscall(), 0);
         lindrustfinalize();
     }
+
+    pub fn ut_lind_fs_rmdir() {
+        lindrustinit();
+        let cage = {CAGE_TABLE.read().unwrap().get(&1).unwrap().clone()};
+
+        let path = String::from("/parent_dir/dir");
+        assert_eq!(cage.mkdir_syscall(&String::from("/parent_dir"), S_IRWXA), 0);
```
jesings, line 593: Remove the &String::from call, just use a &str literal

src/tests/fs_tests.rs
```diff
@@ -582,4 +584,29 @@ mod fs_tests {
         assert_eq!(cage.exit_syscall(), 0);
         lindrustfinalize();
     }
+
+    pub fn ut_lind_fs_rmdir() {
+        lindrustinit();
+        let cage = {CAGE_TABLE.read().unwrap().get(&1).unwrap().clone()};
+
+        let path = String::from("/parent_dir/dir");
```
jesings, line 592: Remove the &String::from call, just use a &str literal

src/tests/fs_tests.rs
```diff
@@ -582,4 +584,29 @@ mod fs_tests {
...
+        assert_eq!(cage.mkdir_syscall(&String::from("/parent_dir"), S_IRWXA), 0);
+        assert_eq!(cage.mkdir_syscall(&path, S_IRWXA), 0);
+        assert_eq!(cage.rmdir_syscall(&path), 0);
```
jesings, line 595: Once you remove the &String::from calls above, remove the borrows of path

src/tests/fs_tests.rs
```diff
@@ -582,4 +584,29 @@ mod fs_tests {
...
+        lindrustfinalize();
+    }
+
+    pub fn ut_lind_fs_rename() {
+        lindrustinit();
+        let cage = {CAGE_TABLE.read().unwrap().get(&1).unwrap().clone()};
+
+        let old_path = String::from("/test_dir");
+        assert_eq!(cage.mkdir_syscall(&old_path, S_IRWXA), 0);
+        assert_eq!(cage.rename_syscall(&old_path, &String::from("/test_dir_renamed")), 0);
```
jesings, line 607: Remove the &String::from calls, just use a &str literal, and do not borrow old_path

jesings (commented):
I only had a few minor changes for you to make, overall it looks great!

src/safeposix/syscalls/fs_calls.rs
rennergade: Hmm. Yeah, I do think the check is worth having here. But I think I agree having it panic instead of return an error here makes more sense because if we have a directory with the mode incorrectly set then something is majorly wrong.

rennergade:
Looks really good. I think if you address the comments @jesings added then this will be good to go.

src/tests/fs_tests.rs
kliu513: All the tests except the latest added one are commented out. I guess people want to focus on the latest one only. Should I uncomment all the tests or it's okay to have them this way since we are still adding more tests?

TristanB22: Hi all -- that is my bad. I wrote the rmdir test before your PR was merged, so I commented it out. Feel free to comment them back in -- my current PR does the same thing and will update main when it is merged.

src/tests/fs_tests.rs
kliu513: Using &str literal directly results in the error message "a directory component in pathname does not exist or is a dangling symbolic link." It seems that we have to use &String::from.

kliu513: Never mind, I just fixed it. Sorry!

kliu513:
I think I have fixed everything mentioned.

jesings (commented):
This looks terrific! @rennergade we can merge whenever

rennergade:
Ok great. Thanks @kliu513

PR #34, opened 2021-07-12
Flock implemented

jesings:
Let's not try to merge this until at least Tristan's PR is merged

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1,4 +1,5 @@
 // File system related system calls
+use libc::{LOCK_NB,LOCK_UN,LOCK_SH,LOCK_EX};
 
 use crate::interface;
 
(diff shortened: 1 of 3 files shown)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1,4 +1,5 @@
 // File system related system calls
+use libc::{LOCK_NB,LOCK_UN,LOCK_SH,LOCK_EX};
```
rennergade, line 2: we shouldn't put any use's in here for consistency, even if these are just constants. let's just define these in our constants file?

rennergade (changes requested):
I'd strongly prefer we put those constants in the constants file, and that we just use the value and not import from libc. Otherwise, this took a lot of work and looks great, so nice job!

rennergade (approved):
looks good!

PR #39, opened 2021-07-13
Added fs utils script (WIP)

jesings:


rennergade (commented):
I think this is really well written. Are there tests that have been run with this?

We've also had fs persistence issues with tests earlier, does this fix those?

jesings:
> I think this is really well written. Are there tests that have been run with this?
> 
> We've also had fs persistence issues with tests earlier, does this fix those?

I've done some tests with copying various directories from my machine, updating them, and removing them with deltree, and it all works as it should. This will not fix the persistence issues although it may make it easier for us to fix them by using the deltree command.

TristanB22 (commented):
This is super cool to read though -- impressive! I don't know if this is super important, but do you think that adding flag options to some of the calls would be beneficial? Especially since that is what Native does? Like, in rmdir, I know that some people use the "-r" option to essentially call rmtree.

rennergade:
Ok I think this is good. I'm going to merge this so it can help Tristan clean up the tests.

Issue #12, opened 2021-06-14
Translate Tests


rennergade:
We want to translate the tests from repy SafePOSIX to here. [This test directory](https://github.com/Lind-Project/nacl_repy/tree/master/seattlelib/tests) has the tests we want to translate.

There's probably more than we need here, at this point we should probably just translate the unit tests that being with "ut_lind_"

TristanB22:
ut_lind_fs_broken_close.py
ut_lind_fs_chmod.py
ut_lind_fs_dir_chdir.py
ut_lind_fs_dir_mode.py
ut_lind_fs_dir_multiple.py
ut_lind_fs_dup.py
ut_lind_fs_dup2.py
ut_lind_fs_fcntl.py
ut_lind_fs_fdflags.py
ut_lind_fs_file_link_unlink.py
ut_lind_fs_file_lseek_past_end.py
ut_lind_fs_file_read_write.py
ut_lind_fs_fstat_complex.py
ut_lind_fs_getdents.py
ut_lind_fs_getuid.py
ut_lind_fs_load_fs.py
ut_lind_fs_mknod.py
ut_lind_fs_multiple_open.py
ut_lind_fs_persistance_setup.py
ut_lind_fs_persistance_test.py
ut_lind_fs_rmdir.py
ut_lind_fs_simple.py
ut_lind_fs_stat_file_complex.py
ut_lind_fs_stat_file_mode.py
ut_lind_fs_statfs.py
ut_lind_fs_truncate.py
ut_lind_net_bind.py
ut_lind_net_bind_multiple.py
ut_lind_net_bind_on_zero.py
ut_lind_net_connect_basic_udp.py
ut_lind_net_getpeername.py
ut_lind_net_getsockname.py
ut_lind_net_listen.py
ut_lind_net_poll.py
ut_lind_net_recvfrom.py
ut_lind_net_select.py
ut_lind_net_shutdown.py
ut_lind_net_socket.py
ut_lind_net_socketoptions.py
ut_lind_net_socketpair.py
ut_lind_net_udp_bad_bind.py
ut_lind_net_udp_connect.py
ut_lind_net_udp_simple.py
ut_seattlelibtests_teststruct.mix

rennergade:
Now that we've completed most of the calls lets make sure all of these are translated, and try to add more corner cases.

TristanB22:
All of the file system tests are done! Will work on net tests and then add some more corner cases.

kliu513:
[...]

PR #42, opened 2021-07-16
Adding uid, load_fs, mknod, multiple_open, persistence setup, persistence test, rmdir, stat_complex, stat_mode tests and fixing zerofill fn

TristanB22:
Please take a closer look at the mknod call -- specifically the second portion. The repy test says 
```
assert lind_test_server.get_fscall_obj(-1).write_syscall(fd, "test") == 4,\
  "Write failed to /dev/random file."
assert lind_test_server.get_fscall_obj(-1).read_syscall(fd, 10) != '',\
  "Read failed from /dev/random file."
```
which doesn't really make sense to me. But, when I write 
```        
assert_eq!(cage.lseek_syscall(fd2, 0, SEEK_SET), 0);
assert_eq!(cage.write_syscall(fd2, str2cbuf("testing"), 7), 7);
assert_eq!(cage.lseek_syscall(fd2, 0, SEEK_SET), 0);
assert_ne!(cage.read_syscall(fd2, buf2.as_mut_ptr(), 7), 0);
```
it seems to work. I guess it is correct, but it confuses me. 

Also, the code in rmdir can be uncommented and run once Kaitlyn's PR is approved. It should work with her syscall provided the success code is 0.

With this PR, I can't write any more of the tests without more syscalls being implemented. Check [this](https://github.com/Lind-Project/safeposix-rust/issues/12) issue for a full synopsis.


src/tests/fs_tests.rs
```diff
@@ -303,38 +378,401 @@ mod fs_tests {
...
+    }
+
+
+
+    pub fn ut_lind_fs_mknod() {
+        // let's create /dev/null
+        lindrustinit();
+        let cage = {CAGE_TABLE.read().unwrap().get(&1).unwrap().clone()};
+        let dev = makedev(&DevNo {major: 1, minor: 3});
+        let path = String::from("/null");
```
jesings, line 628: Again, same comment

src/tests/fs_tests.rs
```diff
@@ -303,38 +378,401 @@ mod fs_tests {
...
+        let cage = {CAGE_TABLE.read().unwrap().get(&1).unwrap().clone()};
+        let path = String::from("/fooFileMode");
+        let fd = cage.open_syscall(&path, O_CREAT | O_EXCL | O_WRONLY, S_IRWXA);
+
+        let mut statdata = StatData::default();
+        assert_eq!(cage.stat_syscall(&path, &mut statdata), 0);
+        assert_eq!(statdata.st_mode, S_IRWXA | S_IFREG as u32);
+
+        //make a file without permissions and check that it is a reg file without permissions
+        let path2 = String::from("/fooFileMode2");
```
jesings, line 768: String::from one last time should be removed

jesings (commented):
Aside from a few spot comments, my issues largely boil down to not asserting the result of stat enough and overusing String::from. First of all, it's more conventional to call "stringconst".to_string() to get a String from a &str literal, but second of all, the syscalls you're calling take a &str rather than a String, and you borrow that String  in each call to make an &String which coerces to a &str. It'd be better if you just assigned the &str literal to the variables and then didn't borrow the path in the call. 

For example:

```rs
let path = "/testpath";
cage.open_syscall(path, O_RDWR, 0);
```

rennergade:
This looks good. Seems like you need to fix some merge conflicts though. Is this finalized?

PR #47, opened 2021-07-25
Add getdents()

kliu513:


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1450,4 +1450,76 @@ impl Cage {
...
+                        // only proceed when inode is a dir
+                        Inode::Dir(dir_inode_obj) => {
+                            let position = normalfile_filedesc_obj.position;
+                            let mut bufcount = 0;
+                            let mut curr_size = 0;
+                            let mut curr_pos = 0;
+                            let mut vec_idx = 0;
+                            let mut count = 0;
+
+                            for (filename, inode) in dir_inode_obj.filename_to_inode_dict.clone().into_iter() {
```
jesings, line 1485: You don't actually insert anything into vec!
What I'd do is first construct and fill in the name u8 vector, then update bufcount, and construct the clipped dirent using a struct instantiator (i.e. `ClippedDirent { d_ino: inode, d_off: bufcount, d_reclen: curr_size}` of course with the requisite casts), and then insert the tuple containing these into vec like `vec.push((curdirent, namevec))`

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1450,4 +1450,76 @@ impl Cage {
...
+                                if curr_pos >= position {
+                                    vec[vec_idx].1 = filename.as_bytes().to_vec();
+                                    for n in 0..(vec.len() + 7) / 8 * 8 - vec.len() {
+                                        vec[vec_idx].1.push(00);
+                                    }
+                                    vec[vec_idx].1.push(libc::DT_UNKNOWN);
```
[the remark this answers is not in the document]
jesings: Also, it's supposed to be in place of the last padding byte, not in addition to it (as that would ruin alignment)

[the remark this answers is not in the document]
jesings: So in order to make that work, you need to make sure that it pads with at leats 1 byte (which means in the case where vec.len is precisely divisible by 8, you must pad with 8 bytes--so here you should add 8 rather than to vec.len().
Then you need to subtract on to the end of that big term that is the number of iterations you push in so that the DT_UNKNOWN is in place of a padding byte rather than in addition to it.

jesings (commented):
This is certainly coming along! There are some comments which should require a little bit of rearchitecting things, but the major thing about this PR is that you should comment as much as you can inside the syscall because this syscall is really confusing and won't make any sense if it's not well documented.

src/safeposix/syscalls/fs_calls.rs
jesings: Also, I think the d_off in the last dirent in the syscall needs to have some value placed into it, what value exactly that is, I'm not sure, some tests would help you figure it out. The man page test uses 4096, but I'm not sure whether that depends on the number of directory entries or not. If you'd like some help testing I'd be happy to help you.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1450,4 +1450,76 @@ impl Cage {
...
+                                    bufcount += curr_size;
+                                    if bufcount > bufsize {break;}
+                                    vec[vec_idx].0.d_ino = inode as u64;
+                                    vec[vec_idx].0.d_off = bufcount as u64;
+                                    vec[vec_idx].0.d_reclen = curr_size as u16;
+                                    count += 1;
+                                }
+                                curr_pos += 1;
+                            }
+                            normalfile_filedesc_obj.position = std::cmp::min(position + count, dir_inode_obj.filename_to_inode_dict.len());
```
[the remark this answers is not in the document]
rennergade: I may need a min function for pipes. May be worth adding our own min/max to the interface, even if we don't import it just for general use.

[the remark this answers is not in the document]
rennergade: I just added rust_min and rust_max to the interface, if you pull main and merge it in here you can use that instead.

src/safeposix/syscalls/fs_calls.rs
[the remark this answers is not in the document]
jesings: Rereading the spec, DT_UNKNOWN actually goes before the first byte that's strictly a padding byte, so the implementation of that should be relatively easy from your current implementation, just put it before your padding.
Additionally, d_reclen should be taken from this value rather than from the padded value.

src/safeposix/syscalls/fs_calls.rs
kliu513: I did some research online and what people usually do is just leave d_off in the last dirent as zero, so I didn't make change on this one. Otherwise, I think everything mentioned has been fixed.

rennergade:
This all looked good. You were right that we had to keep the fd mutable, because we had to update position. The only other issue I saw was you used a std library call in safeposix to get the Dirent size. I moved that to a constant in types to be in the interface.

I took over here since the other issues were all a big merge because we moved types, and moving some stuff around. At this point it looks 99% good.

My only issue is that the test only confirms that getdents doesn't error, it doesn't really confirm it works correctly. @kliu513  Can you make the test more robust?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1537,4 +1537,80 @@ impl Cage {
...
+                            let position = normalfile_filedesc_obj.position;
+                            let mut bufcount = 0;
+                            let mut curr_size = 0;
+                            let mut count = 0;
+                            let mut temp_len = 0;
+
+                            // iterate over filename-inode pairs in dict
+                            for (filename, inode) in dir_inode_obj.filename_to_inode_dict.clone().into_iter() {
+                                // convert filename to a filename vector of u8
+                                let mut vec_filename: Vec<u8> = filename.as_bytes().to_vec();
```
jesings, line 1575: The filename is not null terminated, so we will need to convert between the rust format and the C format. A very hacky way to do this for now would just be to push a null byte at the end, but that's not entirely correct so you can implement that but put an explanatory comment. Alternatively you could put `Cstring::new` into the interface, and use that here.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1537,4 +1537,80 @@ impl Cage {
...
+                                // push properly constructed tuple to vector storing result
+                                vec.push((interface::ClippedDirent{d_ino: inode as u64, d_off: bufcount as u64, d_reclen: curr_size as u16}, vec_filename));
+                                count += 1;
+                            }
+                            // update file position
+                            normalfile_filedesc_obj.position = interface::rust_min(position + count, dir_inode_obj.filename_to_inode_dict.len());
+                            
+                            interface::pack_dirents(vec, dirp);
+
+                            0 // getdents succeeded
```
jesings, line 1600: The number of bytes written is actually what is returned, not 0,  so you can get that from pack_dirents by subtracting baseptr from the last curptr (using a wrapping_sub probably so as not to need unsafe code) and then casting that to a usize then an i32.

jesings (commented):
This looks fantastic! My only issues are the null termination and the return value. It is also worth noting that you do not actually put 0 in the d_off field for the last element, however the maximum d_off is actually what's returned by getdents rather than 0 and that is used by the user program to figure out what is the last entry.

kliu513:
> This all looked good. You were right that we had to keep the fd mutable, because we had to update position. The only other issue I saw was you used a std library call in safeposix to get the Dirent size. I moved that to a constant in types to be in the interface.
> 
> I took over here since the other issues were all a big merge because we moved types, and moving some stuff around. At this point it looks 99% good.
> 
> My only issue is that the test only confirms that getdents doesn't error, it doesn't really confirm it works correctly. @kliu513 Can you make the test more robust?

I think I have fixed everything else except the test. The problem is that I'm confused about what is supposed to be in the u8 vector if everything is correct? I tried to figure it out based on my implementation of the syscall, but the bunch of type conversions out there are so confusing. If any of you can help me with the test stuff I will really appreciate it!!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1572,7 +1572,7 @@ impl Cage {
                             // iterate over filename-inode pairs in dict
                             for (filename, inode) in dir_inode_obj.filename_to_inode_dict.clone().into_iter() {
                                 // convert filename to a filename vector of u8
-                                let mut vec_filename: Vec<u8> = filename.as_bytes().to_vec();
+                                let mut vec_filename: Vec<u8> = String::from(filename).as_bytes().to_vec();
```
jesings, line 1575: This needs to be an `interface::OsStringKey::from(filename)` as the String::from string is not null terminated, and we need to have a null terminated string to pass back to C.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1537,4 +1537,88 @@ impl Cage {
...
+                                // convert filename to a filename vector of u8
+                                let mut vec_filename: Vec<u8> = String::from(filename).as_bytes().to_vec();
+                                
+                                vec_filename.push(DT_UNKNOWN); // push DT_UNKNOWN as d_type (for now)
+                                temp_len = vec_filename.len(); // get length of current filename vector for padding calculation
+                                
+                                // pad filename vector to the next highest 8 byte boundary
+                                for n in 0..(temp_len + 7) / 8 * 8 - temp_len {
+                                    vec_filename.push(00);
+                                }
```
jesings, line 1583: temp_len should be interface::CLIPPED_DIRENT_SIZE + vec_filename.len() because you're padding the whole structure, not the string, by 8 bytes.

jesings (commented):
Nick ended up changing the function to pack the data inside the buffer within the test. What you should probably do is compare the data at in the buffer to what you expect it to be, so something like:

```rs
let nameoffset = basepointer.wrapping_offset(interface::CLIPPED_DIRENT_SIZE);
let returnedname = std::ffi::CStr::from_ptr(nameoffset as *const u8);
assert_eq(returnedname, std::ffi::Cstr::new("."));
```

As either "." or ".." can come first according to the spec (it's unordered) you may want to assert that one of the 2 are true instead. Additionally assert that the d_off of the first dirent is what you'd expect it to be (24 if I'm not mistaken), and that d_reclen of the first and second dirents are what you'd expect it to be (21 and 22 again if I'm not mistaken). You can find these for the first dirent by casting baseptr to a *const ClippedDirent and then to a &ClippedDirent, and getting the data from there, and likewise for the second dirent, casting baseptr.wrapping_offset(interface::CLIPPED_DIRENT_SIZE) to a *const ClippedDirent and then to an &ClippedDirent and doing likewise there. You may or may not also want to check the second dirent's string (probably unnecessary but you should be able to figure out how to locate it and compare it from here).

Other than my 2 comments which should be like 1 line fixes, everything but the test is great.

src/safeposix/syscalls/fs_calls.rs
kliu513: Ah I tried `interface::OsStringKey::from` and it does not exist. Did you mis-remember where it is located or its name?

jesings: Ok the issue there is not that interface::OsStringKey::from doesn't exist, it's that it doesn't have an as_bytes method. So here's my fix:
Change that line to

`let mut vec_filename: Vec<u8> = interface::RustCStr::from_bytes_with_nul(filename.as_bytes()).unwrap().to_bytes().to_vec();`

And then change the line in interface/file.rs which is currently 

`pub use std::ffi::OsString as OsStringKey;`

to be 

`pub use std::ffi::CStr as RustCStr;`

kliu513: `from_bytes_with_nul()` requires the input slice to be null-terminated. Would it be easier if I just push NULL after I convert the string to the u8 vector?

jesings: Yeah that's probably the simplest solution, and workable at least for now. No need to do anything more complicated than that.

jesings (approved):
Looks ready to merge.
Awesome work!

rennergade (approved):
Cool! Awesome job.

PR #56, opened 2021-08-04
Add truncate() and ftruncate()

kliu513:
Hi guys, please review my implementation of and tests for truncate() and ftruncate(). I had some trouble having extra bytes cut off properly last night. Just fixed it! :)


src/safeposix/dispatcher.rs
```diff
@@ -202,6 +202,12 @@ pub extern "C" fn dispatcher(cageid: u64, callnum: i32, arg1: Arg, arg2: Arg, ar
         RENAME_SYSCALL => {
             check_and_dispatch!(cage.rename_syscall, interface::get_cstr(arg1), interface::get_cstr(arg2))
         }
+        FTRUNCATE_SYSCALL => {
+            cage.ftruncate_syscall(unsafe{arg1.dispatch_int}, unsafe{arg2.dispatch_usize})
+        }
+        TRUNCATE_SYSCALL => {
+            cage.truncate_syscall(unsafe{interface::charstar_to_ruststr(arg1.dispatch_cstr)}, unsafe{arg2.dispatch_usize})
+        }
```
jesings, line 210: Since we added the new dispatcher macro and Tristan's type checking, you should update these to 
`check_and_dispatch!(cage.ftruncate_syscall, interface::get_int(arg1), interface::get_usize(arg2))` and
`check_and_dispatch!(cage.truncate_syscall, interface::get_cstr(arg1), interface::get_usize(arg2))`.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1537,4 +1536,105 @@ impl Cage {
...
+                                let mut vec = vec![0u8; blankbytecount];
+                                let buf: *mut u8 = &mut vec[0];
+
+                                // extend file with null-byte buffer
+                                if let Ok(byteswritten) = fileobject.writeat(buf, blankbytecount, filesize) {
+                                    // update filesize if writeat() succeeded
+                                    normalfile_inode_obj.size = length;
+                                } else {
+                                    panic!("writeat() failed");
+                                }
```
jesings, line 1584: We made an easier function to do this, called zerofillat, you can look at the pwrite syscall to see its usage

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1537,4 +1536,105 @@ impl Cage {
...
+                                    // write the kept part to new file from buffer
+                                    if let Ok(byteswritten) = fileobject.writeat(buf, length, 0) {
+                                        // update filesize if writeat() succeeded
+                                        normalfile_inode_obj.size = length;
+                                    } else {
+                                        panic!("writeat() failed");
+                                    }
+                                } else {
+                                   panic!("readat() failed");
+                                }
```
jesings, line 1612: The fact that we did it like this in repy safeposix is because RePy had no way to truncate the system file as it did not access the truncate syscall. We may not want to keep it this way because it seems ridiculously expensive and not very smart to do it this way. @rennergade what are your thoughts?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1537,4 +1536,105 @@ impl Cage {
...
+            0 // ftruncate() has succeeded!
+        } else { 
+            syscall_error(Errno::EBADF, "ftruncate", "fd is not a valid file descriptor")
+        }
+    }
+
+    //------------------TRUNCATE SYSCALL------------------
+    pub fn truncate_syscall(&self, path: &str, length: usize) -> i32 {
+        self.ftruncate_syscall(self.open_syscall(path, O_RDWR, S_IRWXA), length)
+    }
```
jesings, line 1639: Doing it like this is not really what we want, it seems to have been included as a quick hack in the old safeposix. It would be better if you could do it like stat/fstat where you get the inode number of the file in the one that takes the fd from within the fd, and the inode number of the file in the one that takes the file path using metawalk, and then call from both into a shared helper function. It should be pretty straightforward.

jesings (commented):
Very nicely commented There are a couple things that should be changed (and one that we might want to consider changing) from the python implementation here, but it's already looking great!

src/safeposix/syscalls/fs_calls.rs
rennergade: It's actually fine to keep it this way, because we need an open fobj to do trunc.

src/safeposix/syscalls/fs_calls.rs
rennergade: I've added the function `shrink` to the interface, which is a method for EmulatedFile, this should easily replace all this code so just merge it in.

kliu513:
Hi guys, I think I have fixed everything mentioned.

jesings (approved):
Looks fantastic!

rennergade (approved):
Looks good to me as well.

PR #57, opened 2021-08-04
trunc func

rennergade:
Adding a function to the interface to easily truncate files

### DIFF
--- a/src/interface/file.rs
+++ b/src/interface/file.rs
@@ -133,6 +133,19 @@ impl EmulatedFile {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


jesings (commented):
I think truncfile is a little more general of an interface than we really want and also the name doesn't really fit in with the other emulfile methods. I think naming the method shrink and making sure that the size is smaller than the file's current size might be better? Just being picky this overall is fine and straightforward.

rennergade:
Makes sense. Fixed those issues.

jesings (commented):
Fantastic!

PR #59, opened 2021-08-09
Add pipes()

rennergade:
Adds:

- Pipe interface file w/ emulated pipe + reference counting
- PIPE_TABLE
- adds pipe() system call
- Add reference counting to fork/close etc
- adds in pipes to write/read
- test file for pipes (requires large file, so isnt really set up to do this right now w/o fsutils)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1174,18 +1182,21 @@ impl Cage {
...
+                        pipe.set_eof();
                     }
 
-                    if read_references + write_references == 1 {
-                        //del pipetable.pipenumber
+                    if pipe.get_write_ref() + pipe.get_read_ref() == 0 {
+                        // last reference, lets remove it
+                        PIPE_TABLE.write().unwrap().remove(&pipe_filedesc_obj.pipe).unwrap();
                     }
+
```
jesings, line 1199: Minor, but if you're removing the last reference and it's a write reference, it's maybe worth not setting the eof before then?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1605,4 +1616,43 @@ impl Cage {
...
     }
+
+    //------------------PIPE SYSCALL------------------
+
+    pub fn pipe_syscall(&self, pipefd: &mut PipeArray) -> i32 {
+
+        let mut fdtable = self.filedescriptortable.write().unwrap();
+
+        // get next available pipe number, and set up pipe
+        let pipenumber = get_next_pipe();
```
jesings, line 1627: Check it's not -1 and if it is ENFILE should be returned

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1605,4 +1616,43 @@ impl Cage {
...
+
+        let mut pipetable = PIPE_TABLE.write().unwrap();
+
+        pipetable.insert(pipenumber, interface::RustRfc::new(interface::new_pipe(PIPE_CAPACITY)));
+        
+        // get an fd for each end of the pipe and set flags to RD_ONLY and WR_ONLY
+        // append each to pipefds list
+
+        let flags = [O_RDONLY, O_WRONLY];
+        for flag in flags.iter() {
```
jesings, line 1637: You shouldn't need to do .iter() here I don't think

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1605,4 +1616,43 @@ impl Cage {
...
+
+            let thisfd = if let Some(fd) = self.get_next_fd(None, Some(&fdtable)) {
+                fd
+            } else {
+                return syscall_error(Errno::ENFILE, "pipe", "no available file descriptor number could be found");
```
jesings, line 1642: In this case, remove it from the pipetable

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1605,4 +1616,43 @@ impl Cage {
...
+            };
+
+            let newfd = Pipe(PipeDesc {pipe: pipenumber, flags: *flag, advlock: interface::AdvisoryLock::new()});
+            let wrappedfd = interface::RustRfc::new(interface::RustLock::new(newfd));
+            fdtable.insert(thisfd, wrappedfd);
+
+            match *flag {
```
jesings, line 1649: No need for the deref without .iter

src/tests/pipe_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+        assert_eq!(cage1.close_syscall(filefd), 0);
+
+        assert_eq!(cage1.close_syscall(1), 0);
+
+        sender.join().unwrap();
+
+        assert_eq!(cage1.exit_syscall(), 0);
+
+        println!("Transfered in {} us", now.elapsed().as_micros());
```
jesings, line 109: Timing here is not a great idea--for one the vectors and stuff are allocated within this time, as are file reads (but that's probably what you're going for), and for two tests are not optimized by the compiler. Benchmarks are but they're still a rather bare-bones nightly feature. I'd remove the timer from here for now.

jesings (commented):
Looks pretty much there to me. I couldn't quite figure out the pointer math you were doing in your last test, but it *seemed* ok. The end is certainly in sight!

src/safeposix/syscalls/fs_calls.rs
rennergade: Going to leave this as is. Seems not necessary and is cleaner this way imho.

rennergade:
I fixed the above comments so hopefully is ready?

jesings (approved):
Looks ready to me!

PR #60, opened 2021-08-09
shutdown calls

TristanB22:


TristanB22:
I am not sure whether EINVAL is the right error or not for the new function that I made in the net.rs file. Can someone double check that and see if it is the right one?

TristanB22:
I also still need to add some comments to explain what my code is doing

src/safeposix/net.rs
```diff
@@ -97,6 +97,19 @@ impl NetMetadata {
...
+        }
+        else if protocol ==  IPPROTO_TCP && self.used_port_set.contains(&mux_port(addr.clone(), port, domain, TCPPORT)) {
+            self.used_port_set.remove(&mux_port(addr.clone(), port, domain, TCPPORT));
+        }
+        else {
+            return Err(syscall_error(Errno::EINVAL, "release", "provided port is not being used"));
+        }
+        return Ok(0);
+    }
+
```
jesings, line 112: 2 things: 
1. rather than returning a Result<i32, i32> where the Ok is always 0, return a Result<(), i32>.
2. Rather than check that the port set contains a value, call remove on the value and check whether the remove result is not None.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -538,7 +538,85 @@ impl Cage {
...
+            match &*filedesc_enum {
+                Socket(socketfd) => {
+                    //do nothing besides check that the fd is a socket
+                    //can't do anything with the fd in here because it is being borrowed in the "if let wrapped fd ="" line
+                    //so we just go to the match statement below -- "match how"
+                },
+                _ => {
+                    return syscall_error(Errno::ENOTSOCK, "netshutdown", "file descriptor is not a socket");
+                }
+            }
```
jesings, line 556: You shouldn't match on it then, just say
```rs
if let Socket(_) = filedesc_enum {} else {
    //error out
}
```

src/safeposix/syscalls/net_calls.rs
```diff
@@ -538,7 +538,85 @@ impl Cage {
...
+    pub fn _cleanup_socket(&self, fd: i32, partial: bool, fdtable_lock: Option<&mut FdTable>) -> i32 {
+
+        //pass the lock of the FdTable to this helper. If passed table is none, then create new lock instance
+        let mut writer;
+        let fdtable = if let Some(rl) = fdtable_lock {rl} else {
+            writer = self.filedescriptortable.write().unwrap(); 
+            &mut writer
+        };
+
+        let mut mutmetadata = NET_METADATA.write().unwrap();
```
jesings, line 588: Don't get the metadata until you have to

jesings: So get it within the Socket match arm

src/safeposix/syscalls/net_calls.rs
```diff
@@ -538,7 +538,85 @@ impl Cage {
...
+            }
+        } else {
+            return syscall_error(Errno::EBADF, "listen", "invalid file descriptor");
+        }
+
+        //have to take this out of the match because the fdtable already has a mutable borrow
+        //which means that I can't change it with a remove until after the fdtable mutable borrow is finished from the match statement
+        if !partial {
+            fdtable.remove(&fd); 
+        }
```
jesings, line 614: If you put this at the end of the if block in the match it *should* be able to tell that the other mutable borrow is no longer needed. Try it out, and if it doesn't work keep it this way.

src/safeposix/net.rs
TristanB22: Made it so that it checks whether it is true or not. Please let me know if that is what you wanted.

src/safeposix/syscalls/net_calls.rs
TristanB22: It didn't work. I think that we have to keep it like this.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -538,7 +538,85 @@ impl Cage {
...
+                    let localaddr = sockfdobj.localaddr.as_ref().unwrap().clone();
+                    let release_ret_val = mutmetadata._release_localport(localaddr.addr(), localaddr.port(), sockfdobj.protocol, sockfdobj.domain);
+                    if let Err(e) = release_ret_val {return e;}
+                    if !partial {
+                        mutmetadata.socket_object_table.remove(objectid);
+                        sockfdobj.state = ConnState::NOTCONNECTED;
+                    }
+                }
+                _ => {
+                    return syscall_error(Errno::ENOTSOCK, "listen", "file descriptor refers to something other than a socket");
```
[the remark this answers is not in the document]
TristanB22: Alright, I just made it return 0 since the original one doesn't return anything if the fd is not a socket.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -544,15 +544,13 @@ impl Cage {
...
-                    return syscall_error(Errno::ENOTSOCK, "netshutdown", "file descriptor is not a socket");
-                }
+
+            //in the code below, do nothing besides check that the fd is a socket
+            //We can't do anything with the fd in here because it is being borrowed in the "if let wrapped fd ="" line
+            //so we just go to the match statement below -- "match how"
+
+            if let Socket(_) = &*filedesc_enum {} else {
+                return syscall_error(Errno::ENOTSOCK, "netshutdown", "file descriptor is not a socket");
             }
```
jesings: All this logic is just redone in _cleanup_socket (and errnos are even returned) so there's honestly no need to do it here.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -576,43 +574,34 @@ impl Cage {
...
+                let objectid = &sockfdobj.socketobjectid.unwrap();
+                let localaddr = sockfdobj.localaddr.as_ref().unwrap().clone();
+                let release_ret_val = mutmetadata._release_localport(localaddr.addr(), localaddr.port(), sockfdobj.protocol, sockfdobj.domain);
+                if let Err(e) = release_ret_val {return e;}
+                if !partial {
+                    mutmetadata.socket_object_table.remove(objectid);
+                    sockfdobj.state = ConnState::NOTCONNECTED;
                 }
-            }
+            } else {return 0;}
```
jesings, line 594: You will want to return ENOTSOCK in this case

jesings (commented):
A couple of minor edits and this will be good to go

jesings (commented):
If you merge and then just add the call to _cleanup_socket in close_syscall, this will be ready for merge.

TristanB22:
Had to do some workaround stuff in the fs calls close function. Please take a look and see if there is a better way to do it. I don't think that there is, though/

PR #62, opened 2021-08-11
Adding get name syscalls

TristanB22:
I am a little shaky on the gethostname and getsockname logic. I think that I got it, but if you wouldn't mind focusing on that, that would be awesome. Thanks!

I return the actual structs rather than the individual elements of the addresses right now. I am not sure if that is what you all wanted or not.


src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,77 @@ impl Cage {
...
+                            let addr = interface::GenIpaddr::V4(interface::V4Addr::default());
+                            ret_addr.set_port(0);
+                            ret_addr.set_addr(addr);
+                        }
+        
+                        interface::GenSockaddr::V6(_) => {
+                            let addr = interface::GenIpaddr::V6(interface::V6Addr::default());
+                            ret_addr.set_port(0);
+                            ret_addr.set_addr(addr);
+                        }
```
jesings, line 1280: The family is not set for either of these. Also I don't think we can base our assumptions fairly on what was in the ret_addr before. In order to decide what domain it is, use the information in the socketfdobj.domain field. Feel free to add a set_family method to the GenSockaddr object that sets sin_family and sin6_family.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,77 @@ impl Cage {
...
+        if length < 0 {
+            return syscall_error(Errno::EINVAL, "gethostname", "invalid argument");
+        }
+        if DEFAULT_HOSTNAME.chars().count() > length {
+            address_ptr = &DEFAULT_HOSTNAME[..length].as_bytes();
+        } else {
+            address_ptr = &*DEFAULT_HOSTNAME.as_bytes();
+        }
+        return 0;
+    }
```
jesings, line 1308: This function fills in the data incorrectly. Rather than a mutable slice handle which we can change from within the function, we want to change the data that's pointed to by the slice.
So: `pub fn gethostname(self, length: usize, address_ptr: &mut [u8]) -> i32 {`
And then to change the data pointed to by the address_ptr, use the copy_from_slice function, which you can see here: https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice.

jesings (commented):
Aside from my other two comments, there's one major naming thing that should be fixed. What you'd get from `if let Socket()` should be called a sockfdobj. sockobj is a name that we use for the internal socket objects that we get from the NET_METADATA's socket_object_table. These naming conventions may be changed in the future but for now let's try to keep it that way.

jesings:
If you could actually change the names in getsockopt and setsockopt while you're at it, that would be good.

src/safeposix/syscalls/net_calls.rs
TristanB22: I think that I fixed it. If I didn't, please let me know.

src/safeposix/syscalls/net_calls.rs
TristanB22: Should be done now if I understood the comment correctly.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1295,14 +1291,14 @@ impl Cage {
...
+    pub fn gethostname(self, length: usize, address_ptr: &mut [u8]) -> i32 {
         if length < 0 {
             return syscall_error(Errno::EINVAL, "gethostname", "invalid argument");
         }
         if DEFAULT_HOSTNAME.chars().count() > length {
-            address_ptr = &DEFAULT_HOSTNAME[..length].as_bytes();
+            address_ptr.copy_from_slice(&DEFAULT_HOSTNAME[..length].as_bytes());
         } else {
-            address_ptr = &*DEFAULT_HOSTNAME.as_bytes();
+            address_ptr.copy_from_slice(&DEFAULT_HOSTNAME.as_bytes());
```
jesings: As far as I know, for copy_from_slice, both slices have to be the same size, and this is enforced at runtime. You'll need to get a subslice of address pointer to ensure that.

jesings (commented):
Other than my 1 comment, everything looks great!

jesings (approved):
This looks right to me!

PR #64, opened 2021-08-14
Epoll Net Calls

TristanB22:


TristanB22:
I think that epoll ctl could have some code made less redundant. Am working on that now with some if statements.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+
+                //passing None does the same thing as passing 0 in native Linux
+                //which essentially sets the timeout to the max value allowed (which is almost always more than enough time)
+                if Self::select_syscall(&self, fd, &mut reads, &mut writes, &mut errors, None) > 0 {
+                    mask += if !reads.is_empty() {POLLIN} else {0};
+                    mask += if !writes.is_empty() {POLLOUT} else {0};
+                    mask += if !errors.is_empty() {POLLERR} else {0};
+                    return_code += 1;
+                }
+                structpoll.revents = mask;
```
jesings, line 1341: You need to put the fds you want to look at in the sets. What you do is set a boolean flag if it's POLL_IN etc. but you don't actually put it into the requisite set in that case, which would be necessary for select to return anything.

jesings: Furthermore, just a general comment about programming style, you really pretty much never want an if statement that yields/assigns true and false in the if/else cases, because in all such cases the if statement is unnecessary:
i.e.
`if expression {return true;} else {return false;}` can always be replaced by `return expression;`

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+                let write = if events & POLLOUT > 0 {true} else {false};
+                let err = if events & POLLERR > 0 {true} else {false};
+
+                let mut reads = interface::RustHashSet::<i32>::new();
+                let mut writes = interface::RustHashSet::<i32>::new();
+                let mut errors = interface::RustHashSet::<i32>::new();
+
+                let mut mask: u32 = 0;
+
+                //passing None does the same thing as passing 0 in native Linux
```
jesings, line 1333: This is incorrect. Passing None does the same thing as passing NULL in native which according to the man page for select makes it so that it can block indefinitely. What you want to do is pass a duration of 0, which is equivalent to passing a timespec with all fields zeroed in linux.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+            return newfd;
+        } else {
+            return syscall_error(Errno::ENFILE, "epoll create", "no available file descriptor number could be found");
+        }
+    }
+
+    pub fn _epoll_object_deallocator(&self) -> i32 {
+        //seems to only pass in Repy
+        return 0;
+    }
```
jesings, line 1381: Because the object will be freed when it's dropped from the fdtable, no need to even have this function

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+                    }
+                    if events & EPOLLOUT as u32 > 0 {
+                        structpoll.events |= POLLOUT;
+                    }
+                    if events & EPOLLERR as u32 > 0 {
+                        structpoll.events |= POLLERR;
+                    }
+                    poll_fds_vec.push(structpoll);
+
+                }
```
jesings, line 1479: If you feel like it, you might want to change this into a .map expression (not necessary in the slightest, but would be cool if you wanted to play around with that).

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+
+                let mut poll_fds_slice = &mut poll_fds_vec[..];
+                Self::poll_syscall(&self, &mut *poll_fds_slice, timeout);
```
jesings, line 1482: As far as I can tell, you shouldn't have to dereference and rereference here because you already have a mutable slice above?

jesings: I might be wrong though.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+            } else {
+                return syscall_error(Errno::EBADF, "epoll ctl", "provided fd is not a valid file descriptor")
+            }
+        } else {
+            return syscall_error(Errno::EBADF, "epoll ctl", "provided epoll fd is not a valid epoll file descriptor");
+        }
+        return 0;
+    }
+
+    pub fn epoll_wait_syscall(&self, epfd: i32, maxevents: usize, timeout: Option<interface::RustDuration>) -> i32 {
```
jesings, line 1441: For epoll_wait, take a look at the man page for how to return things. Currently you just populate an array that nothing happens to. What you should do is take in a mutable slice of `struct epoll_event`s that is maxevents long and populate up to maxevents of them.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+                        event.events |= EPOLLIN as u32;
+                    }
+                    if result.revents & POLLOUT > 0 {
+                        event.events |= EPOLLOUT as u32;
+                    }
+                    if result.revents & POLLERR > 0 {
+                        event.events |= EPOLLERR as u32;
+                    }
+                    epoll_return.push(event);
+                }
```
jesings, line 1497: See what I said above about actualy populating something that gets returned.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1234,4 +1233,274 @@ impl Cage {
...
+            } else {
+                return syscall_error(Errno::EINVAL, "epoll wait", "provided fd is not an epoll file descriptor");
+            }
+        } else {
+            return syscall_error(Errno::EBADF, "epoll wait", "provided fd is not a valid file descriptor");
+        }
+        return 0;
```
jesings, line 1504: The return value should be the number of `epoll_event`s populated

src/safeposix/syscalls/net_constants.rs
```diff
@@ -319,6 +350,14 @@ pub struct EpollEvent {
...
 }
 
+#[derive(Debug)]
+#[repr(C)]
+pub struct PollStruct {
+    pub events: u32,
+    pub revents: u32,
+    pub fd: i32 
+}
+
```
jesings, line 360: This and EpollEvent, because they are repr(C), should be defined in types.rs

jesings (commented):
Some changes need to be made but you're definitely on the right track!

rennergade (commented):
I agree with @jesings comments. Also could we merge main into the networking branch so that the getdentcs etc commits dont show as diffs here?

src/safeposix/syscalls/net_calls.rs
TristanB22: I think that I fixed this. Please check it out, though.

src/safeposix/syscalls/net_calls.rs
TristanB22: I will try that once I am done with the tests, I think

TristanB22 (commented):
I think that I solved the issues. Please take a look and let me know what you think. Thank you!

jesings (approved):
Awesome work!

PR #67, opened 2021-08-20
Did some fixes in net_calls for net_tests

jesings:


TristanB22 (approved):
Looks good! Thank you!

PR #78, opened 2021-10-12
Flush stdout on print

rennergade:
Flushes prints to make sure they go through.


rennergade:
Makes sense. This was semantically easier and was going to wait and see how much it affected speed,  but I should just make it flush before the end of programs instead.

PR #80, opened 2021-10-15
add stream to dup helper

rennergade:
Simple add of streams to dup


rennergade:
Also removed the trunc calls from dispatcher here since they're unused and are interfereing

PR #81, opened 2021-10-17
Updated exit() syscall

### LINKED ISSUE #79: exit() syscall incomplete signature
We don't receive the `status` param in the dispatcher here, or send it to the syscall, even though NaCl is sending that through the RPC to keep the signatures the same.

We don't actually use `status` for our purposes, since wait etc are handled currently in NaCl, but it would be good to have the function sigs match even if its unused.

This is a pretty beginner fix so maybe @youngfalcon or @kapkic can submit a PR as a way of learning how to use this side of the system.

youngfalcon:
Please note that the return value of this call is never used anywhere. 

fixes #79

### DIFF
--- a/src/safeposix/dispatcher.rs
+++ b/src/safeposix/dispatcher.rs
@@ -71,6 +71,7 @@ const GETPEERNAME_SYSCALL: i32 = 145;
 use crate::interface;
 use super::cage::{Arg, CAGE_TABLE, Cage, FSData, StatData};
 use super::filesystem::{FS_METADATA, load_fs, incref_root, persist_metadata};
+use super::syscalls::sys_constants::*;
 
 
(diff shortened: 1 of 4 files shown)


src/safeposix/dispatcher.rs
```diff
@@ -229,7 +229,7 @@ pub extern "C" fn lindrustfinalize() {
     let drainedcages: Vec<(u64, interface::RustRfc<Cage>)> = cagetable.drain().collect();
     drop(cagetable);
     for (_cageid, cage) in drainedcages {
-        cage.exit_syscall();
+        cage.exit_syscall(0);
```
rennergade, line 232: I think its worth adding EXIT_SUCCESS (and EXIT_FAILURE) to constants and replacing all these zeros.

PR #84, opened 2021-10-26
UDP Socket tests so far now all work!

jesings:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "rustposix"
 version = "0.1.0"
-authors = ["Nicholas Smith Renner <nrenner@nyu.edu>"]
+authors = ["Nicholas Smith Renner <nrenner@nyu.edu>", "Jonathan Eli Singer <jes965@nyu.edu>"]
 edition = "2018"
```
TristanB22, line 5: Should the edition be changed as well? Also, it would be fun if I could have my name on here as well! I totally understand if not, though, as I am not a leader :).

src/interface/comm.rs
```diff
@@ -98,15 +98,15 @@ pub struct SockaddrV6 {
...
+    raw_sys_fd: i32
 }
 
 impl Socket {
     pub fn new(domain: i32, socktype: i32, protocol: i32) -> Socket {
         let fd = unsafe {libc::socket(domain, socktype, protocol)};
         if fd < 0 {panic!("Socket creation failed when it should never fail");}
-        unsafe {libc::fcntl(fd, libc::F_SETFL, libc::O_NONBLOCK);}
-        Socket {raw_sys_fd: fd}
+        Self {refcnt: 1, raw_sys_fd: fd}
```
TristanB22, line 109: Is this change necessary? I think that this breaks testing on MacOS.

src/interface/comm.rs
```diff
@@ -122,10 +122,7 @@ impl Socket {
             GenSockaddr::V6(addrref6) => {((addrref6 as *const SockaddrV6).cast::<libc::sockaddr>(), size_of::<SockaddrV6>())}
             GenSockaddr::V4(addrref) => {((addrref as *const SockaddrV4).cast::<libc::sockaddr>(), size_of::<SockaddrV4>())}
         };
-        unsafe {libc::fcntl(self.raw_sys_fd, libc::F_SETFL, 0);}
-        let f = unsafe {libc::connect(self.raw_sys_fd, finalsockaddr, addrlen as u32)};
-        unsafe {libc::fcntl(self.raw_sys_fd, libc::F_SETFL, libc::O_NONBLOCK);}
-        f
+        unsafe {libc::connect(self.raw_sys_fd, finalsockaddr, addrlen as u32)}
```
TristanB22, line 125: Same comment about fcntl

src/safeposix/net.rs
```diff
@@ -43,11 +41,24 @@ pub const TCPPORT: bool = true;
 pub const UDPPORT: bool = false;
 
 impl NetMetadata {
+    fn port_in_use(&self, tup: &(interface::GenIpaddr, u16, PortType)) -> bool {
```
TristanB22, line 44: Would this function be helpful to add to the creation functions or sockets? It could clean up some code -- not sure though.

src/tests/networking_tests.rs
```diff
@@ -7,21 +7,23 @@ pub mod net_tests {
...
-        ut_lind_net_listen();
-        ut_lind_net_recvfrom(); 
-        ut_lind_net_shutdown();
+        //ut_lind_net_listen();
+        //ut_lind_net_recvfrom(); 
+        //ut_lind_net_shutdown();
         ut_lind_net_socket();
         ut_lind_net_socketoptions();
+        ut_lind_net_udp_bad_bind();
+        ut_lind_net_udp_simple(); //not working right now
```
TristanB22, line 21: The not working right now comment can probably be removed if it is working now. And are the tests supposed to be commented out?

TristanB22 (commented):
It looks good! Just a few, minor comments.

src/interface/comm.rs
jesings: This shouldn't because as per the meeting last week we need not do the nonblocking behavior at all, so not doing it in either fcntl or in the send/recv calls would be fine for mac.

src/safeposix/net.rs
jesings: This code is specifically for the _get_available_x port because instead of just needing to check the used port set for that device, we also need to check the used_port_set for the catch-all device ::0 or 0.0.0.0. For other code, checking the used_port_set suffices and is reasonably clean.

src/tests/networking_tests.rs
jesings: Ok yeah. The other tests are supposed to be commented out because TCP isn't quite working yet, that's what I'm going to be working on next.

Cargo.toml
rennergade: Yeah @TristanB22 you can add your name on your next PR.

rennergade:
Approved.

PR #85, opened 2021-11-01
Keep advisory lock on duplicate instead of creating new one

rennergade:
New advisory locks were being created on fork/dup which is incorrect according to manpage:

```  Locks  created  by flock() are associated with an open file description (see open(2)).  This means that duplicate file descriptors (created by, for example, fork(2) or dup(2)) refer to the same lock, and this lock may be modified or released using any of these
       file descriptors.  Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate file descriptors, or when all such file descriptors have been closed.

       If a process uses open(2) (or similar) to obtain more than one file descriptor for the same file, these file descriptors are treated independently by flock().  An attempt to lock the file using one of these file descriptors may be denied by  a  lock  that  the
       calling process has already placed via another file descriptor```

This PR fixes that.

### DIFF
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -125,9 +125,3 @@ impl AdvisoryLock {
         } else {false}
     }
 }
-
-impl Clone for AdvisoryLock {
-    fn clone(&self) -> Self {
-        AdvisoryLock::new()
-    }
-}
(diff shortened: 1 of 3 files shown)


jesings (approved):
Super simple, approved!

PR #88, opened 2021-12-20
Consolidate pipe len reads

rennergade:
Pairs with pipe IO optimization PR to speed up pipe reads

### DIFF
--- a/src/interface/pipe.rs
+++ b/src/interface/pipe.rs
@@ -94,8 +94,9 @@ impl EmulatedPipe {
         let mut read_end = self.read_end.lock().unwrap();
 
         while bytes_read < length {
-            if (read_end.len() == 0) & self.eof.load(Ordering::Relaxed) { break; }
-            let bytes_to_read = min(length, bytes_read + read_end.len());
+            let pipe_space = read_end.len();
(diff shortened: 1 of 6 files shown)


src/interface/pipe.rs
```diff
@@ -94,8 +94,9 @@ impl EmulatedPipe {
         let mut read_end = self.read_end.lock().unwrap();
 
         while bytes_read < length {
-            if (read_end.len() == 0) & self.eof.load(Ordering::Relaxed) { break; }
-            let bytes_to_read = min(length, bytes_read + read_end.len());
+            let pipe_space = read_end.len();
+            if (pipe_space == 0) & self.eof.load(Ordering::Relaxed) { break; }
+            let bytes_to_read = min(length, bytes_read + pipe_space);
```
jesings, line 99: This was *probably* compiler optimized out but it's good to have it this way if for nothing else except clarity. Looks good!

jesings (approved):

src/interface/pipe.rs
rennergade: It actually resulted in a significant speed increase, so afik it wasn't?

PR #89, opened 2021-12-28
Stat fix

rennergade:
Pairs with NaCl PR, fixes struct ordering.

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -397,11 +397,10 @@ impl Cage {
     }
 
     fn _istat_helper(inodeobj: &GenericInode, statbuf: &mut StatData) {
-        statbuf.st_nlink = inodeobj.linkcount;
         statbuf.st_mode = inodeobj.mode;
+        statbuf.st_nlink = inodeobj.linkcount;
         statbuf.st_uid = inodeobj.uid;
         statbuf.st_gid = inodeobj.gid;
-        statbuf.__pad0 = 0;
         statbuf.st_rdev = 0;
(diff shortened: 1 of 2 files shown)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -422,12 +420,11 @@ impl Cage {
 
     fn _istat_helper_chr_file(inodeobj: &DeviceInode, statbuf: &mut StatData) {
         statbuf.st_dev = 5;
-        statbuf.st_nlink = inodeobj.linkcount as u64;
         statbuf.st_mode = inodeobj.mode;
+        statbuf.st_nlink = inodeobj.linkcount;
         statbuf.st_uid = inodeobj.uid;
         statbuf.st_gid = inodeobj.gid;
         //compose device number into u64
-        statbuf.__pad0 = 0;
```
JustinCappos, line 430: How does this change relate?

rennergade: The `lind_stat` struct that NaCl's trusted side was using to store the stat buffer mistakenly listed `nlink` as 64 bits, even though its only 32 bits. This pad seems to have been added to fix alignment when nlink is 64 bits, but is unnecessary when fixed to 32 bits.

It's being removed here in RustPOSIX to correspond to that change in NaCl.

jesings (commented):
Huh... so we don't have nlink at all?

rennergade:
@jesings I'm not sure where you're seeing that. It's moved below `mode` where it is supposed to be.

jesings:
Oops, I'm just blind. Looks good!

youngfalcon (approved):

PR #90, opened 2021-12-28
Final fixes for currently written network tests (ipv4)

jesings:


src/safeposix/dispatcher.rs
```diff
@@ -398,18 +398,18 @@ pub extern "C" fn lindrustfinalize() {
...
-        dispatcher(1, FORK_SYSCALL, Arg {dispatch_ulong: 2_u64}, Arg {dispatch_int: 34132}, Arg {dispatch_int: 109384}, Arg {dispatch_int: -12341}, Arg {dispatch_int: -12341}, Arg {dispatch_int: 0});
-        {println!("{:?}", CAGE_TABLE.read().unwrap());};
-        dispatcher(2, EXEC_SYSCALL, Arg {dispatch_ulong: 7_u64}, Arg {dispatch_int: 34132}, Arg {dispatch_int: 109384}, Arg {dispatch_int: -12341}, Arg {dispatch_int: -12341}, Arg {dispatch_int: 0});
-        {println!("{:?}", CAGE_TABLE.read().unwrap());};
-        dispatcher(7, EXIT_SYSCALL, Arg {dispatch_ulong: 61_u64}, Arg {dispatch_int: 33987}, Arg {dispatch_int: 123452}, Arg {dispatch_int: -98493}, Arg {dispatch_int: -1}, Arg {dispatch_int: 0});
-        {println!("{:?}", CAGE_TABLE.read().unwrap());};
-
-        lindrustfinalize();
+        //lindrustinit();
+
```
rennergade, line 402: Should this all be commented out? Not sure what the testing strategy is here.

src/safeposix/syscalls/net_constants.rs
```diff
@@ -237,25 +237,34 @@ pub const SHUT_RDWR: i32 = 2;
 
 ////////////////////// setsockopt / getsockopt...
 pub const SOL_SOCKET: i32 = 1;
```
rennergade, line 239: are SOL_SOCKET and SO_DEBUG both supposed to be 1?

src/safeposix/net.rs
```diff
@@ -27,51 +51,124 @@ pub fn mux_port(addr: interface::GenIpaddr, port: u16, domain: i32, istcp: bool)
...
+        }
         let mut porttuple = mux_port(addr.clone(), 0, domain, TCPPORT);
-        for port in (EPHEMERAL_PORT_RANGE_START ..= EPHEMERAL_PORT_RANGE_END).rev().map(|x| x.to_be()) {
-            porttuple.1 = port;
-            if !self.port_in_use(&porttuple) {
-                self.used_port_set.insert(porttuple);
-                return Ok(port);
+
+        //start from the starting location we specified in a previous attempt to get an ephemeral port
+        let next_ephemeral = if domain == AF_INET {self.next_ephemeral_port_tcpv4} else if domain == AF_INET6 {self.next_ephemeral_port_tcpv6} else {unreachable!()};
```
rennergade, line 147: This code block and the above block for udp seem extremely similar? Does it make sense to condense this into a function?

src/interface/comm.rs
```diff
@@ -198,10 +256,26 @@ impl Socket {
             }
         };
     }
+
+    pub fn setsockopt(&self, level: i32, optname: i32, optval: i32) -> i32 {
```
rennergade, line 260: Looking through this, setsockopt doesn't have the same args as lind_setsockopt? How does the optval get changed from a pointer with a length to an i32?

rennergade (commented):
Left a few comments as a first pass. Overall is very impressive. 

Will do another pass after others review.

src/safeposix/dispatcher.rs
jesings: Yeah so this stuff is a test which was useful at the very beginning just to check that the dispatcher was working at all but now I think our C tests work for that, so I will remove this one.

src/safeposix/net.rs
jesings: They do but it's not that easy as if you look inside the loop they have to use different members (next_ephemeral_port_tcpv4/v6 or next_ephemeral_port_udpv4/v6)

src/safeposix/syscalls/net_constants.rs
jesings: Yep
https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/socket.h#L9

src/interface/comm.rs
jesings: This is the libc wrapper, this has nothing to do with what's passed in lind_setsockopt. What you are referring to is the fact that I am not packaging things up in the dispatcher which is the case for both get and set sockopt.  I will fix these now.

jesings: That has been fixed with the latest PR

rennergade: Yeah, that's what I meant. Cool!

src/safeposix/syscalls/net_calls.rs
```diff
@@ -639,13 +664,15 @@ impl Cage {
...
                 if let Some(localaddr) = sockfdobj.localaddr.as_ref().clone() {
-                    let release_ret_val = mutmetadata._release_localport(localaddr.addr(), localaddr.port(), sockfdobj.protocol, sockfdobj.domain);
+                    let release_ret_val = mutmetadata._release_localport(localaddr.addr(), localaddr.port(), sockfdobj.protocol, sockfdobj.domain, fd);
                     if let Err(e) = release_ret_val {return e;}
                 }
                 if !partial {
                     if let None = objectid {} else {
                         mutmetadata.socket_object_table.remove(&objectid.unwrap());
                     }
+                    //hmm but what if there are multiple fds referring to the same socket that need
```
rennergade, line 674: You left yourself a note here. Did this get clarified?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -667,131 +694,113 @@ impl Cage {
...
-                                    let wrappedsock = interface::RustRfc::new(interface::RustLock::new(Socket(newsockobj)));
+                            *addr = remote_addr; //populate addr with what address it connected to
+                            let domain = sockfdobj.domain;
 
-                                    let newfd = if let Some(fd) = self.get_next_fd(None, Some(&fdtable)) {
-                                        fd
-                                    } else {
-                                        return syscall_error(Errno::ENFILE, "accept", "no available file descriptor number could be found");
-                                    };
+                            //THIS DEADLOCKS IF WE DON'T DROP THE FDTABLE HERE...
```
rennergade, line 779: You also left this deadlock note, but I don't see the explicit drop? 

You do use drop explicitly a bunch in this PR, which I thought is sort of frowned upon in Rust?

rennergade (commented):
Generally looks really good. Would like the others to look through as well before merging.

src/safeposix/syscalls/net_calls.rs
jesings: Drop does make locks no longer last the full scope. The note is leftover from when we did manual blocking so I will remove it.

jesings: Maybe we can figure out more of best practices and better ways to do things when we do the merge into main?

src/safeposix/syscalls/net_calls.rs
jesings: Yeah I ended up fixing this a while ago.

src/safeposix/net.rs
```diff
@@ -27,51 +51,124 @@ pub fn mux_port(addr: interface::GenIpaddr, port: u16, domain: i32, istcp: bool)
...
+                            self.next_ephemeral_port_udpv4 = EPHEMERAL_PORT_RANGE_END;
+                        } else if domain == AF_INET6 {
+                            self.next_ephemeral_port_udpv6 = EPHEMERAL_PORT_RANGE_END;
+                        } else {unreachable!()};
+                    } else {
+                        if domain == AF_INET {
+                            self.next_ephemeral_port_udpv4 = ne_port - 1;
+                        } else if domain == AF_INET6 {
+                            self.next_ephemeral_port_udpv6 = ne_port - 1;
+                        } else {unreachable!()};
```
TristanB22, line 131: This is just a small thing, but these unreachable code lines feel unnecessary -- we are just getting the info straight from the Muxport function and the Muxport function has a panic option for unknown inputs. I guess it is safer to not assume that the data is good, but this feels superfluous.

src/safeposix/net.rs
```diff
@@ -27,51 +51,124 @@ pub fn mux_port(addr: interface::GenIpaddr, port: u16, domain: i32, istcp: bool)
...
+                            self.next_ephemeral_port_tcpv4 = EPHEMERAL_PORT_RANGE_END;
+                        } else if domain == AF_INET6 {
+                            self.next_ephemeral_port_tcpv6 = EPHEMERAL_PORT_RANGE_END;
+                        } else {unreachable!()};
+                    } else {
+                        if domain == AF_INET {
+                            self.next_ephemeral_port_tcpv4 = ne_port - 1;
+                        } else if domain == AF_INET6 {
+                            self.next_ephemeral_port_tcpv6 = ne_port - 1;
+                        } else {unreachable!()};
```
TristanB22, line 167: More of the same as above

src/safeposix/syscalls/net_constants.rs
```diff
@@ -237,25 +237,34 @@ pub const SHUT_RDWR: i32 = 2;
...
+pub const SO_BSDCOMPAT: i32 = 14;
+pub const SO_REUSEPORT: i32 = 15;
+pub const SO_PASSCRED: i32 = 16;
+pub const SO_PEERCRED: i32 = 17;
+pub const SO_RCVLOWAT: i32 = 18;
+pub const SO_SNDLOWAT: i32 = 19;
+pub const SO_RCVTIMEO_OLD: i32 = 20;
+pub const SO_SNDTIMEO_OLD: i32 = 21;
+pub const SO_PEERNAME: i32 = 28;
+pub const SO_ACCEPTCONN: i32 = 30;
```
TristanB22, line 265: Did we get the issue where the & and | is causing issues with these numerical values solved? I got the values that we are deleting from Linux I believe. But, since we were treating some of these values as bitfields, it was causing problems.

src/tests/networking_tests.rs
```diff
@@ -720,7 +745,7 @@ pub mod net_tests {
             println!("RECV");
```
TristanB22, line 745: Can get rid of this if the tests are working now -- this was just for me to check when deadlocks were happening.

src/tests/mod.rs
```diff
@@ -19,7 +19,7 @@ mod main_tests {
     #[test]
     pub fn tests() {
         lindrustinit();
-        crate::lib_fs_utils::lind_deltree({CAGE_TABLE.read().unwrap().get(&1).unwrap()}, "/");
+        crate::lib_fs_utils::lind_deltree(CAGE_TABLE.read().unwrap().get(&1).unwrap(), "/");
         lindrustfinalize();
 
         println!("FS TESTS");
```
TristanB22, line 25: Not necessary

src/tests/networking_tests.rs
TristanB22: The same thing on line 736

TristanB22 (commented):
Looking pretty good -- I didn't find anything that was functionally wrong with this code.

src/safeposix/syscalls/net_calls.rs
rennergade: Sure. Just looked into it, I was thinking of something else ie calling the underlying drop mechanism. Calling drop like this seems fine.

kapkic (approved):
Looks good!

src/safeposix/syscalls/net_constants.rs
jesings: Yes I sorted that out by having these numbers be the bit shift amount

src/safeposix/net.rs
jesings: I think it's helpful in case we want to implement unix domain sockets later so we can pinpoint failures

src/tests/mod.rs
jesings: I thought this was for the fs tests so that we could count the files that we created. I removed the curly braces because they're unnecessary and add clutter, but I think this line is necessary.

rennergade: Does this delete the entire filesystem whenever tests are run?

TristanB22: Yes — otherwise the persisted file system screws with the files and tests

jesings: We discussed this over the summer. This won't clobber the lind file system, just the dummy one set up within the rust dirs for testing.

rennergade (approved):

PR #91, opened 2021-12-29
Getcwd

youngfalcon:
Implemented the getcwd syscall.

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1730,12 +1730,11 @@ impl Cage {
     //------------------------------------GETCWD SYSCALL------------------------------------
     
     pub fn getcwd_syscall(&self, buf: *mut u8, bufsize: u32) -> i32 {
-        let cwd: String = format!("{}{}", self.cwd.read().unwrap().to_string_lossy(), "\0");
-        let bytes: Vec<u8> = cwd.as_bytes().to_vec();
+        let mut bytes: Vec<u8> = self.cwd.read().unwrap().to_string_lossy().as_bytes().to_vec();
+        bytes.push(0u8);
         let length = bytes.len();
 
-        use std::convert::TryInto;
-        if bufsize < length.try_into().unwrap() {
+        if (bufsize as usize) < length {
             return syscall_error(Errno::ERANGE, "getcwd", "the length (in bytes) of the absolute pathname of the current working directory exceeds the given size");
         }
         
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1736,8 +1736,8 @@ impl Cage {
         if bufsize < (cwd.len() + 1).try_into().unwrap() {
             return syscall_error(Errno::ERANGE, "getcwd", "the length (in bytes) of the absolute pathname of the current working directory exceeds the given size");
         }
-
-        *buf = CString::new(cwd).unwrap();
+        
+        *buf = std::ffi::CString::new(cwd).unwrap();
         0 //getcwd has succeeded!;
     }
 }
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1726,7 +1726,7 @@ impl Cage {
     //------------------------------------GETCWD SYSCALL------------------------------------
     
     pub fn getcwd_syscall(&self, buf: *mut u8, bufsize: u32) -> i32 {
-        let mut bytes: Vec<u8> = self.cwd.read().unwrap().as_path().display().to_string().as_bytes().to_vec();
+        let mut bytes: Vec<u8> = self.cwd.read().unwrap().to_str().unwrap().as_bytes().to_vec();
         bytes.push(0u8);
         let length = bytes.len();
 
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1731,8 +1731,9 @@ impl Cage {
     
     pub fn getcwd_syscall(&self, buf: *mut u8, bufsize: u32) -> i32 {
         let cwd = format!("{}{}", self.cwd.read().unwrap().into_os_string().into_string().unwrap(), "\0");
-
-        if bufsize < (cwd.len() + 1).std::convert::try_into().unwrap() {
+        
+        use std::convert::TryInto;
+        if bufsize < (cwd.len() + 1).try_into().unwrap() {
             return syscall_error(Errno::ERANGE, "getcwd", "the length (in bytes) of the absolute pathname of the current working directory exceeds the given size");
         }
 
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1730,7 +1730,7 @@ impl Cage {
     //------------------------------------GETCWD SYSCALL------------------------------------
     
     pub fn getcwd_syscall(&self, buf: *mut u8, bufsize: u32) -> i32 {
-        let cwd: String = format!("{}{}", (*(self.cwd.read().unwrap())).clone().into_os_string().into_string().unwrap(), "\0");
+        let cwd: String = format!("{}{}", self.cwd.read().unwrap().to_string_lossy(), "\0");
         let bytes: Vec<u8> = cwd.as_bytes().to_vec();
         let length = bytes.len();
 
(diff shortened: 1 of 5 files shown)


kapkic:
Looks good

kapkic (approved):

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1726,4 +1726,22 @@ impl Cage {
             syscall_error(Errno::EBADF, "getdents", "Invalid file descriptor")
         }
     }
+
+    //------------------------------------GETCWD SYSCALL------------------------------------
+    
+    pub fn getcwd_syscall(&self, buf: *mut u8, bufsize: u32) -> i32 {
+        let cwd: String = format!("{}{}", self.cwd.read().unwrap().to_string_lossy(), "\0");
+        let bytes: Vec<u8> = cwd.as_bytes().to_vec();
```
jesings, line 1734: I don't think this is a very good implementation. We'd prefer to_string to to_string_lossy. Then you'd just have an &str which you can use as_bytes on directly. without needing to format anything. To add the terminating null byte, just push a 0 onto the stack.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1726,4 +1726,22 @@ impl Cage {
...
+        let length = bytes.len();
+
+        use std::convert::TryInto;
```
jesings, line 1737: We try very hard to import as little as possible even from the rust standard library and when we do we make sure to do it through the interface.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1726,4 +1726,22 @@ impl Cage {
...
+        if bufsize < length.try_into().unwrap() {
```
jesings, line 1738: to solve this problem all you need to do is do `bufsize as usize`, no need for any fancy try_into

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1726,4 +1726,22 @@ impl Cage {
...
+            return syscall_error(Errno::ERANGE, "getcwd", "the length (in bytes) of the absolute pathname of the current working directory exceeds the given size");
+        }
+        
+        let slice = unsafe{std::slice::from_raw_parts_mut(buf, length)};
```
jesings, line 1742: We additionally don't use any unsafe code except for in the interface, so this should probably be handled into the dispatcher. I'd say that this function should probably take in an `&mut [u8]`, which you can get from `get_slice_from_string`, which you should use in the dispatcher. Check the implementation of Epoll for something similar we did, although the slice was not of u8s, it should be relatively easy to understand how to transform these things.

src/tests/fs_tests.rs
```diff
@@ -868,7 +869,8 @@ mod fs_tests {
         
         assert_eq!(cage.mkdir_syscall("/getdents", S_IRWXA), 0);
         let fd = cage.open_syscall("/getdents", O_RDWR, S_IRWXA);
-        assert_eq!(cage.getdents_syscall(fd, baseptr, bufsize), 48);
+        use std::convert::TryInto;
+        assert_eq!(cage.getdents_syscall(fd, baseptr, bufsize.try_into().unwrap()), 48);
```
jesings, line 873: Rather than use try_into, use an explicit cast

src/tests/fs_tests.rs
```diff
@@ -889,4 +891,34 @@ mod fs_tests {
...
         lindrustfinalize();
     }
+
+    pub fn ut_lind_fs_dir_chdir_getcwd() {
+        lindrustinit();
+        let cage = {CAGE_TABLE.read().unwrap().get(&1).unwrap().clone()};
+        let needed = "/subdir1\0".as_bytes().to_vec().len();
+
+        use std::convert::TryInto;
+        let needed_u32: u32 = needed.try_into().unwrap();
```
jesings, line 901: Again here just use `as u32`

src/tests/fs_tests.rs
```diff
@@ -889,4 +891,34 @@ mod fs_tests {
...
+        assert_eq!(String::from_utf8_lossy(&buf), "/\0\0\0\0\0\0\0\0");
+
+        cage.mkdir_syscall("/subdir1", S_IRWXA);
+        assert_eq!(cage.access_syscall("subdir1", F_OK), 0);
+        assert_eq!(cage.chdir_syscall("subdir1"), 0);
+
+        assert_eq!(cage.getcwd_syscall(bufptr, 0), -(Errno::ERANGE as i32));
+        assert_eq!(cage.getcwd_syscall(bufptr, needed_u32-1), -(Errno::ERANGE as i32));
+        assert_eq!(cage.getcwd_syscall(bufptr, needed_u32), 0);
+        assert_eq!(String::from_utf8_lossy(&buf), "/subdir1\0");
```
jesings, line 919: You can/should probably use the non lossy versions of these.

src/safeposix/syscalls/fs_calls.rs
jesings: The epoll implementations are in the sockremoval_better branch

jesings (commented):
Overall a very impressive first contribution to rustposix. There are a few changes that should be made before this is merge ready, which I've commented on above.

src/tests/fs_tests.rs
youngfalcon: Why do we not want the lossy versions? Is a panic preferable to an un-exact string?

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Fixed with the last update.

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Fixed with the last update.

src/tests/fs_tests.rs
youngfalcon: Fixed with the last update.

src/tests/fs_tests.rs
youngfalcon: Fixed with the last update.

src/tests/fs_tests.rs
jesings: In the tests we know that you're passing/getting passed valid utf8, so it won't return an error variant (which I assume is what you mean by panic as that would panic on an unwrap). Additionally, the lossy and non-lossy versions actually do somewhat different things--the non-lossy version allocates a new buffer whereas the lossy version does it in place and wraps it within a Cow<> object which makes it copy-on-write. This means it's somewhat more complex and annoying to deal with, and we don't care about the overhead of the utf8 validity check within the test.

youngfalcon: Updated this to use std::str::from_utf8.

(String::from_utf8 does not work directly as it requires the ownership of the passed vector.)

rennergade:
I've added a function to the interface to convert mut buf ptr's with sizes to mut slices. You can merge from the branch `addmutslice` to gain that function here.

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Updated this with the last few commits.

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Updated.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1722,4 +1722,20 @@ impl Cage {
             syscall_error(Errno::EBADF, "getdents", "Invalid file descriptor")
         }
     }
+
+    //------------------------------------GETCWD SYSCALL------------------------------------
+    
+    pub fn getcwd_syscall(&self, buf: *mut u8, bufsize: u32) -> i32 {
+        let mut bytes: Vec<u8> = self.cwd.read().unwrap().to_str().unwrap().as_bytes().to_vec();
+        bytes.push(0u8);
```
rennergade, line 1730: Being pedantic but would love a short comment here about null terminating.

rennergade (approved):
This looks really great now. Succinct and functional.

Just add that one comment and I'll merge it.

src/safeposix/syscalls/fs_calls.rs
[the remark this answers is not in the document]
youngfalcon: This issue was fixed in previous commits.

src/safeposix/syscalls/fs_calls.rs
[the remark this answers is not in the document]
youngfalcon: This issue was fixed in previous commits.

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Updated.

PR #93, opened 2022-01-05
Some fixes for the networking subsystem, serverclient.c now works, some warnings fixed

jesings:


src/safeposix/syscalls/net_calls.rs
```diff
@@ -603,7 +605,7 @@ impl Cage {
                             if let None = sockfdobj.localaddr {
                                 let bindret = sockobj.bind(&ladr);
                                 if bindret < 0 {
-                                    let sockerrno = match Errno::from_discriminant(unsafe{*libc::__errno_location()} as i32) {
+                                    match Errno::from_discriminant(unsafe{*libc::__errno_location()} as i32) {
```
rennergade, line 608: This seems to do some stuff that we want via the interface.

rennergade (commented):
This mostly looks good, one comment.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -737,7 +739,7 @@ impl Cage {
                                 }
                             };
 
-                            if let Err(errval) = acceptedresult {
+                            if let Err(_) = acceptedresult {
                                 match Errno::from_discriminant(unsafe{*libc::__errno_location()} as i32) {
```
TristanB22, line 743: Same as Nick's comment above

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1318,11 +1322,6 @@ impl Cage {
                     return syscall_error(Errno::EINVAL, "epoll wait", "max events argument is not a positive number");
                 }
 
-                //now we know that all of the arguments are valid...
-                let mut reads = interface::RustHashSet::<i32>::new();
-                let mut writes = interface::RustHashSet::<i32>::new();
-                let mut errors = interface::RustHashSet::<i32>::new();
```
TristanB22, line 1324: I think that these are in the original system call from Repy... are they unnecessary when using Rust?

TristanB22 (approved):

kapkic (approved):
Looks good.

src/safeposix/syscalls/net_calls.rs
youngfalcon: Agreed, since we want to restrict unsafe code to the interface.

youngfalcon (commented):
This looks good in general, although moving the unsafe code portion mentioned by @rennergade and @TristanB22  might be ideal.

src/safeposix/syscalls/net_calls.rs
jesings: Yeah they're not actually used in poll. I think they were unused in repy as well, but probably dated back from when poll was handled separately from select in repy.

youngfalcon (approved):

rennergade:
I think this is good to go now!

PR #94, opened 2022-01-07
Fixed the gethostname syscall

youngfalcon:
I took an approach similar to getcwd and used the fill function (copied here with a small fix to allow for length < hostname_length). We should choose to opt for this implementation of the fill function instead of the one in develop when merging networking to develop.

In POSIX, it is not specified whether the returned string should have the null terminator or not when truncating happens. I chose to not include it just like our safeposix implementation.

### DIFF
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -55,7 +55,7 @@ pub fn fillzero(bufptr: *mut u8, count: usize) -> i32 {
 }
 pub fn fill(bufptr: *mut u8, count: usize, values:&Vec<u8>) -> i32 {
     let slice = unsafe{std::slice::from_raw_parts_mut(bufptr, count)};
-    slice.copy_from_slice(&values);
+    slice.copy_from_slice(&values[..count]);
     count as i32
 }
 
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -53,6 +53,11 @@ pub fn fillzero(bufptr: *mut u8, count: usize) -> i32 {
     for i in 0..count {slice[i] = 0u8;}
     count as i32
 }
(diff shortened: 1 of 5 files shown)


src/interface/misc.rs
```diff
@@ -53,6 +53,11 @@ pub fn fillzero(bufptr: *mut u8, count: usize) -> i32 {
     for i in 0..count {slice[i] = 0u8;}
     count as i32
```
rennergade, line 54: Did you implement this a second time? I think what we want to do is remerge the develop updates into networking.

rennergade (commented):
Small merge concern, other than that looks good!

src/interface/misc.rs
youngfalcon: Yes, I re-implemented fill here (although I made a small fix to it that only matters if the given length is shorter than the length of the values vector.

youngfalcon: I was not sure if merging develop into this is a good idea since we were developing the networking branch separately and will merge this into develop later on. 

Please let me know if you would still like me to do that. Otherwise, I can handle this merge issue when merging networking to develop on the GitHub interface in a few seconds.

rennergade: @jesings Can we merge develop into networking to update?

kapkic (approved):
Looks good!

src/interface/misc.rs
jesings: Ok develop is merged into networking @youngfalcon now you should merge networking into this and then remove this reimplementation.

jesings (commented):
Other than the merge and removal of the fill function this looks good!

PR #97, opened 2022-01-19
Socketpair, select, poll fixes

jesings:


jesings:
Do note that TCP socketpair is not the easiest to implement well, because we have no way of knowing whether the listening socket is actually ready to accept yet so we have to rely on just waiting a fixed amount of time and assuming that is good enough.

rennergade (commented):
I'll do a more thorough review in a bit but on first glance there seems to be a random binary and an empty test case which seem odd.

jesings:
That must have been from tristan, it's fixed now.

TristanB22 (approved):
Cool! Thanks for fixing the system calls -- I wasn't super sure how they were intricately working since I don't believe I wrote them. I also see why my code wasn't working now.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1425,9 +1418,13 @@ impl Cage {
...
+            let spin_val = interface::RustRfc::new(interface::RustAtomicBool::new(false));
+            let mut garbage_remote = portlessaddr.clone();
             let thishandle2 = this.clone();
+            
+            let sub_spin_val = spin_val.clone();
             let acceptor = interface::helper_thread(move || {
+                sub_spin_val.store(true, interface::RustAtomicOrdering::Relaxed);
                 let accret = thishandle2.accept_syscall(sock1fd, &mut garbage_remote);
                 if accret < 0 {
                     let sockerrno = match Errno::from_discriminant(unsafe{*libc::__errno_location()} as i32) {
```
rennergade, line 1430: I know this wasn't updated but weren't we not going to use libc crate in the safeposix code?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1442,7 +1439,9 @@ impl Cage {
                 return accret;
             });
     
-            let connret = this.connect_syscall(sock2fd, &addr);
+            while !spin_val.load(interface::RustAtomicOrdering::Relaxed) {}
+            interface::sleep(interface::RustDuration::from_micros(100)); //this is a very bad way of handling it but I don't see a better one for now?
+            let connret = this.connect_syscall(sock2fd, &bound_addr);
             if connret < 0 {
                 let sockerrno = match Errno::from_discriminant(unsafe{*libc::__errno_location()} as i32) {
```
rennergade, line 1446: again

rennergade (commented):
Mostly looks good to me. Let's talk about the tcp connection wait tonight.

src/safeposix/syscalls/net_calls.rs
jesings: Yeah I forgot to merge from networking, that's fixed now.

kapkic (approved):
LGTM!

PR #101, opened 2022-01-27
Link RustPOSIX verbosity to Lind

rennergade:
enables boolean set by lindrustinit if verbose mode is set.

### DIFF
--- a/src/interface/errnos.rs
+++ b/src/interface/errnos.rs
@@ -2,8 +2,14 @@
 // Error handling for SafePOSIX
 use crate::interface;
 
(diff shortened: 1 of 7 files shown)


jesings (commented):
I can't say I really think this makes very much sense as an atomic bool. I think it would make a lot more sense as a synconcecell but whether or not we want to use those (they're part of the same rust library submodule as synclazy) is another question

rennergade:
> I can't say I really think this makes very much sense as an atomic bool. I think it would make a lot more sense as a synconcecell but whether or not we want to use those (they're part of the same rust library submodule as synclazy) is another question

Ok that makes sense to me.

jesings (approved):
This looks a lot better.

youngfalcon (approved):
Looks good

PR #102, opened 2022-01-28
(Partially) Implemented Ioctl

youngfalcon:


rennergade (approved):
looks good to me!

kapkic (approved):
LGTM!

src/interface/types.rs
```diff
@@ -96,6 +96,13 @@ pub struct TimeVal {
     pub tv_usec: i64
 }
 
+#[repr(C)]
+pub union IoctlUnion {
+    pub arg_int: i32,
+    pub arg_char: char
+    //Right now, we do not support passing struct pointers to ioctl as the related call are not implemented
+}
+
```
jesings, line 105: This is incorrect for 2 reasons:
1. Rust chars are not the same as C chars, for example they are 32 bits... further it's a single value not a pointer. What you probably mean here is a const* u8, which is roughly equivalent to a char* in C,
2. we don't have a union within a union, I think get_ioctlunion wouldn't work because it would try to make into a union pointer the value held in the union field (your nacl implementation doesn't put a union pointer into the big arg union). Instead, you can either just try to use the same casting you did in C to take out a pointer variable and get from it an int if necessary, or try getting it from the arg union directly (although this is annoying, and it might be worth trying to see if you can cast the union into an ioctl union? For that I would use either std::mem::transmute or just pointer casting)

jesings (commented):
Mostly good but the implementation details are still not quite there

src/interface/types.rs
youngfalcon: 1. While the difference in char types might be something I need to fix, there only is one ioctl command that receives a char arg (TIOCSTI) and it only operates on one char (Insert the given byte in the input queue.).

2. Why is it important that we don't put a union pointer there in NaCl? We are literally passing an address as a void pointer from c, can't we just interpret it as we see fit here?

The command always takes in a pointer, and therefore what it points is going to be a char, an int, or a struct (aka the elements of the union). I do not know if there is a rust-specific problem with doing this, so let me know if there is something I am missing like that.

jesings: 1. TIOCSTI still takes a char pointer even though it only modifies that one character pointed to.
2. What we're doing in the NaCl dispatch is taking a void* which could actually contain an int within it, and then putting that into a union field. What you're trying to do, is take a union pointer from that union field, which in fact would be trying to interpret that void* as a union pointer which is obviously incorrect, and you can't do that. One of the options I said is that it is just taking the closest rust equivalent of a void*, this being a const* u8, and casting that pointer if/when necessary to an int using an unsafe helper function. However I don't really think that is the best option and I think the initial union cast into an ioctl union would help relegate the weirder unsafe code to specific, early sections of the dispatcher, which is probably what we want.

jesings (approved):
This is actually a lot cleaner than I was imagining, it looks great!

rennergade (approved):
Yeah this looks great!

PR #104, opened 2022-02-01
Socketpair less awful

jesings:
Mostly my original implementation was 
a) because I was under some mistaken assumptions about why the port closure issue was happening, and
b) because I was trying to copy repy
This version has been tested and works though!


rennergade (approved):
Nice. Looks good!

PR #108, opened 2022-02-15
FS tests no longer commented out, and some fixes for them.

jesings:


youngfalcon (approved):
Looks good to me.

jesings:
No, the getdents test was failing because it was being called twice

rennergade:
Aha! Ok great.

PR #112, opened 2022-02-21
Udp fix

jesings:


jesings:
Somehow this failed on forkandopen.c I will look into this

jesings:
I'm rerunning because I couldn't reproduce on the server. If it was a one off we may still need to look into it.

kapkic:
I will also look into that test case. This should only be happening 1- if there's an error on fd creation 2- if foo.txt is already present (which should've been automatically removed by suite before and after running the test cases).

rennergade (commented):
Seems good.

A few dumb questions.

1. Any reason to not have the early constants not be hex for uniformity?
2. Some of these values are the same for multiple defs, is this correct? ie MSG_DONTROUTE and MSG_TRYHARD

jesings:
As for 1 I just copied these constants from the linux kernel source, we could certainly have them in hex if you'd prefer. I therefore also know that these values are 100% correct even if some of them are duplicates.

rennergade:
Cool. That's fine with me then!

kapkic:
Not tagged but LGTM as well 😄

jesings:
Second run passed so I'm not sure what's going on.

PR #113, opened 2022-02-23
Fixed all the necessary things for gethostbyname/flask

jesings:


src/safeposix/syscalls/net_calls.rs
```diff
@@ -46,6 +46,24 @@ impl Cage {
         fdtable.insert(newfd, wrappedsock);
         newfd
     }
```
rennergade, line 48: very pedantic but feel like you want a newline between these functions

src/safeposix/syscalls/net_calls.rs
```diff
@@ -100,66 +118,66 @@ impl Cage {
     pub fn bind_syscall(&self, fd: i32, localaddr: &interface::GenSockaddr) -> i32 {
         self.bind_inner(fd, localaddr, false)
     }
+    fn bind_inner_socket(&self, sockfdobj: &mut SocketDesc, localaddr: &interface::GenSockaddr, prereserved: bool) -> i32 {
```
rennergade, line 123: same?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -100,66 +118,66 @@ impl Cage {
...
+            match Errno::from_discriminant(interface::get_errno()) {
+                Ok(i) => {return syscall_error(i, "sendto", "The libc call to bind failed!");},
+                Err(()) => panic!("Unknown errno value from socket bind returned!"),
+            };
+        }
+
+        sockfdobj.localaddr = Some(newsockaddr);
+ 
+        0
+    }
```
rennergade, line 167: same

src/safeposix/syscalls/net_calls.rs
jesings: For this one given that this function is a factored out version of the previous they're grouped together. If you'd rather it be otherwise then I'd change it.

src/safeposix/syscalls/net_calls.rs
jesings: This one further is also a factored out helper so that's why these are grouped.

src/safeposix/syscalls/net_calls.rs
rennergade: I think the fact that they're consecutive is fine enough to know they're grouped together. It's just a bit hard to read without spacing, though it is pretty pedantic.

rennergade (approved):
Looks good!

PR #114, opened 2022-02-24
Fix deadlock in listen

rennergade:
Re-organizes listen_sycall a bit to prevent deadlock


TristanB22 (approved):
Seems legit!

kapkic (approved):
Clean fix! LGTM

jesings (approved):
Should be fine, I don't see any other better way to deal with this

PR #120, opened 2022-03-11
CLOEXEC fixes (companion PR)

rennergade:
Fixes CLOEXEC in fcntl set, and stops dup/dup2 from passing flag per dup manpage.


kapkic (approved):
LGTM!

TristanB22 (approved):

PR #125, opened 2022-03-23
Blocking fixes

rennergade:
Joint PR between Caglar and I

- Fixes blocking for ioctl FIONBIO, adds dummy for ioctl ASYNC
- Fixes accept nonblock call
- Fixes select fd_set translation (adds parentheses for proper order of operations)

### DIFF
--- a/src/interface/comm.rs
+++ b/src/interface/comm.rs
@@ -206,11 +206,11 @@ impl Socket {
     }
 
     pub fn set_blocking(&self) -> i32 {
-        unsafe{libc::fcntl(self.raw_sys_fd, libc::F_SETFL, 0);}
+        unsafe{libc::fcntl(self.raw_sys_fd, libc::F_SETFL, 0)}
     }
 
     pub fn set_nonblocking(&self) -> i32 {
-        unsafe{libc::fcntl(self.raw_sys_fd, libc::F_SETFL, libc::O_NONBLOCK);}
+        unsafe{libc::fcntl(self.raw_sys_fd, libc::F_SETFL, libc::O_NONBLOCK)}
     }
 
     pub fn accept(&self, isv4: bool) -> (Result<Self, i32>, GenSockaddr) {
--- a/src/interface/comm.rs
+++ b/src/interface/comm.rs
@@ -206,7 +206,7 @@ impl Socket {
     }
 
     pub fn fcntl(&self, cmd: i32, arg: i32) -> i32 {
-        unsafe {libc::fcntl(self.raw_sys_fd, request, arg)}
+        unsafe {libc::fcntl(self.raw_sys_fd, cmd, arg)}
     }
 
     /*
--- a/src/interface/comm.rs
+++ b/src/interface/comm.rs
@@ -205,6 +205,11 @@ impl Socket {
         unsafe {libc::listen(self.raw_sys_fd, backlog)}
     }
 
+    pub fn fcntl(&self, cmd: i32, arg: i32) -> i32 {
+        unsafe {libc::fcntl(self.raw_sys_fd, request, arg)}
+    }
+
+    /*
     pub fn ioctl(&self, request: u32, argptr: *mut i32) -> i32 {
         //Restricted implementation:
         match request {
@@ -217,6 +222,7 @@ impl Socket {
         //General implementation (disabled for now to avoid the ioctl call):
         //unsafe {libc::ioctl(self.raw_sys_fd, request as u64, argptr as *mut libc::c_void)}
     }
+    */
 
     pub fn accept(&self, isv4: bool) -> (Result<Self, i32>, GenSockaddr) {
         return if isv4 {
--- a/src/interface/comm.rs
+++ b/src/interface/comm.rs
@@ -205,7 +205,7 @@ impl Socket {
         unsafe {libc::listen(self.raw_sys_fd, backlog)}
     }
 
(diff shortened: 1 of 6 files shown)


src/interface/comm.rs
```diff
@@ -205,6 +205,10 @@ impl Socket {
         unsafe {libc::listen(self.raw_sys_fd, backlog)}
     }
 
+    pub fn ioctl(&self, request: u32, argptr: *mut i32) -> i32 {
```
rennergade, line 208: I'm not sure we should call to libc ioctl? We can just set it non-blocking with fcntl? @jesings

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1356,38 +1356,48 @@ impl Cage {
...
+                                };
+                            }
+
                             0
                         }
                         _ => {syscall_error(Errno::ENOTTY, "ioctl", "The specified request does not apply to the kind of object that the file descriptor fd references.")}
                     }
                 }
+                FIOASYNC => { //not implemented
+                    println!("ioctl(FIOASYNC) is not implemented, and just returns 0.");
```
rennergade, line 1398: We should probably only do this so its logs on verbose and not always println

src/safeposix/syscalls/net_calls.rs
```diff
@@ -483,7 +483,7 @@ impl Cage {
                                 if buflen != buflenleft {
                                     return (buflen - buflenleft) as i32;
                                 }
-
+
```
rennergade, line 486: Remove whitespace

src/interface/comm.rs
jesings: I think because we can do it with fcntl and we're not supporting any other ioctls for now we should do it with fcntl.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1356,38 +1356,48 @@ impl Cage {
...
-                File(obj) => {filetype = 4; &mut obj.flags},
-            };
 
             match request {
                 FIONBIO => {
                     let arg_result = interface::get_ioctl_int(ptrunion);
-                    //matching the tuple
-                    match (arg_result, filetype) {
+                    //matching the tuple and passing in filedesc_enum
+                    match (arg_result, &mut *filedesc_enum) {
```
jesings, line 1364: Much cleaner, nice

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1356,38 +1356,48 @@ impl Cage {
...
                             }
                             else { //set for non-blocking I/O
                                 *flags |= O_NONBLOCK;
                             }
+
+                            let sid = Self::getsockobjid(&mut *sockfdobj);
+                            let locksock = NET_METADATA.write().unwrap().socket_object_table.get(&sid).unwrap().clone();
+                            let sockobj = locksock.read().unwrap();
+
+                            let ioctlret: i32 = sockobj.ioctl(request, &mut arg as *mut i32);
```
jesings, line 1383: Yeah this would be replaced by an fcntl

src/safeposix/syscalls/net_calls.rs
```diff
@@ -739,10 +739,21 @@ impl Cage {
...
+                                            _ => panic!("Unknown domain in accepting socket"),
+                                        }
+                                    },
+                                    _ => {
+                                        match sockfdobj.domain {
+                                            PF_INET => sockobj.accept(true),
+                                            PF_INET6 => sockobj.accept(false),
+                                            _ => panic!("Unknown domain in accepting socket"),
+                                        }
+                                    }
```
jesings, line 756: This is incorrect and overkill. Incorrect because flags may contain the O_NONBLOCK bit set but not be equal to it, an if statement with an & check would be what we're looking for.

jesings (commented):
Great progress, a few minor changes to be made.

src/safeposix/syscalls/net_calls.rs
rennergade: ah, makes sense

rennergade: fixed

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Updated to log only on verbose.

src/safeposix/syscalls/net_calls.rs
youngfalcon: Done.

kapkic (approved):
Looks great!

src/safeposix/syscalls/fs_calls.rs
youngfalcon: Updated this. Any comments on the new implementation?

src/interface/comm.rs
youngfalcon: Updated this. Any comments on the new implementation?

youngfalcon (approved):

src/interface/comm.rs
jesings: I think it would be better if you directly called our emulated Socket fcntl call from ioctl, because that way it's clearer if we ever add more ioctl flags that that won't work right.

jesings (approved):
Mostly good, I still had one preference

rennergade (commented):
This looks good to me now!

PR #126, opened 2022-03-27
Finished dashmap implementation

jesings:


src/safeposix/dispatcher.rs
```diff
@@ -382,7 +382,7 @@ pub extern "C" fn dispatcher(cageid: u64, callnum: i32, arg1: Arg, arg2: Arg, ar
             check_and_dispatch!(cage.gethostname_syscall, interface::get_mutcbuf(arg1), interface::get_isize(arg2))
         }
         MKDIR_SYSCALL => {
-            check_and_dispatch!(cage.mkdir_syscall, interface::get_cstr(arg1), interface::get_uint(arg2))
+            check_and_dispatch!(cage.mkdir_syscall, interface::get_cstr(arg1), interface::get_uint(arg2), Ok::<Option<&FilesystemMetadata>, i32>(None))
```
rennergade, line 385: I'm assuming this is for lindfs to work properly, which is fine, but leaving a comment would probably be helpful for us in the future when were scratching our heads.

src/safeposix/filesystem.rs
```diff
@@ -108,30 +105,91 @@ impl FilesystemMetadata {
...
+pub fn format_fs() {
+    let newmetadata = FilesystemMetadata::blank_fs_init();
 
-    let mut mutmetadata = FS_METADATA.write().unwrap();
+    let mut rootinode = newmetadata.inodetable.get_mut(&1).unwrap(); //get root to populate its dict
+    if let Inode::Dir(ref mut rootdir) = *rootinode {
+        rootdir.filename_to_inode_dict.insert("dev".to_string(), 2);
+    }
+    drop(rootinode);
+
```
rennergade, line 133: Small comment here would also be nice since this is a change in how were adding these.

src/safeposix/filesystem.rs
```diff
@@ -153,35 +211,31 @@ pub fn load_fs() {
...
-
     } else {
-       *mutmetadata = FilesystemMetadata::blank_fs_init();
-       drop(mutmetadata);
-       create_log();
-
-       load_fs_special_files(&utilcage);
-       let metadata = FS_METADATA.read().unwrap();
-       persist_metadata(&metadata);
+        if interface::pathexists(LOGFILENAME.to_string()) {
```
rennergade, line 226: Should we let this run if get to this state?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -63,15 +57,15 @@ impl Cage {
                     atime: time, ctime: time, mtime: time,
                 });
 
-                let newinodenum = mutmetadata.nextinode;
-                mutmetadata.nextinode += 1;
-                if let Inode::Dir(ind) = mutmetadata.inodetable.get_mut(&pardirinode).unwrap() {
+                let newinodenum = FS_METADATA.nextinode.load(interface::RustAtomicOrdering::Relaxed);
+                FS_METADATA.nextinode.store(newinodenum + 1 as usize, interface::RustAtomicOrdering::Relaxed);
```
rennergade, line 61: Yeah. While the rest of the PR makes most of these operations so much easier to read, this is now harder. I shouldn't really have to think to determine that this is a simple increment.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -163,8 +154,8 @@ impl Cage {
                     return syscall_error(Errno::EPERM, "mkdir", "Mode bits were not sane");
                 }
 
-                let newinodenum = mutmetadata.nextinode;
-                mutmetadata.nextinode += 1;
+                let newinodenum = metadata.nextinode.load(interface::RustAtomicOrdering::Relaxed);
+                metadata.nextinode.store(newinodenum + 1 as usize, interface::RustAtomicOrdering::Relaxed);
```
rennergade, line 158: same

src/safeposix/syscalls/net_calls.rs
```diff
@@ -242,14 +235,14 @@ impl Cage {
                                     Err(e) => return e,
                                 };
 
-                                //unlock fdtable so that we're fine to call bind_inner
+                                //unlock self.filedescriptortable so that we're fine to call bind_inner
```
rennergade, line 238: is this comment still relevant?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -408,10 +399,9 @@ impl Cage {
                                 None => {return syscall_error(Errno::ENOTCONN, "send", "The descriptor is not connected");},
                             };
 
-                            //drop fdtable lock so as not to deadlock, this should not introduce
+                            //drop self.filedescriptortable lock so as not to deadlock, this should not introduce
```
rennergade, line 402: relevant?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -686,11 +667,11 @@ impl Cage {
             return syscall_error(Errno::EBADF, "cleanup socket", "invalid file descriptor");
         }
 
-        //We have to take this out of the match because the fdtable already has a mutable borrow
-        //which means that I can't change it with a remove until after the fdtable mutable borrow is finished from the match statement
+        //We have to take this out of the match because the self.filedescriptortable already has a mutable borrow
```
rennergade, line 670: does this still make sense?

rennergade (changes requested):
Overall an incredible job here. This actually mostly makes the codebase much easier to read in general.

I've shared a few comments that we should address but they're all minor.

src/safeposix/filesystem.rs
jesings: I think yes, because it is possible that the user removed one and not the other manually, and I don't think it makes that much sense for us to panic in that case, considering there is only one possible thing they could have wanted to do in this case (run it with a new fs).

src/safeposix/net.rs
```diff
@@ -137,14 +137,14 @@ impl NetMetadata {
...
-    pub fn _get_available_tcp_port(&mut self, addr: interface::GenIpaddr, domain: i32, rebindability: bool) -> Result<u16, i32> {
+    pub fn _get_available_tcp_port(&self, addr: interface::GenIpaddr, domain: i32, rebindability: bool) -> Result<u16, i32> {
         if !NET_DEVICES_LIST.contains(&addr) {
             return Err(syscall_error(Errno::EADDRNOTAVAIL, "bind", "Specified network device is not set up for lind or does not exist!"));
         }
         let mut porttuple = mux_port(addr.clone(), 0, domain, TCPPORT);
 
         //start from the starting location we specified in a previous attempt to get an ephemeral port
-        let next_ephemeral = if domain == AF_INET {self.next_ephemeral_port_tcpv4} else if domain == AF_INET6 {self.next_ephemeral_port_tcpv6} else {unreachable!()};
+        let next_ephemeral = if domain == AF_INET {self.next_ephemeral_port_tcpv4.load(interface::RustAtomicOrdering::Relaxed)} else if domain == AF_INET6 {self.next_ephemeral_port_tcpv6.load(interface::RustAtomicOrdering::Relaxed)} else {unreachable!()};
```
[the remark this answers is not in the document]
jesings: Yeah this was all tristan, I will replace these with atomic adds which should be much nicer and then you can tell me what you think.

src/safeposix/filesystem.rs
rennergade: I don't think we should panic, but maybe cleanup the filesystem and shutdown gracefully with a warning. Then the user can just run again but will explicitly know this is happening.

Also on a similar note, does format_fs get rid of the log if it exists as well?

src/safeposix/net.rs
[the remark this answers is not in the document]
jesings: Oh wait no this is a different part than I thought.

src/safeposix/filesystem.rs
jesings: Hmm so right now it warns and keeps going. You'd rather it warn and fail and force them to run it again?
Currently yes it does get rid of the log by calling create_log right after it does the format_fs.

src/safeposix/syscalls/sys_calls.rs
```diff
@@ -11,13 +11,13 @@ use super::fs_constants::*;
...
-        let mut mutcagetable = CAGE_TABLE.write().unwrap();
+        let mutcagetable = &CAGE_TABLE;
 
         //construct new cage struct with a cloned fdtable
-        let mut newfdtable = interface::RustHashMap::new();
+        let newfdtable = interface::RustHashMap::new();
         {
-            let mut mutmetadata = FS_METADATA.write().unwrap();
-            for (key, value) in self.filedescriptortable.read().unwrap().iter() {
+            for refmulti in self.filedescriptortable.iter() {
```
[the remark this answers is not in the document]
jesings: There really is no better name for refmulti, RefMulti is the name of the struct that dashmap exposes through its iterator, so refmulti is about as descriptive as you can get.

src/safeposix/syscalls/fs_calls.rs
jesings: This is where atomic_add should be used instead

src/tools/fs_utils.rs
```diff
@@ -158,10 +158,14 @@ fn main() {
...
-            *metadata = FilesystemMetadata::blank_fs_init();
-            drop(metadata);
-            load_fs_special_files(&utilcage);
+
+            let mut logobj = LOGMAP.write().unwrap();
+            let log = logobj.take().unwrap();
+            let _close = log.close().unwrap();
+            drop(logobj);
+
+            format_fs();
```
rennergade, line 167: I meant here, we dont explicilty delete the log? This was probably my error initially...

src/safeposix/filesystem.rs
rennergade: Yeah I think that makes more sense.

src/tools/fs_utils.rs
jesings: I don't entirely understand what you are saying. I believe you are saying we don't remove the log here. Yes that is true, but because we continue on we overwrite the log file in create_log and it ultimately gets removed at the end if it exits correctly. Because you want me to exit rather than carry on then yes I will have to remove it.

src/safeposix/syscalls/net_calls.rs
jesings: The word mutable does not but otherwise yes it does, but I can restructure it to be better anyway.

src/tools/fs_utils.rs
jesings: Oh ok now I see what you're saying

jesings: Also wait I'm just confused we already do remove the log in format_fs

Cargo.toml
```diff
@@ -12,10 +12,11 @@ test = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0.136", features = ["derive", "rc"] }
```
TristanB22, line 15: Just wanted to make sure... does this look ok? I put this in here and it might have flown under Jonathan's radar.

src/safeposix/dispatcher.rs
```diff
@@ -429,15 +429,24 @@ pub extern "C" fn lindrustinit(verbosity: isize) {
...
 pub extern "C" fn lindrustfinalize() {
     //wipe all keys from hashmap, i.e. free all cages
-    let mut cagetable = CAGE_TABLE.write().unwrap();
-    let drainedcages: Vec<(u64, interface::RustRfc<Cage>)> = cagetable.drain().collect();
-    drop(cagetable);
-    for (_cageid, cage) in drainedcages {
+    let mut remainingcages: Vec<(u64, interface::RustRfc<Cage>)> = vec![];
+
+    //dashmap doesn't allow you to get key, value pairs directly, it only allows you to get a
+    //RefMulti struct which can be decomposed into the key and value
```
TristanB22, line 435: good call on this comment... I didn't even think about adding that

src/safeposix/filesystem.rs
```diff
@@ -108,30 +105,96 @@ impl FilesystemMetadata {
...
+            // Restore metadata
+            interface::serde_deserialize_from_bytes(&metadatabytes).unwrap()
+        } else {
+            FilesystemMetadata::blank_fs_init()
+        }
+    }
+}
 
-    let mut mutmetadata = FS_METADATA.write().unwrap();
+pub fn format_fs() {
```
TristanB22, line 125: I see how this works. Nice solution... unfortunate that we had to do everything manually, though.

src/tests/fs_tests.rs
```diff
@@ -74,39 +67,9 @@ pub mod fs_tests {
 
 
 
-    pub fn persistencetest() {
```
TristanB22, line 77: Is it impossible to have an explicit test for persist now? I assume that it is pretty difficult to test now, just wanted to double check.

TristanB22 (commented):
Awesome! Nice solution on format. Thanks for taking a look at this... I hope it wasn't too painstaking to finish up.

Cargo.toml
jesings: I think the only thing that I will change here is stop fixing such a specific version, 1.0 should be good enough,

src/tests/fs_tests.rs
jesings: Yeah it is impossible without unsafe code at which point it's really not worth it

rennergade:
Went over security concerns for rust crates today. Seems like this should be good. We can merge dashmap/parking lot when we fix the current deadlock bug.

rennergade (approved):
Looks great!

PR #130, opened 2022-04-29
SHM

rennergade:
Opening this for comments:

Implementation of:
- shmget
- shmat
- shmdt
- shmctl


src/safeposix/shm.rs
```diff
@@ -0,0 +1,97 @@
...
+pub fn new_shm_segment(key: i32, size: usize, cageid: i32, uid: u32, gid: u32, mode: u16) -> ShmSegment {
+    ShmSegment::new(key, size, cageid, uid, gid, mode)
+}
+
+impl ShmSegment {
+    pub fn new(key: i32, size: usize, cageid: i32, uid: u32, gid: u32, mode: u16) -> ShmSegment {
+        let filebacking = interface::new_shm_backing(key, size).unwrap();
+
+        let time = interface::timestamp() as isize; //We do a real timestamp now
+        let permstruct = interface::IpcPermStruct { __key: key, uid: uid, gid: gid, cuid: uid, cgid: gid, mode: mode, __seq: 0 };
```
TristanB22, line 30: Why are there underlines before key and seq?

src/safeposix/syscalls/fs_constants.rs
```diff
@@ -114,6 +114,31 @@ pub const SEEK_SET: i32 = 0;
...
 
+pub const IPC_PRIVATE: i32 = 0o0;
+pub const IPC_CREAT: i32 = 0o1000;
+pub const IPC_EXCL: i32 = 0o2000;
+
+pub const IPC_RMID: i32 = 0;
+pub const IPC_SET: i32 = 1;
+pub const IPC_STAT: i32 = 2;
+
+pub const	SHM_DEST: i32 = 0o1000;
```
TristanB22, line 125: Weird spacing

TristanB22 (commented):
Seems pretty good – I only noticed these small things. Hopefully I didn't miss too much.

src/safeposix/shm.rs
rennergade: That's how they're referenced in the linux header file, so I'm keeping it consistent.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1844,4 +1845,100 @@ impl Cage {
...
+    
+            if rm {
+                let key = segment.key;
+                drop(segment);
+    
+                metadata.shmtable.remove(&shmid);
+                metadata.shmkeyidtable.remove(&key);
+            }
+
+        } else { return syscall_error(Errno::EINVAL, "shmdt", "No shared memory segment at shmaddr"); }
```
jesings, line 1916: I think there might be race conditions with shmat and shmdt relating to nattach. I think you may have to wrap shm_nattach in a lock, and lock around shmat/shmdt to ensure things?
This bears further looking into.

jesings (commented):
Overall much simpler than I was thinking! A few things still to do.

src/safeposix/shm.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.shminfo.shm_nattach -= 1;
+        self.shminfo.shm_dtime = interface::timestamp() as isize;
+    }
+}
+
+pub struct ShmMetadata {
+    pub nextid: interface::RustAtomicI32,
+    pub shmkeyidtable: interface::RustHashMap<i32, i32>,
+    pub shmtable: interface::RustHashMap<i32, ShmSegment>,
+    pub rev_shmtable: interface::RustHashMap<i32, interface::RustHashMap<u32, i32>>
```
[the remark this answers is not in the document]
rennergade: Ah, yes definitely makes sense. I originally was using this to check for attachments upon RMID, but then I realized there is a struct field already that handles that.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1844,4 +1845,100 @@ impl Cage {
...
+        } else {
+            if 0 == (shmflg & IPC_CREAT) {
+                return syscall_error(Errno::ENOENT, "shmget", "tried to use a key did not exist, and IPC_CREAT was not specified");
+            }
+            shmid = metadata.new_keyid(key);
+            let mode = (shmflg & 0x1FF) as u16; // mode is 9 least signficant bits of shmflag, even if we dont really do anything with them
+
+            let segment = new_shm_segment(key, size, self.cageid as i32, DEFAULT_UID, DEFAULT_GID, mode);
+            metadata.shmtable.insert(shmid, segment);
+        }
```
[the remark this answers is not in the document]
rennergade: Cool, I'll try to copy/understand what you did here.

src/safeposix/syscalls/fs_calls.rs
rennergade: Shouldn't having dashmap references, both to shm metadata and the info struct deal with concurrency here?

jesings: But we don't have the dashmap reference after segment is dropped, also you don't have a reference to the shmtable because you just have an i32

rennergade:
This is now a full PR request @jesings @TristanB22 

Please review again. Thanks!

src/safeposix/shm.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn map_shm(&mut self, shmaddr: *mut u8, prot: i32) -> i32{
+        let fobjfdno = self.filebacking.as_fd_handle_raw_int();
+        self.shminfo.shm_nattch += 1;
+        self.shminfo.shm_atime = interface::timestamp() as isize;
+        interface::libc_mmap(shmaddr, self.size as usize, prot, MAP_SHARED | MAP_FIXED, fobjfdno, 0)
+    }
+
+    pub fn unmap_shm(&mut self, shmaddr: *mut u8) {
+        interface::libc_mmap(shmaddr, self.size as usize, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);
```
jesings, line 44: You shouldn't need MAP_PRIVATE for PROT_NONE

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1844,4 +1845,100 @@ impl Cage {
...
+            interface::RustHashEntry::Vacant(vacant) => {
+                if 0 == (shmflg & IPC_CREAT) {
+                    return syscall_error(Errno::ENOENT, "shmget", "tried to use a key that did not exist, and IPC_CREAT was not specified");
+                }
+                drop(vacant);
+                shmid = metadata.new_keyid(key);
+                let mode = (shmflg & 0x1FF) as u16; // mode is 9 least signficant bits of shmflag, even if we dont really do anything with them
+
+                let segment = new_shm_segment(key, size, self.cageid as u32, DEFAULT_UID, DEFAULT_GID, mode);
+                metadata.shmtable.insert(shmid, segment);
```
jesings, line 1873: This is not the correct way to use the entry API here.
`new_keyid` should give you a new id but not actually insert it into the table.
You should use `vacant.insert(shmid)` to actually populate the entry

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1844,4 +1845,100 @@ impl Cage {
...
+        let metadata = &SHM_METADATA;
+        let prot : i32;
+        if let Some(mut segment) = metadata.shmtable.get_mut(&shmid) {
+
+            if 0 != (shmflg & SHM_RDONLY) {
+                prot = PROT_READ;
+            }  else { prot = PROT_READ | PROT_WRITE; }
+            metadata.rev_shm_add(self.cageid as u32, shmaddr, shmid);
+            segment.map_shm(shmaddr, prot)
+        } else { return syscall_error(Errno::EINVAL, "shmat", "Invalid shmid value"); }
```
jesings, line 1891: weird to have an implicit return in the if case and an explicit one in the else case.

jesings (approved):
nearly there!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1844,4 +1845,100 @@ impl Cage {
...
+        let metadata = &SHM_METADATA;
+
+        if let Some(mut segment) = metadata.shmtable.get_mut(&shmid) {
+            match cmd {
+                IPC_STAT => {
+                    *buf = segment.shminfo;              
+                }
+                IPC_RMID => {
+                    segment.rmid = true;
+                    segment.shminfo.shm_perm.mode |= SHM_DEST as u16;
```
jesings, line 1934: One more thing: if nattch is 0 you remove instantly here, you need a check for that

src/safeposix/shm.rs
rennergade: I think you're technically correct. But I just mirrored what we had above from munmap, which was originally what NaCl had suggested.

src/safeposix/syscalls/fs_calls.rs
rennergade: Ah ok.

src/tests/fs_tests.rs
```diff
@@ -948,4 +949,31 @@ pub mod fs_tests {
...
+        lindrustinit(0);
+        let cage = {CAGE_TABLE.get(&1).unwrap().clone()};
+        let key = 31337;
+        let mut shmidstruct = ShmidsStruct::default();
+
+        // shmget returns an identifier in shmid
+        let shmid = cage.shmget_syscall(key, 1024, 0666|IPC_CREAT);
+    
+        // shmat to attach to shared memory
+        cage.shmat_syscall(shmid, 0xffffff60 as *mut u8, 0);
```
TristanB22, line 964: we don't want this and other function's return values?

TristanB22 (approved):
There are a lot of functions which return values, but we don't use the values. Is that on purpose? If so, I think that we are good.

src/tests/fs_tests.rs
rennergade: You're right is could be done better, also pointed out an error to me, addresses don't match in shmat/shmdt

PR #131, opened 2022-05-04
Tests fix

jesings:
In order for this to actually work you need to 
cargo build
./target/debug/lind_fs_utils format
cargo test --lib


rennergade:
Thanks!

rennergade (approved):

PR #133, opened 2022-05-15
EmulatedFiles via mmap DRAFT PR

rennergade:
I'm putting this up as a very rough draft to get some comments.

I was able to get substantial improvements [here](https://github.com/Lind-Project/lind_project/issues/241#issuecomment-1125612991). I'll post some actual numbers and flamegraphs in this issue tomorrow. The upside here is that we not only read from mapped files, which is faster, we do away from having to call the seek syscall on every write/read. This change made Lind out perform native in all of the above scenarios.

I'd love some feedback on the actual software engineering here as well as any safety concerns.


src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
 
         let absolute_filename = fs::canonicalize(&path)?;
+        let rawfd = f.as_raw_fd() as i32;
 
         OPEN_FILES.insert(filename.clone());
-        let filesize = f.metadata()?.len();
+        let filesize = f.metadata()?.len() as usize; 
 
-        Ok(EmulatedFile {filename: filename, abs_filename: absolute_filename, fobj: Some(Arc::new(Mutex::new(f))), filesize: filesize as usize})
+        let mapsize = ((filesize / MAP_1GB) + 1) * MAP_1GB;
```
jesings, line 135: Why is the map granularity in gigabytes?

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
+
+        f.set_len(mapsize as u64)?;
+
+        let emfile: Vec<u8>;
+        unsafe {
+            let filemap_addr = mmap(0 as *mut c_void, mapsize, PROT_READ | PROT_WRITE, MAP_SHARED, rawfd, 0 as i64);
+            let _syncret = msync(filemap_addr, mapsize, MS_ASYNC);
```
jesings, line 142: You should not need to msync immediately after a mapping. The memory region starts out with the pages not mapped in memory and when a fault happens the backing file is read and the page is populated. 
Msync writes changes from the memory region back into memory, but as nothing's been written to that region of memory here, nothing happens

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
     }
 
     pub fn close(&self) -> std::io::Result<()> {
         OPEN_FILES.remove(&self.filename);
+        unsafe {
+            let mut fobj = self.fobj.lock();
+            let map = fobj.take().unwrap();
+
+            let (oldmap_addr, _oldlen, _cap) = map.into_raw_parts();
+            munmap(oldmap_addr as *mut c_void, self.mapsize);
```
jesings, line 156: It would make more sense to me to unmap oldlen rather than self.mapsize

jesings: I guess in general is there any reason to have self.mapsize if the length is just map.len()?

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
-            None => panic!("{} is already closed.", self.filename),
-            Some(f) => { 
-                let fobj = f.lock();
-                fobj.set_len(length as u64)?;
-                self.filesize = length;         
-                Ok(())
-            }
-        }
+        let realfobj = self.realfobj.lock();
+        // realfobj.set_len(length as u64)?;
```
jesings, line 194: Why is this still commented out, don't you want to actually truncate the backing file?

src/interface/file.rs
```diff
@@ -90,8 +92,11 @@ pub fn openfile(filename: String, create: bool) -> std::io::Result<EmulatedFile>
 pub struct EmulatedFile {
     filename: String,
     abs_filename: RustPathBuf,
-    fobj: Option<Arc<Mutex<File>>>,
+    fobj: Arc<Mutex<Option<Vec<u8>>>>,
```
jesings, line 95: Not sure why this is a vec rather than a raw slice

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
-                let mut stringbuf = Vec::new();
-                let mut fobj = f.lock();
-                fobj.read_to_end(&mut stringbuf)?;
-                Ok(stringbuf) // return new buf string
-            }
-        }
+        let mut stringbuf = vec![0; self.filesize];
+        let mut realfobj = self.realfobj.lock();
+        realfobj.read_exact(&mut stringbuf)?;
+        Ok(stringbuf) // return new buf string
```
jesings, line 254: I think you should just clone the mapping data.

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
         }
 
-        Ok(bytes_written)
+        let mut fobjopt = self.fobj.lock();
+        let fobj = fobjopt.as_mut().unwrap();
+        if offset > self.filesize {
+            panic!("Seek offset extends past the EOF!");
+        }
+        let fileslice = &mut fobj[offset..(offset + count)];
+        fileslice.copy_from_slice(buf.as_slice());
```
jesings, line 291: No need for buf, just do fileslice.fill(0)

jesings (commented):
Looks on the right track. I think we're sort of taking an idiosyncratic approach to this by still doing some data handling using the backing file but most using the mmap and I've made comments on that. Overall, this needs some polish but looks promising given the performance increases you have talked about.

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
-                if offset > self.filesize {
-                  panic!("Seek offset extends past the EOF!");
-                }
-                fobj.seek(SeekFrom::Start(offset as u64))?;
-                let bytes_read = fobj.read(buf)?;
-                Ok(bytes_read)
-            }
+        let mut fobjopt = self.fobj.lock();
+        let fobj = fobjopt.as_mut().unwrap();
+        if offset > self.filesize {
```
TristanB22, line 214: I feel like this should be at the start so we don't do all of the other code above and then check if offset is too much.

TristanB22: and is it something that we have to panic over?

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
-
         if offset + length > self.filesize {
             self.filesize = offset + length;
+            if self.filesize > self.mapsize { self.remap_file() }
         }
 
-        Ok(bytes_written)
+        let mut fobjopt = self.fobj.lock();
+        let fobj = fobjopt.as_mut().unwrap();
+        if offset > self.filesize {
```
TristanB22, line 239: same

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
-
         if offset + count > self.filesize {
             self.filesize = offset + count;
+            if self.filesize > self.mapsize { self.remap_file() }
         }
 
-        Ok(bytes_written)
+        let mut fobjopt = self.fobj.lock();
+        let fobj = fobjopt.as_mut().unwrap();
+        if offset > self.filesize {
```
TristanB22, line 287: same

src/interface/file.rs
```diff
@@ -122,168 +127,186 @@ impl EmulatedFile {
...
-            }
-        }
 
         if offset + length > self.filesize {
             self.filesize = offset + length;
+            if self.filesize > self.mapsize { self.remap_file() }
+        }
+    
+        let mut realfobj = self.realfobj.lock();
+        if offset > self.filesize {
```
TristanB22, line 269: same

src/interface/file.rs
rennergade: true @yizhuoliang this should be an error not a panic

src/interface/file.rs
```diff
@@ -122,190 +126,216 @@ impl EmulatedFile {
...
+        }
+
+        Ok(EmulatedFile {filename: filename, abs_filename: absolute_filename, fobj: Arc::new(Mutex::new(Some(emfile))), realfobj: Arc::new(Mutex::new(f)), filesize: filesize, rawfd: rawfd, mapsize: mapsize})
     }
 
     pub fn close(&self) -> std::io::Result<()> {
         OPEN_FILES.remove(&self.filename);
+        unsafe {
+            let mut fobj = self.fobj.lock();
+            let map = fobj.take().unwrap();
```
yzhang71, line 151: I don't think line 150 and line 151 require unsafe block to execute.

src/interface/file.rs
```diff
@@ -122,190 +126,216 @@ impl EmulatedFile {
...
+        buf.copy_from_slice(fileslice);
+
+        Ok(real_length)
     }
 
     // Write to file from provided C-buffer
     pub fn writeat(&mut self, ptr: *const u8, length: usize, offset: usize) -> std::io::Result<usize> {
 
-        let bytes_written;
-
```
yzhang71, line 201: In function readat, we have the check of
```
        let mut real_length = length;
        if offset + length > self.filesize { real_length = self.filesize - offset }
```
I think we should also add it here

src/interface/file.rs
```diff
@@ -122,190 +127,211 @@ impl EmulatedFile {
...
+        unsafe {
+            let mut fobj = self.fobj.lock();
+            let map = fobj.take().unwrap();
+
+            let (oldmap_addr, oldlen, _cap) = map.into_raw_parts();
+            let newmap_addr = mremap(oldmap_addr as *mut c_void, oldlen, self.mapsize, MREMAP_MAYMOVE);
+            madvise(newmap_addr.offset(oldlen as isize) as *mut c_void, self.mapsize, MADV_WILLNEED);
+
+            emfile =  Vec::<u8>::from_raw_parts(newmap_addr as *mut u8, self.mapsize, self.mapsize);
+        }
```
kuzeyardabulut, line 182:
```suggestion
        let mut fobj = self.fobj.lock();
        let map = fobj.take().unwrap();
        unsafe{
              let (oldmap_addr, oldlen, _cap) = map.into_raw_parts();
              let newmap_addr = mremap(oldmap_addr as *mut c_void, oldlen, self.mapsize, MREMAP_MAYMOVE);
              madvise(newmap_addr.offset(oldlen as isize) as *mut c_void, self.mapsize, MADV_WILLNEED);
  
              emfile =  Vec::<u8>::from_raw_parts(newmap_addr as *mut u8, self.mapsize, self.mapsize);
        }
```


Can we do that?

PR #135, opened 2022-05-26
Networking fix for localports and socket connections

jesings:
This should fix the nginx issue we have been seeing


rennergade (approved):
This looks great. Elegant solution and well commented.

TristanB22 (approved):
Looks good!

PR #136, opened 2022-06-09
UNIX Domain Sockets

rennergade:
This PR allows us to create and use "unix domain sockets" though we currently fake this by using regular inet4 sockets.

This PR adds the infrastructure necessary to support these sockets including:

- UNIX sockaddr struct with relevant functions
- Socket Inode types, with proper refcount/linkcount and removal in close
- Cleanup of any socket inodes on shutdown
- Tables to translate between our fake domain socket and actual inet socket
- Changes to socket/bind/connect/accept to support domain sockets 
- Ability to perform non-blocking connect w/ select() call


TristanB22 (approved):
Cool! Just took a brief look — I’ll look more into it later.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -249,34 +321,55 @@ impl Cage {
...
                         }
 
+                        if let interface::GenSockaddr::Unix(_) = remoteaddr { sockobj.0.set_blocking(); };
+
                         sockobj.1 = ConnState::CONNECTED;
                         sockfdobj.remoteaddr = Some(remoteaddr.clone());
                         sockfdobj.errno = 0;
+                        if inprogress {
+                            sockobj.1 = ConnState::INPROGRESS;
+                            return syscall_error(Errno::EINPROGRESS, "connect", "The libc call to connect is in progress.");
```
TristanB22, line 371: Why don't we do all this where I commented above?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -249,34 +321,55 @@ impl Cage {
...
                         };
 
-                        let connectret = sockobj.0.connect(remoteaddr);
+                        let mut inprogress = false;
+                        let connectret = sockobj.0.connect(&remoteclone);
                         if connectret < 0 {
                             match Errno::from_discriminant(interface::get_errno()) {
-                                Ok(i) => {return syscall_error(i, "connect", "The libc call to connect failed!");},
+                                Ok(i) => {
+                                    if i == Errno::EINPROGRESS { inprogress = true; }
```
TristanB22, line 356: Here is where the comment below refers to

TristanB22 (approved):
Just a few things to note:

- I feel like you use unreachable when you should use panic sometimes in your unix domain sockets review but that's just semantics
- you could easily shave off 3 lines or so from swap_unixaddr by instead of doing a contains_key and get, just if letting on the get, which would be much cleaner
- please remember to uncomment the fs_tests

src/safeposix/filesystem.rs
```diff
@@ -372,6 +389,30 @@ pub fn normpath(origp: interface::RustPathBuf, cage: &Cage) -> interface::RustPa
     newp
 }
 
+pub fn remove_domain_sock(truepath: interface::RustPathBuf) {
+    match metawalkandparent(truepath.as_path()) {
+        //If the file does not exist
+        (None, ..) => { panic!("path does not exist") }
+        //If the file exists but has no parent, it's the root directory
+        (Some(_), None) => { panic!("cannot unlink root directory") }
```
justyoyo123, line 397: Why do we panic rather than throw an exception here? Would this terminate the whole program?

justyoyo123 (approved):

src/interface/comm.rs
```diff
@@ -125,6 +150,29 @@ impl GenIpaddr {
...
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+pub struct SockaddrUnix {
+    pub sun_family: u16,
+    pub sun_path: [u8; 108]
+}
+
+pub fn new_sockaddr_unix(family: u16, path: &[u8]) -> SockaddrUnix {
+    let pathlen = path.len();    
+    let mut array_path : [u8; 108] = [0; 108];
+    array_path[0..pathlen].copy_from_slice(path);
```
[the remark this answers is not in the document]
rennergade: The path could be less than 108 and the slice lengths need to be equal to use copy from slice. You do make a good point that we should cap pathlen at 108.

src/safeposix/filesystem.rs
rennergade: This is called during cleanup and is on a list of files we're keeping to cleanup. If the file doesn't exist but we think it does, something went very wrong.

rennergade: By "files" here I mean unix domain sockets, which are files but not like you think of them. We don't persist them from run to run so this function is making sure that they're removed before file persistence.

src/interface/comm.rs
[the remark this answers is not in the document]
rennergade: I added a length check here just to be safe, though we only currently call this function with hardcoded strings.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1183,7 +1253,7 @@ impl Cage {
     }
 
     pub fn _close_helper_inner(&self, locked_filedesc: interface::RustRfc<interface::RustLock<FileDescriptor>>) -> i32 {
-        let mut filedesc_enum = locked_filedesc.write();
+        let filedesc_enum = locked_filedesc.write();
```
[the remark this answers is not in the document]
rennergade: No because were not changing the actual enum.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1526,7 +1621,7 @@ impl Cage {
                             syscall_error(Errno::EOPNOTSUPP, "mmap", "lind currently does not support mapping character files")
                         }
 
-                        Inode::Dir(_) => {syscall_error(Errno::EACCES, "mmap", "the fildes argument refers to a file whose type is not supported by mmap")}
+                        _ => {syscall_error(Errno::EACCES, "mmap", "the fildes argument refers to a file whose type is not supported by mmap")}
                     }
                 }
                 _ => {syscall_error(Errno::EACCES, "mmap", "the fildes argument refers to a file whose type is not supported by mmap")}
```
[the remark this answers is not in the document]
rennergade: I think were keeping this separate in case we want to expand this here, either way not really part of this PR.

src/safeposix/syscalls/net_calls.rs
rennergade: It technically could be done up here but I think would be less readable and would require doubling some code.

rennergade: I wanted to lump the change in connection states together.

PR #137, opened 2022-06-19
Getifaddrs

rennergade:
Companion to the libc update:

This changes gen_netdevs to a c program and has the bash script compile it. This gives us a full list of device information. 

We store all that info in a string in rustposix, and return that to libc when requested. The NET_DEVICES list is now generated from parsing the netdevs string for IPs.


youngfalcon (approved):
I see no problems, although I only did a brief review.

kapkic (approved):
Looks good!

gen_netdevs.c
```diff
@@ -0,0 +1,34 @@
...
+            printf("%d ", ifa->ifa_flags);
+            printf("%s ", addr);
+            printf("%s ", naddr);
+            printf("%s\n", baddr);
+        }
+    }
+
+    freeifaddrs(ifap);
+    return 0;
+}
```
justyoyo123, line 34: I think the printfs could all be combined into 1 printf.

PR #138, opened 2022-06-23
Lazy updates

rennergade:
Fixes issues for OnceCell and SyncLazy since they have been renamed in the[ latest rust update.](https://github.com/rust-lang/rust/pull/98165)

### DIFF
--- a/src/interface/errnos.rs
+++ b/src/interface/errnos.rs
@@ -2,9 +2,9 @@
 // Error handling for SafePOSIX
 use crate::interface;
 
-use std::lazy::SyncOnceCell;
+use std::sync::OnceLock;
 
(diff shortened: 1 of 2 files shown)


kapkic (commented):
OnceLock seems to be failing in Actions.

rennergade:
> OnceLock seems to be failing in Actions.

Yeah, so the Rust devs decided to rename these in nightly for the latest release. So with older rustup nightly versions this fix won't work, but if we don't have this it won't work in the future. Seems crazy that they renamed these mid-version without any real disclaimer.

Anyways, this is one of the things inspiring the docker updates.

justyoyo123 (approved):
From what I understand, looks good to me.

PR #141, opened 2022-08-06
netshutdown

rennergade:
This adds the libc shutdown to the interface so we can use it to properly half-close sockets. This also let me implement SHUT_RD, and rearrange some of the function flow to make more sense.

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -739,14 +739,8 @@ impl Cage {
 
     pub fn netshutdown_syscall(&self, fd: i32, how: i32) -> i32 {
         match how {
-            SHUT_RD => {
-                return syscall_error(Errno::EOPNOTSUPP, "netshutdown", "partial shutdown read is not implemented");
-            }
-            SHUT_WR => {
-                return Self::_cleanup_socket(self, fd, true);
-            }
-            SHUT_RDWR => {
-                return Self::_cleanup_socket(self, fd, false);
+            SHUT_RDWR | SHUT_RD | SHUT_WR => {
+                return Self::_cleanup_socket(self, fd, how);
             }
             _ => {
                 //See http://linux.die.net/man/2/shutdown for nuance to this error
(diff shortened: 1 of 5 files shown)


src/safeposix/syscalls/net_calls.rs
```diff
@@ -669,7 +669,7 @@ impl Cage {
                             return 0; //Already done!
                         }
 
-                        ConnState::CONNECTED | ConnState::INPROGRESS => {
+                        ConnState::CONNECTED | ConnState::CONNRDONLY | ConnState::CONNWRONLY | ConnState::INPROGRESS => {
```
TristanB22, line 672: If we add CONNRDWRONLY, this is a spot for it.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -739,14 +739,8 @@ impl Cage {
...
         match how {
-            SHUT_RD => {
-                return syscall_error(Errno::EOPNOTSUPP, "netshutdown", "partial shutdown read is not implemented");
-            }
-            SHUT_WR => {
-                return Self::_cleanup_socket(self, fd, true);
-            }
-            SHUT_RDWR => {
-                return Self::_cleanup_socket(self, fd, false);
+            SHUT_RDWR | SHUT_RD | SHUT_WR => {
```
TristanB22, line 742: same thing here as above

src/safeposix/syscalls/net_calls.rs
```diff
@@ -755,73 +749,75 @@ impl Cage {
...
 
-                            //check reuseaddr/port
-                            for optname in [sockfdobj.options & SO_REUSEPORT, sockfdobj.options & SO_REUSEADDR] {
-                                if optname != 0 {
-                                    if sockobjthere.0.setsockopt(SOL_SOCKET, optname, 1) < 0 {
-                                        panic!("Setsockopt within known ok conditions failed within shutdown!");
-                                    }
-                                }
-                            }
+                    let sockobjtherelock = NET_METADATA.socket_object_table.get(&soid).unwrap().clone();
```
TristanB22, line 758: Is this really a lock? Small thing, but if I remember correctly the parking lot thing that I implemented isn't necessarily a lock. If that's true, maybe this should be named something else

src/safeposix/syscalls/net_calls.rs
```diff
@@ -755,73 +749,75 @@ impl Cage {
...
 
-                            //now we have completely recreated the socket but unconnected
-                        } else {
-                            //Reaching this means that the socket is closed. Removing the sockobj
-                            //indicates that the sockobj will drop, and therefore close
-                            NET_METADATA.socket_object_table.remove(&soid).unwrap();
+                    match how {
+                        SHUT_RD => {
+                            if sockobjthere.1 == ConnState::CONNRDONLY { releaseflag = true; }
+                            sockobjthere.1 = ConnState::CONNWRONLY;
```
TristanB22, line 772: this feels like it should be in an else. I don't know, the code makes sense to me, but I would hate to have an issue where we change the connection to a wronly and the releaseflag is for some reason not set and then the connection remains open as a wronly

src/safeposix/syscalls/net_calls.rs
```diff
@@ -755,73 +749,75 @@ impl Cage {
...
+                        }
+                        SHUT_WR => {
+                            if sockobjthere.1 == ConnState::CONNWRONLY { releaseflag = true; }
+                            sockobjthere.1 = ConnState::CONNRDONLY;
```
TristanB22, line 776: same

src/safeposix/syscalls/net_constants.rs
```diff
@@ -384,5 +384,5 @@ pub const EPOLL_CTL_MOD: i32 = 3;
 //for internal use
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum ConnState {
-    NOTCONNECTED, CONNECTED, LISTEN, INPROGRESS
+    NOTCONNECTED, CONNECTED, CONNRDONLY, CONNWRONLY, LISTEN, INPROGRESS
```
TristanB22, line 387: are we always going to do RDONLY | WRONLY to represent RDWRONLY? Maybe it's worth adding RDWR in?

TristanB22 (commented):
Great! This is awesome progress!

justyoyo123:
I've reviewed this PR that adds libc shutdown and new connection states and it makes sense from what I am able to understand.

src/safeposix/syscalls/net_calls.rs
rennergade: CONNRDWRONLY is the same as just CONNECTED.

rennergade: The only reason we need to include all of them in there is because of how Rust deals with enums and match.

youngfalcon (approved):
I've reviewed this PR that introduces a libc shutdown to the rustposix interface, adds new connection states, improves socket connection checks, and updates the cleanup functionality. I have not noticed any outstanding issues at this time.

PR #145, opened 2022-08-23
Pipe sanity fixes

rennergade:
This PR:

- Fixes an error where I was using & instead of && ... though it somehow worked
- Matches the variables nonblocking to the actual calls and the inner functions
- Changes the eof check to SeqCst, which after testing doesnt add any performance overhead but esnsures correctness. (Somehow this would error when testing very large multi-gig pipelines when using Relaxed).

### DIFF
--- a/src/interface/pipe.rs
+++ b/src/interface/pipe.rs
@@ -61,7 +61,7 @@ impl EmulatedPipe {
 
     // Write length bytes from pointer into pipe
     // BUG: This only currently works as SPSC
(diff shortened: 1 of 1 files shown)


Devansh-Patel (approved):
I've reviewed this PR that fixes the inconsistencies in naming of blocking and non blocking boolean variable, and it looks good to me.

justyoyo123:
I have reviewed this PR that addresses the inconsistency of the non blocking variable and it looks good to me.

PR #148, opened 2022-09-03
Check select

Devansh-Patel:
This PR fixes the issue of select system call not checking whether the pipe is ready to read/write or not.

### DIFF
--- a/src/interface/pipe.rs
+++ b/src/interface/pipe.rs
@@ -59,8 +59,9 @@ impl EmulatedPipe {
         if (flags & O_RDWRFLAGS) == O_WRONLY {self.refcount_write.fetch_sub(1, Ordering::Relaxed);}
     }
     pub fn check_select_read(&self) -> bool {
-        let mut read_end = self.read_end.lock();
-        let mut pipe_space = read_end.len();
+        let read_end = self.read_end.lock();
+        let pipe_space = read_end.len();
+
         if (pipe_space > 0) && !self.eof.load(Ordering::SeqCst){
             return true;
         }
@@ -70,7 +71,7 @@ impl EmulatedPipe {
     }
     pub fn check_select_write(&self) -> bool {
 
-        let mut write_end = self.write_end.lock();
+        let write_end = self.write_end.lock();
 
         let pipe_space = write_end.remaining();
         if pipe_space == self.size {
--- a/src/interface/pipe.rs
+++ b/src/interface/pipe.rs
@@ -62,7 +62,7 @@ impl EmulatedPipe {
         let read_end = self.read_end.lock();
         let pipe_space = read_end.len();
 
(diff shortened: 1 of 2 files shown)


src/interface/pipe.rs
```diff
@@ -58,6 +58,29 @@ impl EmulatedPipe {
         if (flags & O_RDWRFLAGS) == O_RDONLY {self.refcount_read.fetch_sub(1, Ordering::Relaxed);}
         if (flags & O_RDWRFLAGS) == O_WRONLY {self.refcount_write.fetch_sub(1, Ordering::Relaxed);}
     }
+    pub fn check_select_read(&self) -> bool {
+        let read_end = self.read_end.lock();
+        let pipe_space = read_end.len();
+
+        if (pipe_space > 0) && !self.eof.load(Ordering::SeqCst){
```
jesings, line 65: Why do we check EOF here?
How can there be stuff in the read end of the pipe if we've already reached an EOF.
If the EOF is after the stuff in the read end, then select should still exit.

src/interface/pipe.rs
```diff
@@ -58,6 +58,29 @@ impl EmulatedPipe {
...
+            return false;
+        }
+    }
+    pub fn check_select_write(&self) -> bool {
+
+        let write_end = self.write_end.lock();
+
+        let pipe_space = write_end.remaining();
+        if pipe_space == self.size {
+            return false;
```
jesings, line 78: Don't we want to check if the remaining space is nonzero?

jesings (changes requested):
I don't think this is semantically correct, but the structure looks good.

src/interface/pipe.rs
rennergade: I believe this actually should be an OR and just a eof load (non-negated).

One the EOF is set, select will always say the pipe is readable, even if theres nothing in it. But before its set, its only readable when there is data in the pipe.

Devansh-Patel: Noted.

src/interface/pipe.rs
Devansh-Patel: Noted.

src/interface/pipe.rs
```diff
@@ -58,6 +58,29 @@ impl EmulatedPipe {
...
+
+        let write_end = self.write_end.lock();
+
+        let pipe_space = write_end.remaining();
+        if pipe_space == 0 {
+            return false;
+        }
+        else {
+            return true;
+        }
```
jesings, line 82: This could just be replaced by `return pipe_space != 0`, which I think is cleaner, but this isn't a big deal.

jesings (approved):
One minor code cleanup thing, but overall this looks great!

src/interface/pipe.rs
Devansh-Patel: Affirmative.

rennergade (approved):

PR #150, opened 2022-09-12
Moved mutexes into rustposix using lbc and made it nicer in general

jesings:


src/interface/misc.rs
```diff
@@ -165,3 +166,91 @@ impl AdvisoryLock {
...
+    pub fn create() -> Result<Self, i32> {
+        let inner;
+        let libcret;
+        unsafe {
+            inner = std::mem::MaybeUninit::uninit();
+            libcret = libc::pthread_mutex_init((&mut inner.assume_init()) as *mut libc::pthread_mutex_t, std::ptr::null());
+        }
+        if libcret < 0 { Err(libcret) } else { Ok(Self {inner: unsafe{inner.assume_init()}}) }
+    }
+    pub fn lock(&self) -> i32 {
```
rennergade, line 184: Its a bit easier to read if you separate these functions with a newline between them.

rennergade (approved):
This looks semantically correct. Nice job adding this as well as the type checking.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2129,4 +2129,267 @@ impl Cage {
...
+            syscall_error(Errno::EBADF, "cond_destroy", "Condvar handle does not refer to a valid condvar!")
+        }
+        //the RawCondvar is destroyed on Drop
+
+        //this is currently assumed to always succeed, as the man page does not list possible
+        //errors for pthread_cv_destroy
+    }
+
+    pub fn cond_signal_syscall(&self, cv_handle: i32) -> i32 {
+        let cvtable = self.cv_table.read();
```
TristanB22, line 2295: maybe I am being naive, can we do the same thing that we did with the other tables here? The parking lot thing that I helped implement for speed?

TristanB22 (approved):
I think that this all makes sense to me.

src/safeposix/syscalls/fs_calls.rs
jesings: We lock the table with a parking lot mutex, but the mutexes themselves are used from programs which expect them to have the standard pthreads semantics, so we can't really use any rust mutexes here super easily, whether parking lot or no.

PR #152, opened 2022-10-10
Remove Pipetable

rennergade:
This PR removes the idea of the pipetable, and transfers the pipe references to be internal to pipe fds.

This gives us a performance boost because we don't need to go through a locked structure on every I/O command.


TristanB22 (approved):
Logically this all makes sense to me -- I'm not seeing any major issues. More comments explaining what we are doing could be good in general, but that's nit-picky.

jesings (approved):
Looks great!

justyoyo123 (approved):

PR #154, opened 2022-11-09
No Lock Cagetable

rennergade:
This PR changes the CAGETABLE from a lazy global dashmap to a pub static mut vector without any concurrency primitives.

This is being implemented because it yields ~10-15% performance improvement in I/O heavy programs.

The safety proof here relies on that indexes for the table are assigned in NaCl and are concurrency safe. Inserts are only done on init and fork/exec, while deletes are only done via exit().

There is a problem with the situation when there are multiple threads in a cage and one of them calls exit(), though this is currently a problem as well [which results in deadlock](https://github.com/Lind-Project/lind_project/issues/285).

### DIFF
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -56,10 +56,10 @@ pub fn cagetable_clear() {
             let cageopt = cage.take();
             if cageopt.is_some() { exitvec.push(cageopt.unwrap()); }
         }
+    }
 
-        for cage in exitvec {
-            cage.exit_syscall(EXIT_SUCCESS);
-        }
+    for cage in exitvec {
+        cage.exit_syscall(EXIT_SUCCESS);
     }
 }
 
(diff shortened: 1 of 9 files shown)


AlpacaMax (commented):
Looks good to me

src/interface/misc.rs
```diff
@@ -26,6 +26,38 @@ pub use serde_cbor::{ser::to_vec_packed as serde_serialize_to_bytes, from_slice
...
+
+pub fn cagetable_getref(cageid: u64) -> RustRfc<Cage> {
+    unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().clone() }
+}
+
+pub fn cagetable_clear() {
+    unsafe {
+        for cage in CAGE_TABLE.iter_mut() {
+            let cageopt = cage.take();
+            if cageopt.is_some() { cageopt.unwrap().exit_syscall(EXIT_SUCCESS); }
```
jesings, line 56: Ideally this should be out of the unsafe block, maybe collect these into a vector and operate on them that way?

jesings (approved):
Looks great! One nitpick

PR #155, opened 2022-11-10
Deltree fix

rennergade:
This fixes a minor deadlock in rmdir where an inode isn't dropped early enough, as well as adding a few comments


jesings (approved):
Looks good!

AlpacaMax (commented):
Looks good to me

PR #156, opened 2022-11-14
Vector FD Table

rennergade:
This changes the implementation of the fdtable from a DashMap of locked fds to a vector of locked fds. This removes one layer of locking by getting rid of the DashMap, and thus improves I/O performance.

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -44,7 +44,12 @@ impl Cage {
     }
 
     fn _socket_inserter(&self, sockfd: SocketDesc) -> i32 {
-        return self.get_next_fd(None, Socket(sockfd));
+
+        let (fd, guardopt) = self.get_next_fd(None);
+        if fd < 0 { return fd }
+        let fdoption = &mut *guardopt.unwrap();
+        fdoption.insert(Socket(sockfd));
+        return fd;
     }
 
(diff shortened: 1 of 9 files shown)


src/safeposix/syscalls/net_calls.rs
```diff
@@ -44,7 +44,12 @@ impl Cage {
     }
 
     fn _socket_inserter(&self, sockfd: SocketDesc) -> i32 {
-        return self.get_next_fd(None, Socket(sockfd));
+
```
jesings, line 47: This space is odd

src/safeposix/syscalls/net_calls.rs
```diff
@@ -523,7 +520,7 @@ impl Cage {
                             };
```
jesings, line 521: Again it may be possible to omit this line but I'm not sure

jesings (approved):
This looks a lot cleaner and overall great!

TristanB22 (approved):
I agree with what Jonathan noted on this PR. Looking through the implementations that you added, everything seems to make sense to me. Although, I am a little confused why the Dashmap implementation wasn't adding the performance improvements that we were hoping for and now we are being locks back. But, I will defer to you all on that one.

rennergade:
@TristanB22 We're not actually "bringing locks back". The individual fds have always had their own locks. We moved from a globally locked hashmap -> dashmap to remove that outer lock which gave us some speed improvements, but dashmap still has some concurrency needs that makes it slower than what were changing to here. With this version we don't need any outer concurrency primitives at all.

PR #157, opened 2022-12-02
Optimize pipe write

rennergade:
This PR adds a check that tries to optimize pipe writes to 4096 bytes. 

It will still write immediately if the pipe is empty. If its not, it will wait until there is at least 4096 bytes of room before writing.

### DIFF
--- a/src/interface/pipe.rs
+++ b/src/interface/pipe.rs
@@ -17,6 +17,7 @@ use std::fmt;
 const O_RDONLY: i32 = 0o0;
 const O_WRONLY: i32 = 0o1;
 const O_RDWRFLAGS: i32 = 0o3;
+const PAGE_SIZE: usize = 4096;
 
 pub fn new_pipe(size: usize) -> EmulatedPipe {
     EmulatedPipe::new_with_capacity(size)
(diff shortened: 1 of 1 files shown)


src/interface/pipe.rs
```diff
@@ -97,7 +98,10 @@ impl EmulatedPipe {
         }
 
         while bytes_written < length {
-            let bytes_to_write = min(length, bytes_written as usize + write_end.remaining());
+            let remaining = write_end.remaining();
+            // we write if the pipe is empty, otherwise we try to limit writes to 4096 bytes (unless whats leftover of this write is < 4096)
+            if remaining != self.size  && (length - bytes_written) > PAGE_SIZE && remaining < PAGE_SIZE { continue };
```
jesings, line 103: So this was a busy loop before and is still a busy loop now?
Is there no possibility of a TOCTTOU attack here?
Not entirely sure about the semantics but if there's no mutual exclusion that seems like it would be possible?

jesings (commented):
Looks good mostly I just want to be very sure there's no TOCTTOU that was already there or that we are introducing that would result in consecutive writes clobbering each other.

src/interface/pipe.rs
rennergade: So because of the mutexes around write/read we can assume that theres only one active writer until it completes and/or one active reader.

remaining()/len() both give the amount we can work with in the pipe, either an amount of space to write to for the writer, or an amount to read from for the reader. So we check remaining - then a reader reads some - then we write, we still have at least that amount to write to (but would be more now). This works in the opposite direction as well.

Basically there isn't a TOCTTOU for moving the head/available length for reads, or tail/available space for writes.

jesings: Ok that makes sense to me! Looks good.

TristanB22 (approved):
After reviewing what Jonathan commented on in this code, I think that everything that is being implemented in here makes sense. I am generally hesitant to include math when reading and writing data in low level code, but looking at it it seems like this math that you are implementing makes sense.

AlpacaMax (commented):
Looks good to me.

PR #159, opened 2023-01-19
Overhauled socket representation in rustposix

jesings:
Fixes https://github.com/Lind-Project/lind_project/issues/230
May make nginx work

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1276,7 +1276,7 @@ impl Cage {
                 }
                 if cleanflag {
                     let mut fdclone = filedesc_enum.clone();
-                    let retval = self._cleanup_socket_inner(&mut fdclone, false, false);
+                    let retval = self._cleanup_socket_inner(&mut fdclone, -1, false);
                     if retval < 0 {
                         return retval;
                     }
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1170,7 +1170,7 @@ impl Cage {
 
     pub fn close_syscall(&self, fd: i32) -> i32 {
         //check that the fd is valid
-            return Self::_close_helper(self, fd);
+        return Self::_close_helper(self, fd);
     }
 
     pub fn _close_helper_inner(&self, fd: i32) -> i32 {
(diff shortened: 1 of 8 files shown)


jesings:
Test suite all passes, including socketpair.c

src/safeposix/syscalls/net_calls.rs
```diff
@@ -537,118 +543,126 @@ impl Cage {
...
+
+                            if retval < 0 {
+                                //If we have already read from a peek but have failed to read more, exit!
+                                if buflen != buflenleft {
+                                    return (buflen - buflenleft) as i32;
+                                }
+
+                                match Errno::from_discriminant(interface::get_errno()) {
+                                    Ok(i) => {
+                                        if sockfdobj.flags & O_NONBLOCK == 0 && i == Errno::EAGAIN {
```
rennergade, line 602: Commenting here to explain why we're doing this would be good.

rennergade (approved):
This looks very very good. I requested some more comments in a few places, and I feel generally the more the merrier here since this is a decently major change.

rennergade:
Also the runner is being weird so lets get that fixed.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1107,71 +1108,69 @@ impl Cage {
...
+            Pipe(pipe_filedesc_obj) => {
+                pipe_filedesc_obj.pipe.incr_ref(pipe_filedesc_obj.flags);
+            }
+            Socket(_socket_filedesc_obj) => {
+                //we handle the closing of these on drop
+            }
+            Stream(_normalfile_filedesc_obj) => {
+                // no stream refs
+            }
+            _ => {return syscall_error(Errno::EACCES, "dup or dup2", "can't dup the provided file");},
```
TristanB22, line 1142: Random idea but I wonder if there is a way to change this error so that we don't have to pass the string of the system call name to the error. Maybe we can make some sort of truth table that maps some enum to the names? Might just look cleaner.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -675,66 +689,61 @@ impl Cage {
             match filedesc_enum {
                 Socket(ref mut sockfdobj) => {
                     //get or create the socket and bind it before listening
-                    let sid = Self::getsockobjid(sockfdobj);
-                    let locksock = NET_METADATA.socket_object_table.get(&sid).unwrap().clone();
-                    let mut sockobj = locksock.write();
-                    match sockobj.1 {
+                    let sock_tmp = sockfdobj.handle.clone();
+                    let mut sockhandle = sock_tmp.write();
```
TristanB22, line 693: This is nice -- much cleaner than before.

TristanB22 (approved):
I think that this looks good. I didn't see any issues with the implementation in my review.

src/safeposix/syscalls/fs_calls.rs
jesings: That sounds reasonable but should absolutely be done in the scope of a different PR if and when we do it.

rennergade:
Okay, this looks great. Have one more question then should be ready to merge.

For this implementation we catch EAGAIN for blocking sockets, since we've set a timeout. How does this interfere with a user defining a timeout on a blocking socket? Do we even allow that?

rennergade:
Okay thats what I thought. Sounds good.

> We don't currently allow user defined blocking socket timeouts. If and when we do we might need more complicated logic.

PR #162, opened 2023-02-01
EPIPE and Blocking Read fix

rennergade:
This adds the EPIPE error as well as fixing a problem with blocking reads where it would try to read more even if there wasnt anything left in the pipe.


AlpacaMax (commented):
Looks good to me

PR #165, opened 2023-03-23
Io bypass

rennergade:
Creates quick dispatcher bypass functions for I/O increasing performance time

### DIFF
--- a/src/safeposix/dispatcher.rs
+++ b/src/safeposix/dispatcher.rs
@@ -127,6 +127,9 @@ macro_rules! check_and_dispatch_socketpair {
     };
 }
 
+
+// the following "quick" functions are implemented for research purposes
+// to increase I/O performance by bypassing the dispatcher and type checker
 #[no_mangle]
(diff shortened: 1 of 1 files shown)


src/safeposix/dispatcher.rs
```diff
@@ -127,6 +127,16 @@ macro_rules! check_and_dispatch_socketpair {
...
+#[no_mangle]
+pub extern "C" fn quick_write(fd: i32, buf: *const u8, count: usize, cageid: u64) -> i32 {
+  unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().write_syscall(fd, buf, count) }
+}
+
+#[no_mangle]
+pub extern "C" fn quick_read(fd: i32, buf: *mut u8, size: usize, cageid: u64) -> i32 {
+    unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().read_syscall(fd, buf, size) }
+}
+
```
jesings, line 139: Do we need to do as_ref here?

jesings (approved):
Again, I don't like how ridiculously POSIX unconformant this is as something for long term but it is what it is.

PR #166, opened 2023-03-29
Sched yield

rennergade:
Per previous conversation with @moyix. 

This adds sched_yield() calls to pipe loops when we encounter a full or empty pipe. This doesn't change overall pipe performance and avoids us busy looping when there are more cages than processors.


rennergade:
Confirmed this with more testing. If we don't do these yields and (cages > # processors) we get ~100x slowdown.

AlpacaMax (commented):
Looks good to me

jesings (approved):
Looks fine by me!

PR #169, opened 2023-04-28
Implement SIGCHLD and SIGPIPE

AlpacaMax:


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -799,7 +799,10 @@ impl Cage {
 
                     let mut nonblocking = false;
                     if pipe_filedesc_obj.flags & O_NONBLOCK != 0 { nonblocking = true;}
-                    return pipe_filedesc_obj.pipe.write_to_pipe(buf, count, nonblocking) as i32
+                    
+                    let retval = pipe_filedesc_obj.pipe.write_to_pipe(buf, count, nonblocking) as i32;
+                    if retval == -(Errno::EPIPE as i32) { interface::kill_inner(self.cageid, 13); } // Trigger SIGPIPE
+                    retval
```
rennergade, line 805: I believe retval has to be outside the end bracket here, does this compile?

src/safeposix/syscalls/sys_calls.rs
```diff
@@ -216,6 +216,11 @@ impl Cage {
 
         //may not be removable in case of lindrustfinalize, we don't unwrap the remove result
         interface::cagetable_remove(self.cageid);
+        
+        // Trigger SIGCHLD
+        if self.cageid != self.parent {
+            interface::kill_inner(self.parent, 17);
```
rennergade, line 222: I think we should add the signal numbers as constants in sys_constants, will make this more readable and potentially be helpful for other things.

src/interface/misc.rs
```diff
@@ -198,6 +198,12 @@ pub fn lind_sigismember(set: SigsetType, signum: i32) -> bool {
     set & (1 << (signum - 1)) != 0
 }
 
+// Signals
+pub fn kill_inner(cage_id: u64, sig: i32) -> i32 {
```
rennergade, line 202: I see you moved this to the interface, which I think is a good idea. We don't usally use the "inner" name here though. Maybe rename this `lind_kill`?

rennergade (changes requested):
Had a few change requests. Otherwise this looks great.

src/safeposix/syscalls/fs_calls.rs
jesings: This looks fine to me actually, the retval here is just an implicit return, and putting it outside the end bracket would make the match statement break.

AlpacaMax: Yeah the other branches of the match statement also use implicit returns. So this should work.

rennergade (approved):
Awesome this looks great!

PR #170, opened 2023-05-23
Add alarm() and setitimer()

AlpacaMax:


src/interface/timer.rs
```diff
@@ -26,3 +29,103 @@ pub fn readtimer(now: RustInstant) -> RustDuration {
...
+                        // Calling self.set_itimer will automatically turn of the timer if
+                        // next_duration is ZERO
+                    }
+                } else {
+                    break;
+                }
+            }
+
+            thread::sleep(RustDuration::from_millis(1)); // One jiffy
+        }
```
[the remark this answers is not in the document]
AlpacaMax: I think we do. If not it will hold the lock for one millisecond without doing anything.

src/interface/timer.rs
```diff
@@ -26,3 +29,104 @@ pub fn readtimer(now: RustInstant) -> RustDuration {
...
+    }
+
+    pub fn set_itimer(&self, curr_duration: RustDuration, next_duration: RustDuration) {
+        let mut guard = self._ac.lock().unwrap();
+        self._set_itimer(&mut guard, curr_duration, next_duration);
+    }
+
+    pub fn tick(&self) {
+        loop {
```
rennergade, line 99: If I'm reading this right, once a ticking thread is launched it never terminates? Should we terminate the thread if it is cancelled.

rennergade (commented):
I think this looks pretty good. I'm a bit concerned about running threads when the timer isn't running.

src/interface/timer.rs
AlpacaMax: So inside the loop, it will check if is_ticking is true. If it's not then it breaks the loop and terminates the thread. The reason I don't make it a while loop is that it involves acquiring and releasing the mutex twice, which could cause some race conditions as is_ticking could be changed in between.

PR #173, opened 2023-06-09
Fixed incorrect error being returned when the path does not exist.

kpan2034:
Changed error being returned from `EEXIST` to `ENOENT` when path does not exist.


jesings (approved):
Nice catch

rennergade:
Looks good!

PR #176, opened 2023-06-24
Adding mknod and fchdir system calls

kuzeyardabulut:


yizhuoliang:
At line 1043 of `fs_syscalls.rs`, I'm not sure but, is this safe to do the unrap() without checking? I don't know if this would be a problem if the given fd doesn't have associated inode?

Besides, for the `inodeandparent` function, I think maybe call it `filenamefrominode` would be more descriptive? I add comments anyway.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1027,6 +1027,68 @@ impl Cage {
...
+                                        first_iteration = false;
+                                    }
+
+                                    match inodeandparent(*parent_dir_inode, inodenum) {
+                                        Some(name) => {
+                                            let mut new_path_string = String::with_capacity(name.len() + path_string.len() + 1);
+                                            new_path_string.push_str(&name);
+                                            new_path_string.push('/');
+                                            new_path_string.push_str(&path_string);
+                                            path_string = new_path_string;
```
RusherRG, line 1063:
```suggestion
                                            path_string = name + "/" + &path_string;
```
Not sure if this works but this would be a better way of concatenating strings.

RusherRG (changes requested):

kuzeyardabulut:
> At line 1043 of fs_syscalls.rs, I'm not sure but, is this safe to do the unrap() without checking? I don't know if this would be a problem if the given fd doesn't have associated inode?

I don't think using unwrap() here would cause a security hole. There have been cases where this approach has been used in code in the past. I believe the primary problem would be that the program will not be able to do proper error handling. Therefore, I made changes to the code to ensure proper error handling.


> Besides, for the inodeandparent function, I think maybe call it filenamefrominode would be more descriptive? I add comments anyway.

I changed the name of the **inodeandparent** function to **filenamefrominode**.

src/safeposix/syscalls/fs_calls.rs
kuzeyardabulut: It worked.

Yaxuan-w (commented):
looks good to me

RusherRG (approved):

justyoyo123 (approved):

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1027,6 +1027,68 @@ impl Cage {
...
+
+        if let Some(filedesc_enum) = &*unlocked_fd {
+            match filedesc_enum {
+                File(normalfile_filedesc_obj) => {
+                    let mut inodenum = normalfile_filedesc_obj.inode;
+                    loop{
+                        let mut thisinode = match FS_METADATA.inodetable.get_mut(&inodenum) {
+                            Some(inode) => inode,
+                            None => {
+                                return syscall_error(Errno::ENOTDIR, "fchdir", "valid inodenum");
```
kpan2034, line 1045: - Since the inode itself is invalid should `EBADF` be returned or `ENOTDIR`?
- Can we make the error message "invalid inodenum" instead for more clarity?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1027,6 +1027,68 @@ impl Cage {
...
+                            _ => {
+                                return syscall_error(Errno::ENOTDIR, "fchdir", "the last component in fd is not a directory");
+                            }
+                        }
+                    }
+                },
+                _ => return syscall_error(Errno::ENOTDIR, "fchdir", "the last component in fd is not a file"),
+            }
+        } else {
+            return syscall_error(Errno::ENOENT, "fchdir", "invalid file descriptor");
```
kpan2034, line 1083: - `ENOENT` -> `EBADF`, maybe?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1027,6 +1027,68 @@ impl Cage {
...
+
+                                        first_iteration = false;
+                                    }
+
+                                    match filenamefrominode(*parent_dir_inode, inodenum) {
+                                        Some(name) => {
+                                            path_string = name + "/" + &path_string;
+                                            inodenum = *parent_dir_inode;
+                                        },
+                                        None => return syscall_error(Errno::ENOTDIR, "fchdir", "path not found"),
```
kpan2034, line 1066: Since we pass an open `fd` to `fchdir`, can the case where the path is not found ever arise, assuming a valid `fd`?

kpan2034 (commented):
I was looking through the `fchdir` man page and had a few questions regarding the error returned when `fchdir` fails.

Specifically, I'm curious about the following errors values that the man page says can be returned:

1. **EACCES**: Is it possible that we encounter permission issues while traversing the path? I see that `fchdir` is called with an open `fd`, so I don't know if this can be an issue on safeposix but I'm wondering if this has been accounted for.

2. **EBADF**:  If the provided `fd` is invalid, should the error be `ENOENT`or `EBADF`? How does this behave on native Linux?

kuzeyardabulut:
> I was looking through the `fchdir` man page and had a few questions regarding the error returned when `fchdir` fails.
> 
> Specifically, I'm curious about the following errors values that the man page says can be returned:
> 
> 1. **EACCES**: Is it possible that we encounter permission issues while traversing the path? I see that `fchdir` is called with an open `fd`, so I don't know if this can be an issue on safeposix but I'm wondering if this has been accounted for.
> 2. **EBADF**:  If the provided `fd` is invalid, should the error be `ENOENT`or `EBADF`? How does this behave on native Linux?

I guess I overlooked the error types. I edited it now. If there is an error type that needs to be changed now, please let me know.

About your questions: 

1. If you have permission to open the directory, you should be able to call fchdir() on it without encountering an EACCES error.

      However, it's theoretically possible that the permissions of the directory could change between when you open it and when you attempt to change to it. In such a case, if the execute (search) permission for the directory is revoked, fchdir() would fail with EACCES.

2. If you pass an invalid file descriptor to fchdir(), the error that you get should be **EBADF**. 

* EBADF stands for **bad file descriptor**. It is the error that gets returned when a function like fchdir() is called with an invalid file descriptor -- one that isn't open, or isn't appropriate for the operation being attempted.

* ENOENT stands for "No such file or directory". It is the error that gets returned when a function like open() or chdir() is called with a pathname that doesn't correspond to an existing file or directory.

rennergade (changes requested):
I'm having trouble walking through the logic here.

Why do we need to loop?

I think this code needs lots of comments to explain it step by step, then ill be able to give a more thorough review.

kuzeyardabulut:
I've added comment lines as much as I can. Please let me know if there is anything unclear or illogical.

Yaxuan-w (commented):
For `mknod` in fs_calls.rs line 201, I noticed that it only considered case of creating character device files (`S_IFCHR`). But in `mknod` [man page](https://man7.org/linux/man-pages/man2/mknod.2.html), `S_IFREG`, `S_IFBLK`, `S_IFIFO`, or `S_IFSOCK` should also be considered as file type. I'm not sure why we restrict the file type, and should we follow the man page?

rennergade:
Following up on @Yaxuan-w and @yizhuoliang comments

The man page says:
```
       EPERM  mode requested creation of something other than a regular
              file, FIFO (named pipe), or UNIX domain socket, and the
              caller is not privileged (Linux: does not have the
              CAP_MKNOD capability); also returned if the filesystem
              containing pathname does not support the type of node
              requested.
```

For now we probably should restrict to CHR types because we'd need to do additional infrastructure for the rest. So this is good.

rennergade:
So I think a lot of the comments you left may be too fine grained, probably from my prompt to add more comments. Ones that are self-explanatory from the LOC are kind of unnecessary, like for ex. "here we set the flag to true" .

The longer block of code is more what im looking for, but is probably better suited at the beginning so the reader knows what theyre about to encounter. Like I still don't really understand why were looping. A better comment may be just a step by step explanation of the algorithm you're implementing before you do it.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1027,6 +1027,30 @@ impl Cage {
         }
     }
 
+    //------------------------------------FCHDIR SYSCALL------------------------------------
+   
+    pub fn fchdir_syscall(&self, fd: i32) -> i32 {
+        let unlocked_fd = self.filedescriptortable[fd as usize].read();
+
+        let path_string = match &*unlocked_fd {
```
rennergade, line 1035: this is much cleaner, nice

src/safeposix/filesystem.rs
```diff
@@ -321,6 +321,76 @@ pub fn convpath(cpath: &str) -> interface::RustPathBuf {
     interface::RustPathBuf::from(cpath)
 }
 
+// find your absolute pathname from inodenum
+pub fn pathnamefrominodenum(inodenum: usize) -> Option<String>{
```
rennergade, line 325: I would like a longer top level comment above the function describing this algorithm with bullet points. once we add that i think this will be good to go.

rennergade (approved):
This is good now! thanks @kuzeyardabulut

PR #177, opened 2023-06-26
Fix recvloop

rennergade:
This PR fixes the inner part of recv so it properly loops for blocking recvs.

We had implemented a method to be able to cancel blocking recv calls (in case of fault etc) by using a timeout on the kernel blocking recvs, and restarting on EAGAIN. For some reason the actual loop never got PR'd so thats what this fix does.

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -612,6 +612,8 @@ impl Cage {
                                                     // until the individual thread is signaled to cancel itself
                                                     loop { interface::cancelpoint(self.cageid); }
                                                 }
+                                                drop(sockhandle);
+                                                sockhandle = sock_tmp.write();
                                     
                                                 continue; // EAGAIN, try again
                                             }
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -620,7 +620,6 @@ impl Cage {
                                         },
                                         Err(()) => panic!("Unknown errno value from socket recvfrom returned!"),
                                     };
-
                                 }
                                 break; // we're okay to move on
                             }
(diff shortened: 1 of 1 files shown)


jesings (approved):
Looks good to me

kuzeyardabulut:
I think everything looks good

Yaxuan-w (approved):
By introducing the loop, the code can properly handle the EAGAIN error case in a blocking socket. It restarts the receive operation, allowing it to continue attempting to receive data until either data becomes available or the cancellation is triggered. If my understanding is correct, it looks good to me.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -583,43 +583,47 @@ impl Cage {
...
-                                            if self.cancelstatus.load(interface::RustAtomicOrdering::Relaxed) {
-                                                // if the cancel status is set in the cage, we trap around a cancel point
-                                                // until the individual thread is signaled to cancel itself
-                                                loop { interface::cancelpoint(self.cageid); }
-                                            }
-                                
-                                            continue;
-                                        }
+                                if retval < 0 {
+                                    //If we have already read from a peek but have failed to read more, exit!
```
yizhuoliang, line 597: read from a "peer"?

yizhuoliang:
I think the loop for handling EAGAIN is okay. Just have some questions might not relavant to this PR
1) why in the case we read the `bufleft` partially, we just return without retrying?
2) why we need to wait for the cancel signal if we already know it will be cancelled from the cancellation status?

rennergade:
> I think the loop for handling EAGAIN is okay. Just have some questions might not relavant to this PR
> 
>     1. why in the case we read the `bufleft` partially, we just return without retrying?
> 
>     2. why we need to wait for the cancel signal if we already know it will be cancelled from the `self.cancelstatus`?

Good questions Coulson.

1. In this scenario we've already read something through "peek" which means we have something to return. The recv manpage specifies that if recv reads anything we can return.
2. The cancelstatus flag signifies that a thread is ready to be cancelled, so here we put it in a holding pattern until cancel is called on the thread

PR #178, opened 2023-06-29
Move Cancel Out of Interface Pipe Loop

rennergade:
This PR moves the cancel point function out of the interface pipe loop, and handles pipe reads similarly to how we now handle inner recv calls. This is to standardize how cancellation works and this format will eventually let us handle Domain Socket recvs much easier. We also no longer have to add the cageid as a parameter to read_from_pipe which was ugly.

### DIFF
--- a/src/interface/pipe.rs
+++ b/src/interface/pipe.rs
@@ -121,7 +121,7 @@ impl EmulatedPipe {
 
     // Read length bytes from the pipe into pointer
     // Will wait for bytes unless pipe is empty and eof is set.
-    pub fn read_from_pipe(&self, ptr: *mut u8, length: usize, nonblocking: bool, cageid: u64) -> i32 {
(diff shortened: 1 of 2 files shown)


justyoyo123 (approved):
This makes sense to me, looks good.

RusherRG (approved):

jesings (commented):
Ok this looks good for the most part, I can't say I understand why this is necessary but it doesn't look incorrect.

rennergade:
> Ok this looks good for the most part, I can't say I understand why this is necessary but it doesn't look incorrect.

So there's a dead lock with domain sockets using internal pipes where close is trying to get the socket handle but it's stuck in the pipe read loop. Doing this allows us to replicate what we did in recv by releasing the socket handle after a certain amount of time, as well as standardizing where were doing the cancellation point.

PR #179, opened 2023-06-30
fix: lindfs cp for large file sizes

RusherRG:
On Linux, write() (and similar system calls) will transfer at most 0x7ffff000 (2,147,479,552) bytes, returning the number of bytes actually transferred. (This is true on both 32-bit and 64-bit systems.). In current Linux, this limit is governed by `MAX_RW_COUNT=2147479552`. 

As of now, when a file of size greater than the `MAX_RW_COUNT` is copied using `lindfs cp` it copies only the first `MAX_RW_COUNT` bytes resulting in an incomplete transfer of the file to lindfs. To make sure that the entire file is transferred. I have divided the file contents into chunks of `MAX_RW_COUNT` bytes and write each chunk.

### DIFF
--- a/src/tools/lib_fs_utils.rs
+++ b/src/tools/lib_fs_utils.rs
@@ -10,6 +10,8 @@ use crate::interface::errnos::{Errno, syscall_error};
 use crate::interface::types::{ClippedDirent, CLIPPED_DIRENT_SIZE};
 use crate::interface;
 
+const LINUX_MAX_RW_COUNT: usize = 0x7FFFF000;
+
 //we currently handle symlinks as normal files
 
 pub fn update_dir_into_lind(cage: &Cage, hostfilepath: &interface::RustPath, lindfilepath: &str) {
@@ -176,9 +178,9 @@ fn cp_into_lind(cage: &Cage, hostfilepath: &interface::RustPath, lindfilepath: &
 
     let veclen = filecontents.len();
     let mut writtenlen: usize = 0;
+    
     //on Linux, write() (and similar system calls) will transfer at most 0x7ffff000 (2,147,479,552) bytes
-    const LINUX_MAX_RW_COUNT: usize = 2147479552;
-    //dividing filecontents into chunks of 2,147,479,552 bytes and writing each chunk
+    //dividing filecontents into chunks of 0x7ffff000 (2,147,479,552) bytes and writing each chunk
     for chunk in filecontents.chunks(LINUX_MAX_RW_COUNT) {
         writtenlen += cage.write_syscall(lindfd, chunk.as_ptr(), chunk.len()) as usize;
     }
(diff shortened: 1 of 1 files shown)


src/tools/lib_fs_utils.rs
```diff
@@ -173,12 +173,17 @@ fn cp_into_lind(cage: &Cage, hostfilepath: &interface::RustPath, lindfilepath: &
...
     let lindfd = cage.open_syscall(lindtruepath.to_str().unwrap(), O_CREAT | O_TRUNC | O_WRONLY, S_IRWXA);
     assert!(lindfd >= 0);
-    let veclen = filecontents.len();
-    let fileslice = filecontents.as_slice();
-    let writtenlen = cage.write_syscall(lindfd, fileslice.as_ptr(), veclen);
 
+    let veclen = filecontents.len();
+    let mut writtenlen: usize = 0;
+    //on Linux, write() (and similar system calls) will transfer at most 0x7ffff000 (2,147,479,552) bytes
+    const LINUX_MAX_RW_COUNT: usize = 2147479552;
```
rennergade, line 180: I'd move this to the top of the file, maybe declare it using the exponent function for easier reading with a comment

rennergade (changes requested):
minor comments but this is a nice fix

yizhuoliang:
> minor comments but this is a nice fix

Code are easy to understand, already relocate the declaration and expression.

kuzeyardabulut:
The code looks fine to me. As I understand it, when a value greater than 0x7ffff000 bytes is entered in the write syscall, the value is parsed and written with multiple write syscall.

kuzeyardabulut (approved):

src/tools/lib_fs_utils.rs
```diff
@@ -173,12 +175,17 @@ fn cp_into_lind(cage: &Cage, hostfilepath: &interface::RustPath, lindfilepath: &
...
-    let fileslice = filecontents.as_slice();
-    let writtenlen = cage.write_syscall(lindfd, fileslice.as_ptr(), veclen);
 
+    let veclen = filecontents.len();
+    let mut writtenlen: usize = 0;
+    
+    //on Linux, write() (and similar system calls) will transfer at most 0x7ffff000 (2,147,479,552) bytes
+    //dividing filecontents into chunks of 0x7ffff000 (2,147,479,552) bytes and writing each chunk
+    for chunk in filecontents.chunks(LINUX_MAX_RW_COUNT) {
+        writtenlen += cage.write_syscall(lindfd, chunk.as_ptr(), chunk.len()) as usize;
```
kpan2034, line 185: We could consider keeping `total_written_len` and `current_written_len` as two separate variables, for clarity while reading the code.

(We should only do this if the compiler optimizes it so that it does not impact performance)

kpan2034 (approved):
One minor thing that we could consider, otherwise LGTM, so approved.

The change basically enables copying files over `0x7FFFF000` bytes at a time by splitting them into chunks.

Yaxuan-w (approved):
Code is easy to understand with comments. The entire block is divided into chunks according to the upper limit value of write() and written accordingly. LGTM so approved.

rennergade:
This looks good now. Merging.

PR #189, opened 2023-08-25
swap error order in shmget

rennergade:
## Description
Swap error order in shmget
Fixes # (issue)

Apparently shmget checks if the key exists before it does a size check. It seems trivial but apparently postgres utilizes this to do something when it returns EEXIST. This fix fixes the order so pg can operate correctly.

### Type of change
- [ x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

I tested this with a simple test calling shmget twice and checking the errnos

## Checklist:

- [x ] My code follows the style guidelines of this project


RusherRG (approved):
Looks good to me ✌️

PR #190, opened 2023-08-31
Add file descriptor boundary check

Yaxuan-w:
## Description

Fixes [#317](https://github.com/Lind-Project/lind_project/issues/317)

 

### Type of change

Found a bug when fixing libc errno returns: safeposix didn't have boundary check for file descriptors. Then, when users don't initialize fd, it will cause error since the value of the variable would depend on random data in memory.

## How Has This Been Tested?

- `make test` under /lind_project
- `cargo test --lib` under /lind_project/src/safeposix-rust

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -479,7 +479,8 @@ impl Cage {
     //------------------------------------FSTAT SYSCALL------------------------------------
 
     pub fn fstat_syscall(&self, fd: i32, statbuf: &mut StatData) -> i32 {
-        let unlocked_fd = self.filedescriptortable[fd as usize].read();
+        let tmp = self.get_filedescriptor(fd).unwrap();
+        let unlocked_fd = tmp.read();
         if let Some(filedesc_enum) = &*unlocked_fd {
 
             //Delegate populating statbuf to the relevant helper depending on the file type.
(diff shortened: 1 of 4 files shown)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -544,7 +545,8 @@ impl Cage {
     //------------------------------------FSTATFS SYSCALL------------------------------------
 
     pub fn fstatfs_syscall(&self, fd: i32, databuf: &mut FSData) -> i32 {
-        let unlocked_fd = self.filedescriptortable[fd as usize].read();
+        let tmp = self.get_filedescriptor(fd).unwrap();
```
rennergade, line 548: i think we should have a better name than tmp for these, can you not call this plus the read() unlock on the same line?

Yaxuan-w: `self.get_filedescriptor(fd).unwrap().read()` will cause lifetime error, so I separate them

rennergade: ok lets change all the tmp's to 'checkedfd'

rennergade (approved):
This looks great. Fixes the bug and is easier to read.

RusherRG (approved):
The checks related to the range of the fd value make sense. I think we had seen a few segfaults because the checks weren't present, which should be fixed now. Nice fix :tada:

PR #192, opened 2023-09-29
Refactor implementation of RustSemaphore

RusherRG:
## Description

Fixes https://github.com/Lind-Project/lind_project/issues/321

 

Refactored the implementation of RustSemaphore struct because previously it was using `Atomicu32` to store the value variable but it was causing a [Time-of-check to time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bug. If there were 2 processes waiting for the lock and a 3rd process released the lock then in such scenarios both the processes that were waiting used to get the lock which shouldn't be happening. 

This bug was fixed by using `Mutex<u32>` instead of `Atomicu32` making sure that only one thread/process can read/update the value of the variable at any given time.

### Type of change

## How Has This Been Tested?

- `lind_project/tests/test_cases/coreutils_suite/shm_sem_t.sh`
- `lind_project/tests/test_cases/coreutils_suite/shm_sem_fork_t.sh`

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Any dependent changes have been added to a pull request and/or merged in other modules (native-client, lind-glibc, lind-project)

### DIFF
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -391,17 +391,18 @@ impl RustSemaphore {
             let mut value = self.value.lock();
             if *value == 0 {
                 interface::lind_yield();
-                continue;
+            } else {
+                *value = if *value > 0 { *value - 1 } else { 0 };
             }
-                
-            *value = if *value > 0 { *value - 1 } else { 0 };
-            break;
         }
     }
 
-    pub fn unlock(&self) {
+    pub fn unlock(&self) -> bool {
         let mut value = self.value.lock();
-        *value = if *value < (SEM_VALUE_MAX - 1) { *value + 1 } else { SEM_VALUE_MAX };
+        if *value < SEM_VALUE_MAX {
+            *value = *value + 1;
+            return true;
+        } else { return false; }
     }
 
     pub fn get_value(&self) -> i32 {
@@ -412,10 +413,10 @@ impl RustSemaphore {
         let mut value = self.value.lock();
         if *value == 0 {
             return false;
+        } else {
+            *value = if *value > 0 { *value - 1 } else { 0 };
+            return true;
         }
-
-        *value = if *value > 0 { *value - 1 } else { 0 };
-        return true;
     }
 
     pub fn timedlock(&self, timeout: Duration) -> bool {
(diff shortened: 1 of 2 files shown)


src/interface/misc.rs
```diff
@@ -381,70 +381,58 @@ pub struct RustSemaphore {
...
-            Err(_) => false,
+    pub fn lock(&self) {
+        loop {
+            let mut value = self.value.lock();
+            if *value == 0 {
+                interface::lind_yield();
+                continue;
+            }
+                
+            *value = if *value > 0 { *value - 1 } else { 0 };
```
rennergade, line 397: this should be an if/else clause, if you do it that way you dont need the continue

src/interface/misc.rs
```diff
@@ -381,70 +381,58 @@ pub struct RustSemaphore {
...
-
-        match result {
-            Ok(_) => true,
-            Err(_) => false,
+        let mut value = self.value.lock();
+        if *value == 0 {
+            return false;
         }
+
+        *value = if *value > 0 { *value - 1 } else { 0 };
```
rennergade, line 417: again should be if/else

rennergade (changes requested):
A few minor changes requested. If you can add a few comments explaining the reasoning that would be good too.

RusherRG:
> A few minor changes requested. If you can add a few comments explaining the reasoning that would be good too.

Yeah, I'll make those changes. Also, do you mean adding comments to the code?

src/interface/misc.rs
```diff
@@ -381,70 +381,58 @@ pub struct RustSemaphore {
...
-        });
-
-        match result {
-            Ok(_) => true,
-            Err(_) => false,
-        }
-
+    pub fn unlock(&self) {
+        let mut value = self.value.lock();
+        *value = if *value < (SEM_VALUE_MAX - 1) { *value + 1 } else { SEM_VALUE_MAX };
```
Yaxuan-w, line 404: According to the [sem_post](https://man7.org/linux/man-pages/man3/sem_post.3.html) man page, we may need to report error=EOVERFLOW when the value larger than SEM_VALUE_MAX?

rennergade: Good catch we probably need to return an error here

RusherRG: Thanks for pointing this out. I've made the relevant changes.

rennergade (approved):
Looks good now. Approved!

PR #194, opened 2023-10-01
Fix semaphore fork deadlock

rennergade:
## Description

Fixes # (issue)

This fixes an issue where fork would deadlock when duplicating multiply referenced semaphores

### Type of change

## How Has This Been Tested?
Fixes pgbench
Test suite

## Checklist:

- [ x] My code follows the style guidelines of this project
- [x ] My changes generate no new warnings


Yaxuan-w (approved):
As my understanding, in the previous code, two cages share the same reference. When the child process is created, it will try to copy the memory space of the entire process, resulting in multiple references pointing to the same semaphore in the parent process and the child process. The modified code shares a copy of the value to avoid this problem. LGTM so approved.

PR #195, opened 2023-10-04
Semaphore clone for threads

rennergade:
## Description

Fixes a deadlock issue in threaded programs with semaphores trying to lock/unlock from the same table. We clone the semaphores to avoid the dashmap deadlock.

### Type of change

## How Has This Been Tested?

Ran rust test suite and tested python program where deadlock occured.

## Checklist:


Yaxuan-w (approved):
As my understanding, if we don't call drop manually, sementry will still remain a mutable reference until semaphore.lock() is called, which will cause the Rust compiler to report an error because multiple mutable references cannot be held at the same time. This change fixes this problem, LGTM so approved.

PR #196, opened 2023-10-05
Modify a argument of dup2_helper and dup syscalls

yizhuoliang:
## Description
the edits are quite simple, I just moved the bound checking, checking if old = new, and the code of getting the filedesc_enum into both dup_syscall and dup2_syscall, so that now dup2_helper just accept the filedesc_enum as a parameter instead of the oldfd

### Type of change

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1107,13 +1107,11 @@ impl Cage {
             None => STARTINGFD,
         };
 
-        if start_fd == fd { return start_fd; } //if the file descriptors are equal (which is not plausible here), return
-
         // get the filedesc_enum
         let checkedfd = self.get_filedescriptor(fd).unwrap();
         let filedesc_enum = checkedfd.write();
         let filedesc_enum = if let Some(f) = &*filedesc_enum {f} else {
-            return syscall_error(Errno::EBADF, "dup2","Invalid old file descriptor.");
+            return syscall_error(Errno::EBADF, "dup","Invalid old file descriptor.");
         };
 
         //checking whether the fd exists in the file table
@@ -1123,7 +1121,7 @@ impl Cage {
     pub fn dup2_syscall(&self, oldfd: i32, newfd: i32) -> i32{
         //checking if the new fd is out of range
         if newfd >= MAXFD || newfd < 0 {
-           return syscall_error(Errno::EBADF, "dup or dup2", "provided file descriptor is out of range");
+           return syscall_error(Errno::EBADF, "dup2", "provided file descriptor is out of range");
        }
 
        if newfd == oldfd { return newfd; } //if the file descriptors are equal, return the new one
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1409,7 +1409,6 @@ impl Cage {
                     0
                 }
                 (F_DUPFD, arg) if arg >= 0 => {
-                    drop(filedesc_enum);
                     self._dup2_helper(&filedesc_enum, arg, false)
                 }
                 //TO DO: implement. this one is saying get the signals
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1107,29 +1107,40 @@ impl Cage {
             None => STARTINGFD,
         };
 
-        //checking whether the fd exists in the file table
-        return Self::_dup2_helper(&self, fd, start_fd, false)
-    }
+        if start_fd == fd { return start_fd; } //if the file descriptors are equal (which is not plausible here), return
 
-    pub fn dup2_syscall(&self, oldfd: i32, newfd: i32) -> i32{
-        //if the old fd exists, execute the helper, else return error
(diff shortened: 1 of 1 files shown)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1107,29 +1107,40 @@ impl Cage {
             None => STARTINGFD,
         };
 
+        if start_fd == fd { return start_fd; } //if the file descriptors are equal (which is not plausible here), return
```
RusherRG, line 1110: Do we need this check as there is already a check for `new_fd == old_fd` in the `dup2_syscall`?

RusherRG (changes requested):
Requested small changes. This needed to be fixed as it was causing stalling because the lock on the file descriptor wasn't being released. Nice fix :partying_face: 

Everything else looks great and we have already tested it. :v:

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1398,8 +1409,7 @@ impl Cage {
                     0
                 }
                 (F_DUPFD, arg) if arg >= 0 => {
```
rennergade, line 1411: we probably need to add the check to see if fd and arg are equal to each other here like as in dup/dup2

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1107,29 +1107,40 @@ impl Cage {
...
-        //if the old fd exists, execute the helper, else return error
-        return Self::_dup2_helper(&self, oldfd, newfd, true);
-    }
-
-    pub fn _dup2_helper(&self, oldfd: i32, newfd: i32, fromdup2: bool) -> i32 {
         //checking if the new fd is out of range
         if newfd >= MAXFD || newfd < 0 {
-            return syscall_error(Errno::EBADF, "dup or dup2", "provided file descriptor is out of range");
-        }
+           return syscall_error(Errno::EBADF, "dup or dup2", "provided file descriptor is out of range");
```
Yaxuan-w, line 1126: `return syscall_error(Errno::EBADF, "dup2", "provided file descriptor is out of range")`

Yaxuan-w (commented):
I'm a bit confused by `dup2`, do we need to report error when `close()` fail? Because in man page it says: 

> the close is performed silently (i.e., any errors during the close are not reported by dup2()).
Everything else lgtm!

src/safeposix/syscalls/fs_calls.rs
rennergade: That's an interesting catch. Its outside of the scope of this PR a bit but does pertain to the same material.

@coulson Can you test removing the close_result lines so the close is silent?

src/safeposix/syscalls/fs_calls.rs
yizhuoliang: As I commented, this would not happen, I just leave it here for redundancy, but yeah, maybe I'll delete this

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1107,29 +1107,40 @@ impl Cage {
             None => STARTINGFD,
         };
 
+        if start_fd == fd { return start_fd; } //if the file descriptors are equal (which is not plausible here), return
+
+        // get the filedesc_enum
+        let checkedfd = self.get_filedescriptor(fd).unwrap();
+        let filedesc_enum = checkedfd.write();
+        let filedesc_enum = if let Some(f) = &*filedesc_enum {f} else {
+            return syscall_error(Errno::EBADF, "dup2","Invalid old file descriptor.");
```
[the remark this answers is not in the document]
yizhuoliang: thanks

src/safeposix/syscalls/fs_calls.rs
yizhuoliang: the close_result lines should not be removed, because if the close fail, dup2 should also return error, right?

src/safeposix/syscalls/fs_calls.rs
rennergade: dup and dup2 are separate, i think we need to ad this back

src/safeposix/syscalls/fs_calls.rs
rennergade: The man page says that it doesn't

yizhuoliang: Well, I checked Linux code, dup2 does return normally no matter the close succeed or not. I will change this code to be align with this.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1144,9 +1146,7 @@ impl Cage {
                 drop(fdguard);
                 //close the fd in the way of the new fd. If an error is returned from the helper, return the error, else continue to end
                 let close_result = Self::_close_helper_inner(&self, newfd);
-                if close_result < 0 {
-                    return close_result;
-                }
+                // mirror the implementation of linxu, ignore the potential error of the close here
```
rennergade, line 1149: linux not linxu

yizhuoliang: haha sure

PR #197, opened 2023-10-23
Signals

rennergade:
## Description

This branch is a collaboration between @AlpacaMax @jesings and @rennergade  which implements signal handling and associated functions. This allows us to properly do signal handling as well as run postgres and the LAMP stack.

Fixes # [(issue)](https://github.com/Lind-Project/lind_project/issues/296)

- implements per thread kill mechanism wrapping pthread_kill
- implements thread id tracking for kill
- implements storing signal handlers and masks as well as interfacing using sigaction and sigprocmask
- implements timer mechanism for sigalrm
- implements SIGCHLD on cage exit as well as SIGPIPE on broken pipe

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Signals do not work within rustposix without nacl, so no tests are added here. We have added various signal tests to lind_project as well as testing postgres/LAMP stack

## Checklist:

### DIFF
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -51,6 +51,13 @@ pub fn cagetable_getref(cageid: u64) -> RustRfc<Cage> {
     unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().clone() }
 }
 
+pub fn cagetable_getref_opt(cageid: u64) -> Option<RustRfc<Cage>> {
+    unsafe { match CAGE_TABLE[cageid as usize].as_ref() {
+        Some(cage) => Some(cage.clone()),
+        None => None
+    }}
+}
+
 pub fn cagetable_clear() {
     let mut exitvec = Vec::new();
     unsafe {
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -134,7 +134,8 @@ pub fn cancelpoint(cageid: u64) {
 
 pub fn sigcheck(cageid: u64) -> bool {
     let cage = cagetable_getref(cageid);
-    cage.pending_signal.load(RustAtomicOrdering::Relaxed)
+    let pthread_id = get_pthreadid();
+    cage.pending_signal.contains(&pthread_id)
 }
 
 pub fn fillrandom(bufptr: *mut u8, count: usize) -> i32 {
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -16,7 +16,7 @@ use std::str::{from_utf8, Utf8Error};
 pub use std::sync::{Arc as RustRfc};
 pub use parking_lot::{RwLock as RustLock, RwLockWriteGuard as RustLockGuard, Mutex, Condvar};
 
-use libc::{mmap, pthread_self, pthread_exit};
+use libc::{mmap, pthread_self, pthread_exit, pthread_kill};
 use std::ffi::c_void;
 
 pub use serde::{Serialize as SerdeSerialize, Deserialize as SerdeDeserialize};
@@ -98,6 +98,10 @@ pub fn get_errno() -> i32 {
 
 pub fn lind_threadexit() {
     unsafe { pthread_exit(0 as *mut c_void); }
+}
+
+pub fn lind_threadkill(thread_id: u64, sig: i32) -> i32 {
+    unsafe { pthread_kill(thread_id, sig) as i32 }
 }
 
 pub fn get_pthreadid() -> u64 {
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -230,11 +230,24 @@ pub struct AdvisoryLock {
     advisory_condvar: Condvar
 }
 
+/*
+* AdvisoryLock is used to implement advisory locking for files.
+* Specifically, it is used by the flock syscall.
+* If works as follows: The underying mutex has a guard value associated with it.
+* A guard value of zero indicates that it is unlocked.
+* In case an exclusive lock is held, the guard value is set to -1.
+* In case a shared lock is held, the guard value is incremented by 1.
+*/
 impl AdvisoryLock {
     pub fn new() -> Self {
-        Self {advisory_lock: RustRfc::new(Mutex::new(0)), advisory_condvar: Condvar::new()}
+        Self {
+            advisory_lock: RustRfc::new(Mutex::new(0)),
+            advisory_condvar: Condvar::new(),
+        }
     }
 
+    // lock_ex is used to acquire an exclusive lock
+    // if the lock cannot be obtained, it waits
     pub fn lock_ex(&self) {
         let mut waitedguard = self.advisory_lock.lock();
         while *waitedguard != 0 {
@@ -243,6 +256,8 @@ impl AdvisoryLock {
         *waitedguard = -1;
     }
 
+    // lock_sh is used to acquire a shared lock
+    // if the lock cannot be obtained, it waits
     pub fn lock_sh(&self) {
         let mut waitedguard = self.advisory_lock.lock();
         while *waitedguard < 0 {
@@ -250,41 +265,60 @@ impl AdvisoryLock {
         }
         *waitedguard += 1;
     }
+    // try_lock_ex is used to try to acquire an exclusive lock
+    // if the lock cannot be obtained, it returns false
     pub fn try_lock_ex(&self) -> bool {
         if let Some(mut guard) = self.advisory_lock.try_lock() {
             if *guard == 0 {
-              *guard = -1;
-              return true
+                *guard = -1;
+                return true;
             }
         }
         false
     }
+    // try_lock_sh is used to try to acquire a shared lock
+    // if the lock cannot be obtained, it returns false
     pub fn try_lock_sh(&self) -> bool {
         if let Some(mut guard) = self.advisory_lock.try_lock() {
             if *guard >= 0 {
-              *guard += 1;
-              return true
+                *guard += 1;
+                return true;
             }
         }
         false
     }
 
+    /*
+     * unlock is used to release a lock
+     * If a shared lock was held(guard value > 0), it decrements the guard value by one
+     * if no more shared locks are held (i.e. the guard value is now zero), then it notifies a waiting writer
+     * If an exclusive lock was held, it sets the guard value to zero and notifies all waiting readers and writers
+     */
     pub fn unlock(&self) -> bool {
         let mut guard = self.advisory_lock.lock();
 
-        if *guard < 0 {
+        // check if a shared lock is held
+        if *guard > 0 {
+            // release one shared lock by decrementing the guard value
             *guard -= 1;
-  
-            //only a writer could be waiting at this point
-            if *guard == 0 {self.advisory_condvar.notify_one();}
+
+            // if no more shared locks are held, notify a waiting writer and return
+            // only a writer could be waiting at this point
+            if *guard == 0 {
+                self.advisory_condvar.notify_one();
+            }
             true
         } else if *guard == -1 {
-            if *guard != -1 {return false;}
+            // check if an exclusive lock is held
+            // release the exclusive lock by setting guard to 0
             *guard = 0;
-  
-            self.advisory_condvar.notify_all(); //in case readers are waiting
+
+            // notify any waiting reads or writers and return
+            self.advisory_condvar.notify_all();
             true
-        } else {false}
+        } else {
+            false
+        }
     }
 }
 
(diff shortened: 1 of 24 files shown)


src/interface/timer.rs
```diff
@@ -26,3 +29,104 @@ pub fn readtimer(now: RustInstant) -> RustDuration {
...
+                    next_duration: RustDuration::ZERO,
+                    is_ticking: false,
+                }
+            ))
+        }
+    }
+
+    // Similar to getitimer. Returns (next value, current value)
+    pub fn get_itimer(&self) -> (RustDuration, RustDuration) {
+        let guard = self._ac.lock().unwrap();
```
justyoyo123, line 68: Should we add an error check here in case unwrap() fails?

src/interface/timer.rs
```diff
@@ -26,3 +29,104 @@ pub fn readtimer(now: RustInstant) -> RustDuration {
...
+                    start_instant: RustInstant::now(),
+                    curr_duration: RustDuration::ZERO,
+                    next_duration: RustDuration::ZERO,
+                    is_ticking: false,
+                }
+            ))
+        }
+    }
+
+    // Similar to getitimer. Returns (next value, current value)
```
RusherRG, line 66:
```suggestion
    // Similar to getitimer. Returns (current value, next value)
```

Yaxuan-w (commented):
As my understanding, this change implements signal structures, kill syscall, and some corresponding system calls have been modified so that they can return EINVAL errors. I don't quite understand, is it the child process will clear the pending signals in the parent process, so we implemented `rustposix_thread_init` in dispatcher.rs?

Yaxuan-w:
and there's an error when compiling safeposix: 
`error[E0425]: cannot find value 'TEST' in module 'interface'`
`[./src/mklind:395] Error: [function: build_rustposix()] [arguments: (nil)].`

src/safeposix/syscalls/sys_constants.rs
```diff
@@ -15,6 +15,8 @@ pub const DEFAULT_GID : u32 = 1000;
 
 // RESOURCE LIMITS
 
+pub const SIGNAL_MAX : i32 = 64;
```
RusherRG, line 18: Since we only have 32 signals and even in Lind-GlibC we are checking for 32 signals only, can we set this to 64?

RusherRG: And also update SigSetType to be `u32` instead of `u64`?

AlpacaMax:
> As my understanding, this change implements signal structures, kill syscall, and some corresponding system calls have been modified so that they can return EINVAL errors. I don't quite understand, is it the child process will clear the pending signals in the parent process, so we implemented `rustposix_thread_init` in dispatcher.rs?

Yes. I believe that's in the specification of signals.

src/safeposix/syscalls/sys_constants.rs
AlpacaMax: We reserve the upper 32 bits of `sigset` in case we want to implement real-time signals in the future.

src/interface/types.rs
```diff
@@ -571,3 +611,43 @@ pub fn get_duration_from_millis(union_argument: Arg) ->Result<Option<interface::
 pub fn arg_nullity(union_argument: &Arg) -> bool {
     unsafe{union_argument.dispatch_cbuf}.is_null()
 }
+
+pub fn get_sigactionstruct<'a>(union_argument: Arg) -> Result<Option<&'a mut SigactionStruct>, i32> {
+    let pointer = unsafe{union_argument.dispatch_sigactionstruct};
+
+    if !pointer.is_null() {
```
justyoyo123, line 618: Could possibly put this if/else into a helper function since it is the same across all these new functions

src/interface/timer.rs
yizhuoliang: this unwrap would fail only when PoisonError happens, if that happens, means the system is already corrupted internally, then we would want this thread also panic isn't it?

src/safeposix/dispatcher.rs
```diff
@@ -550,6 +585,29 @@ pub extern "C" fn lindthreadremove(cageid: u64, pthreadid: u64) {
...
+    if !interface::lind_sigismember(sigset.load(interface::RustAtomicOrdering::Relaxed), signo) {
+        return match cage.signalhandler.get(&signo) {
+            Some(action_struct) => {
+                    action_struct.sa_handler // if we have a handler and its not blocked return it
+
+            },
+            None => 0, // if we dont have a handler return 0
+        };
+    } else { 
+        let mutpendingset = sigset.load(interface::RustAtomicOrdering::Relaxed);
```
yizhuoliang, line 604: Do we have implicit locking on anything? If not, this load-then-store will always have a risk of race condition, isn't it?

yizhuoliang: Similarly, latter in this PR, the pattern of "firstly load a sigset, after some modification on it, store it back" is prevalent, how is race condition prevented for those syscalls? Even though everything is atomic variable, we can still have situations like
syscall1 ( load X)
syscall2( load X)
syscall1(turn on some bits in X, store it back)
syscall2(turn on some bits in X, store it back)

It's just not obvious to me that, why situations like this will not happen.

src/interface/timer.rs
rennergade: Yeah I think @yizhuoliang is right and this would cause a panic.

src/safeposix/dispatcher.rs
rennergade: Yeah the cage.sigset is a DashMap so when we 'get' from it it explicitly locks the hashmap. Good thing to make sure about though. Thanks!

src/interface/misc.rs
```diff
@@ -52,6 +58,13 @@ pub fn cagetable_getref(cageid: u64) -> RustRfc<Cage> {
...
 }
 
+pub fn cagetable_getref_opt(cageid: u64) -> Option<RustRfc<Cage>> {
+    unsafe { match CAGE_TABLE[cageid as usize].as_ref() {
+        Some(cage) => Some(cage.clone()),
+        None => None
+    }}
+}
+
 pub fn cagetable_clear() {
```
yzhang71, line 68: Above code can be simplified as follow:
`unsafe { CAGE_TABLE[cageid as usize].as_ref().map(|cage| cage.clone()) }
`
Here we are using the unsafe block, which means we are bypassing some of Rust's safety checks. In this case, we are indexing into CAGE_TABLE with cageid as usize. We must ensure that cageid is always a valid index (i.e., it's never out of bounds). And make sure len of CAGE_TABLE is not 0.

src/interface/types.rs
```diff
@@ -541,6 +563,24 @@ pub fn duration_fromtimeval(union_argument: Arg) -> Result<Option<interface::Rus
...
+pub fn get_constitimerval<'a>(union_argument: Arg) -> Result<Option<&'a ITimerVal>, i32> {
+    let pointer = unsafe{union_argument.dispatch_conststructitimerval};
+    if !pointer.is_null() {
+        Ok(Some(unsafe{& *pointer}))
+    } else {
+        Ok(None)
+    }
+}
+
 pub fn duration_fromtimespec(union_argument: Arg) -> Result<interface::RustDuration, i32> {
```
yzhang71, line 584: When you have an if block that ends with a return statement, you don't need to use an else block after it. The reason is that if the condition in the if block is true, the function will return, and any code after the if block won't be executed. If the condition is false, the function will continue executing the code after the if block. So, the else is redundant.

In both functions, the else block after the if isn't necessary. Since you're returning from inside the if block, you can just have the Ok(None) as the next statement.

src/interface/types.rs
```diff
@@ -571,3 +611,43 @@ pub fn get_duration_from_millis(union_argument: Arg) ->Result<Option<interface::
...
+
+pub fn get_constsigsett<'a>(union_argument: Arg) -> Result<Option<&'a SigsetType>, i32> {
+    let pointer = unsafe{union_argument.dispatch_constsigsett};
+
+    if !pointer.is_null() {
+        Ok(Some(unsafe{& *pointer}))
+    } else {
+        Ok(None)
+    }
+}
```
yzhang71, line 653: Same as previous cases, the else block after the if isn't necessary. Since you're returning from inside the if block, you can just have the Ok(None) as the next statement.

src/safeposix/dispatcher.rs
```diff
@@ -149,6 +154,20 @@ pub extern "C" fn quick_read(fd: i32, buf: *mut u8, size: usize, cageid: u64) ->
...
+    cage.main_threadid.store(pthreadid, interface::RustAtomicOrdering::Relaxed);
+    let inheritedsigset = cage.sigset.remove(&0); // in cases of a forked cage, we've stored the inherited sigset at entry 0
+    if inheritedsigset.is_some() { 
+        cage.sigset.insert(pthreadid, inheritedsigset.unwrap().1);
+    } else { cage.sigset.insert(pthreadid, interface::RustAtomicU64::new(0)); }
+
+    cage.pendingsigset.insert(pthreadid, interface::RustAtomicU64::new(0));
+    cage.trusted_signal_flag.insert(pthreadid, signalflag);
+}
+
```
yzhang71, line 170: 1. Error Handling: The code doesn't handle possible errors. For example, what if cagetable_getref fails? We should handle these potential errors rather than ignoring them.

2. Use of unwrap(): Using unwrap() can cause program to panic if the Option is None. It's usually more idiomatic in Rust to handle these cases using pattern matching or methods like map_or.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1576,10 +1580,19 @@ impl Cage {
...
                 }
                 let mut poll_fds_vec: Vec<PollStruct> = vec![];
+                let mut rm_fds_vec: Vec<i32> = vec![];
                 let mut num_events: usize = 0;
                 for set in epollfdobj.registered_fds.iter() {
                     let (&key, &value) = set.pair();
 
+                    // check if any of the registered fds were closed, add them to remove list
+                    let checkedregfd = self.get_filedescriptor(key).unwrap();
+                    let unlocked_regfd = checkedregfd.read();
```
yzhang71, line 1590: I would suggest Error Handling with unwrap: Using unwrap() can cause panics. It's preferable to handle potential errors gracefully.

src/safeposix/syscalls/sys_calls.rs
```diff
@@ -128,6 +128,17 @@ impl Cage {
             } else {panic!("We changed from a directory that was not a directory in chdir!");}
         } else {panic!("We changed from a directory that was not a directory in chdir!");}
 
+        // we grab the parent cages main threads sigset and store it at 0
+        // we do this because we haven't established a thread for the cage yet, and dont have a threadid to store it at
+        // this way the child can initialize the sigset properly when it establishes its own mainthreadid
+        let newsigset = interface::RustHashMap::new();
+        if !interface::RUSTPOSIX_TESTSUITE.load(interface::RustAtomicOrdering::Relaxed) { // we don't add these for the test suite
+            let mainsigsetatomic = self.sigset.get(&self.main_threadid.load(interface::RustAtomicOrdering::Relaxed)).unwrap();
+            let mainsigset = interface::RustAtomicU64::new(mainsigsetatomic.load(interface::RustAtomicOrdering::Relaxed));
```
yzhang71, line 137: I would suggest Error Handling with unwrap: Using unwrap() can cause panics. It's preferable to handle potential errors gracefully.

rennergade:
@yzhang71 Going to answer a bunch of your comments all at once. We use unwrap() (and sometimes panic!) in places where if there is an error, we want the system to shutdown. This is similar to a kernel panic. In these situations we don't want Lind to continue running or to handle them gracefully.

src/safeposix/dispatcher.rs
```diff
@@ -141,14 +146,30 @@ macro_rules! check_and_dispatch_socketpair {
...
-  unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().write_syscall(fd, buf, count) }
+    interface::check_cageid(cageid);
+    unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().write_syscall(fd, buf, count) }
 }
 
 #[no_mangle]
 pub extern "C" fn quick_read(fd: i32, buf: *mut u8, size: usize, cageid: u64) -> i32 {
+    interface::check_cageid(cageid);
     unsafe { CAGE_TABLE[cageid as usize].as_ref().unwrap().read_syscall(fd, buf, size) }
 }
```
yzhang71, line 157: The call to **interface::check_cageid(cageid)** within both the **quick_write** and **quick_read** functions serves as a runtime check to validate the provided cageid. However, this introduces additional runtime overhead. Given that both **writing** and **reading** operations are frequently called during execution, the cumulative overhead may be substantial.

PR #201, opened 2023-11-20
Hashset to bitmap

yizhuoliang:
## Description

Fixes issue #337 in Lind_project

Previously the dispatcher will convert fd_set into hashsets before calling `select_syscall`, which leads to severe overheads, and the adhoc hashsets are very short-living. So I made select and epoll to use bitmaps. As a fd_set is just a bitmap in itself, we don't need to convert anything now. The dispatcher treat original fd_set as a pointer to a `mutcbuf` and pass it into `select_syscall`. The helper functions to check and edit the bitmaps are implemented in types.rs

The unit tests were updted for the new select syscall. Notably, having a faster `select` reveals a concurrency issue of the tester, where the server doesn't wait for client threads to connect. This is also fixed with std barriers.

### Type of change

### DIFF
--- a/src/interface/types.rs
+++ b/src/interface/types.rs
@@ -397,6 +397,18 @@ pub fn fd_set_to_hashset(union_argument: Arg, nfds: i32) -> Result<interface::Ru
     }
     return Ok(hashset);
 }
+// I don't want everything to assume the size of fd_set is 1024 bits, so this
+// helper function will check up to the byte of the highest fd given by nfds
+pub fn is_fd_set_empty(fdset: *const u8, highest_fd: i32) -> bool {
+    let nbytes = (highest_fd as usize + 7) / 8;
+    for i in 0..nbytes {
+        unsafe {
+            if *fdset.add(i) != 0 {return false;}
+        }
+    }
+    return true;
+}
+
 pub fn copy_out_to_fd_set(union_argument: Arg, nfds: i32, hashset: interface::RustHashSet<i32>) {
     let pointer = unsafe{union_argument.dispatch_mutcbuf};
     if pointer.is_null() {return;} //do nothing if it's null
--- a/src/interface/types.rs
+++ b/src/interface/types.rs
@@ -238,6 +238,7 @@ pub fn get_mutcbuf(union_argument: Arg) -> Result<*mut u8, i32> {
     return Err(syscall_error(Errno::EFAULT, "dispatcher", "input data not valid"));
 }
 
+// for the case where the buffer pointer being Null is normal
 pub fn get_mutcbuf_null(union_argument: Arg) -> Result<*mut u8, i32> {
     return Ok (unsafe{union_argument.dispatch_mutcbuf});
 }
--- a/src/interface/types.rs
+++ b/src/interface/types.rs
@@ -383,7 +383,7 @@ pub fn get_sockpair<'a>(union_argument: Arg) -> Result<&'a mut SockPair, i32> {
     return Err(syscall_error(Errno::EFAULT, "dispatcher", "input data not valid"));
 }
 
-// turn off the fd bit in fd_set
+// turn on the fd bit in fd_set
 pub fn fd_set_set(fd_set: *mut u8, fd: i32) {
     let byte_offset = fd / 8;
     let byte_ptr = fd_set.wrapping_offset(byte_offset as isize);
@@ -391,7 +391,7 @@ pub fn fd_set_set(fd_set: *mut u8, fd: i32) {
     unsafe{*byte_ptr |= 1 << bit_offset;}
 }
 
-// turn on the fd bit in fd_set
+// turn off the fd bit in fd_set
 pub fn fd_set_unset(fd_set: *mut u8, fd: i32) {
     let byte_offset = fd / 8;
     let byte_ptr = fd_set.wrapping_offset(byte_offset as isize);
@@ -404,7 +404,7 @@ pub fn fd_set_check_fd(fd_set: *const u8, fd: i32) -> bool {
     let byte_offset = fd / 8;
     let byte_ptr = fd_set.wrapping_offset(byte_offset as isize);
     let bit_offset = fd & 0b111;
-    return (unsafe{*byte_ptr}) & (1 << bit_offset) == 0;
+    return (unsafe{*byte_ptr}) & (1 << bit_offset) != 0;
 }
 
 // I don't want everything to assume the size of fd_set is 1024 bits, so this
--- a/src/interface/types.rs
+++ b/src/interface/types.rs
@@ -238,6 +238,10 @@ pub fn get_mutcbuf(union_argument: Arg) -> Result<*mut u8, i32> {
     return Err(syscall_error(Errno::EFAULT, "dispatcher", "input data not valid"));
 }
 
+pub fn get_mutcbuf_null(union_argument: Arg) -> Result<*mut u8, i32> {
+    return Ok (unsafe{union_argument.dispatch_mutcbuf});
+}
+
 pub fn get_cstr<'a>(union_argument: Arg) -> Result<&'a str, i32> {
    
     //first we check that the pointer is not null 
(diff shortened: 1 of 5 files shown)


src/safeposix/dispatcher.rs
```diff
@@ -351,18 +351,11 @@ pub extern "C" fn dispatcher(cageid: u64, callnum: i32, arg1: Arg, arg2: Arg, ar
             let nfds = get_onearg!(interface::get_int(arg1));
             if nfds < 0 { //RLIMIT_NOFILE check as well?
```
rennergade, line 352: it may make more sense to do this check in NaCl

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1240,19 +1234,27 @@ impl Cage {
        loop { //we must block manually
            
             // 1. iterate thru readfds
-            let mut res = self.select_readfds(readfds, &mut new_readfds, &mut retval);
-            if res != 0 {return res}
+            if !readfds.is_null() {
```
rennergade, line 1237: I had asked you to wrap these in an Option as Some() or None() I think that is better practice with Rust here

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1240,19 +1234,27 @@ impl Cage {
...
+                let res = self.select_readfds(nfds, readfds,  &mut retval);
+                if res != 0 {return res}
+            }
             
             // 2. iterate thru writefds
-            res = self.select_writefds(writefds, &mut new_writefds, &mut retval);
-            if res != 0 {return res}
+            if !writefds.is_null() {
```
rennergade, line 1243: same as above

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1240,19 +1234,27 @@ impl Cage {
...
+            }
             
             // 3. iterate thru exceptfds
-            for fd in exceptfds.iter() {
-                //we say none of them ever have exceptional conditions
-                let checkedfd = self.get_filedescriptor(*fd).unwrap();
-                let unlocked_fd = checkedfd.read();
-                if let None = *unlocked_fd { return syscall_error(Errno::EBADF, "select", "invalid file descriptor"); }
+            // currently we don't really do select on execptfds, we just check if those fds are valid
+            if !exceptfds.is_null() {
```
rennergade, line 1250: same

src/tests/networking_tests.rs
```diff
@@ -692,30 +693,44 @@ pub mod net_tests {
...
+        assert_eq!(interface::fd_set_check_fd(inputs, filefd), true);
+        assert_eq!(interface::fd_set_check_fd(outputs, filefd), true);
 
         assert_eq!(cage.fork_syscall(2), 0);
         assert_eq!(cage.fork_syscall(3), 0);
 
         assert_eq!(cage.close_syscall(clientsockfd1), 0);
         assert_eq!(cage.close_syscall(clientsockfd2), 0);
 
+        let barrier = Arc::new(Barrier::new(3));
```
rennergade, line 719: a brief comment explaining the barrier would be helpful for people looking at this in the future

src/interface/types.rs
```diff
@@ -379,40 +384,42 @@ pub fn get_sockpair<'a>(union_argument: Arg) -> Result<&'a mut SockPair, i32> {
...
-        let bit_offset = i & 0b111;
-        let byte_ptr = pointer.wrapping_offset(byte_offset as isize);
-
-        if hashset.contains(&i) {
-            //if it's in the hash set, set the bit ot 1
-            unsafe{*byte_ptr |= 1 << bit_offset;}
-        } else {
-            //else, set the bit to 0
-            unsafe{*byte_ptr &= !(1 << bit_offset);}
+pub const FD_SET_SIZE :i32 = 1024;
```
rennergade, line 387: FD_SET_SIZE may make more sense to go in net_constants

rennergade (changes requested):
Looks very good! Requested a few changes.

src/safeposix/dispatcher.rs
rennergade: or move actually into the select_syscall, i think either of those options would be cleaner

justyoyo123 (approved):
This PR optimizes the `select_syscall` by making use of bitmaps instead of hashsets for fd_set. This increases performance by getting rid of the overhead caused by converting fd_set to a hashet. After addressing Nick's comments, I believe this is good to go.

src/interface/types.rs
```diff
@@ -379,40 +384,42 @@ pub fn get_sockpair<'a>(union_argument: Arg) -> Result<&'a mut SockPair, i32> {
...
+pub fn fd_set_check_fd(fd_set: *const u8, fd: i32) -> bool {
+    let byte_offset = fd / 8;
+    let byte_ptr = fd_set.wrapping_offset(byte_offset as isize);
+    let bit_offset = fd & 0b111;
+    return (unsafe{*byte_ptr}) & (1 << bit_offset) != 0;
+}
+
+// I don't want everything to assume the size of fd_set is 1024 bits, so this
+// helper function will check up to the byte of the highest fd given by nfds
+pub fn is_fd_set_empty(fd_set: *const u8, highest_fd: i32) -> bool {
```
RusherRG, line 414:
```suggestion
pub fn fd_set_is_empty(fd_set: *const u8, highest_fd: i32) -> bool {
```
Just for consistency

src/safeposix/syscalls/net_calls.rs
RusherRG: Yeah I agree, that would be good.

RusherRG (changes requested):

RusherRG:
Great fix. As @justyoyo123 said, it would help improve the performance of the `select_syscall` to help avoid unnecessary stalls in applications like Postgres.

src/interface/types.rs
```diff
@@ -379,40 +389,41 @@ pub fn get_sockpair<'a>(union_argument: Arg) -> Result<&'a mut SockPair, i32> {
...
+
+// return true if the bit for fd is set in fd_set, false otherwise
+pub fn fd_set_check_fd(fd_set: *const u8, fd: i32) -> bool {
+    let byte_offset = fd / 8;
+    let byte_ptr = fd_set.wrapping_offset(byte_offset as isize);
+    let bit_offset = fd & 0b111;
+    return (unsafe{*byte_ptr}) & (1 << bit_offset) != 0;
+}
+
+// I don't want everything to assume the size of fd_set is 1024 bits, so this
```
rennergade, line 416: can we rephrase this from "I" to "we"

src/tests/networking_tests.rs
```diff
@@ -1,32 +1,33 @@
 #[cfg(test)]
 pub mod net_tests {
-    use crate::interface;
+    use crate::interface::{self, fd_set_check_fd, fd_set_insert, fd_set_remove};
```
rennergade, line 3: why do we have to import these separately?

rennergade: we call functions from the interface in here already without specific imports so don't really understand this change

src/tests/networking_tests.rs
```diff
@@ -692,30 +693,43 @@ pub mod net_tests {
...
+        assert_eq!(interface::fd_set_check_fd(inputs, filefd), true);
+        assert_eq!(interface::fd_set_check_fd(outputs, filefd), true);
 
         assert_eq!(cage.fork_syscall(2), 0);
         assert_eq!(cage.fork_syscall(3), 0);
 
         assert_eq!(cage.close_syscall(clientsockfd1), 0);
         assert_eq!(cage.close_syscall(clientsockfd2), 0);
 
+        // those barriers ensures that the clients finish the connect before we do the select
```
rennergade, line 717: "these"

src/tests/networking_tests.rs
```diff
@@ -692,30 +693,43 @@ pub mod net_tests {
...
             let cage2 = interface::cagetable_getref(2);
             assert_eq!(cage2.close_syscall(serversockfd), 0);
 
             assert_eq!(cage2.connect_syscall(clientsockfd1, &socket), 0);
+            barrier_clone1.wait();
             assert_eq!(cage2.send_syscall(clientsockfd1, str2cbuf("test"), 4, 0), 4);
 
             interface::sleep(interface::RustDuration::from_millis(1));
-
+
```
rennergade, line 732: extra whitespace

src/interface/types.rs
```diff
@@ -172,6 +172,7 @@ pub union Arg {
 
 
 use std::mem::size_of;
+use std::ptr::null;
```
rennergade, line 175: i think this can be removed?

rennergade (approved):
Awesome!

PR #203, opened 2023-12-08
Adding sync_file_range Syscall

kuzeyardabulut:


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -7,6 +7,7 @@ use crate::safeposix::filesystem::*;
 use crate::safeposix::net::{NET_METADATA};
 use crate::safeposix::shm::*;
 use super::fs_constants::*;
+use std::convert::TryInto;
```
rennergade, line 10: do you still use this? if so we need to move this into the interface files, i think it may have been exported already?

rennergade (changes requested):
a few nitpicks about returning errors. we basically never want to panic when a user could possibly influence it, and want to return libc errnos whenever possible.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1909,6 +1909,45 @@ impl Cage {
...
+                File(ref mut normalfile_filedesc_obj) => {
+                    if is_rdonly(normalfile_filedesc_obj.flags) {
+                        return syscall_error(Errno::EBADF, "sync_file_range", "specified file not open for sync");
+                    }
+                    let inodeobj = FS_METADATA.inodetable.get(&normalfile_filedesc_obj.inode).unwrap();
+                    match &*inodeobj {
+                        Inode::File(_) => {
+                            let fobj = FILEOBJECTTABLE.get(&normalfile_filedesc_obj.inode).unwrap();
+                            let result = fobj.sync_file_range(offset, nbytes, flags).unwrap();
+                            match result {
```
rennergade, line 1928: this seems wrong to do it like this, we should match result for ok or err instead of unwrapping it and wrap the errnos in err?

You should also be able to handle the libc errors with from discriminant (check out net calls for a few examples.

src/interface/file.rs
```diff
@@ -173,6 +172,25 @@ impl EmulatedFile {
         }
     }
 
+    pub fn sync_file_range(&self, offset: isize, nbytes: isize, flags: u32) -> i32 {
+        let fd = &self.as_fd_handle_raw_int();
+        let valid_flags = libc::SYNC_FILE_RANGE_WAIT_BEFORE | libc::SYNC_FILE_RANGE_WRITE | libc::SYNC_FILE_RANGE_WAIT_AFTER;
+        if !(flags & !valid_flags == 0){
+            return syscall_error(Errno::EINVAL, "sync_file_range", "flags specifies an invalid bit");
+        }
+        let result = unsafe { libc::sync_file_range(*fd, offset.try_into().unwrap(), nbytes.try_into().unwrap(), flags) };
```
rennergade, line 181: use "as" instead of try into

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1909,6 +1910,42 @@ impl Cage {
...
+                File(ref mut normalfile_filedesc_obj) => {
+                    if is_rdonly(normalfile_filedesc_obj.flags) {
+                        return syscall_error(Errno::EBADF, "sync_file_range", "specified file not open for sync");
+                    }
+                    let inodeobj = FS_METADATA.inodetable.get(&normalfile_filedesc_obj.inode).unwrap();
+                    match &*inodeobj {
+                        Inode::File(_) => {
+                            let fobj = FILEOBJECTTABLE.get(&normalfile_filedesc_obj.inode).unwrap();
+                            let result = fobj.sync_file_range(offset, nbytes, flags);
+                            if result == 0 || result == -(EINVAL as i32) {
```
rennergade, line 1929: this looks great but lets add a short comment here to explain the logic

rennergade (approved):
This looks great now!

RusherRG (approved):

PR #209, opened 2024-02-22
Minor Select Optimizations

rennergade:
## Description
Minor performance optimizations for select (change to readlocks, yield instead of sleep)


RusherRG (approved):
This looks good and logical. We have also tested this out and noticed a small improvement in performance.

PR #211, opened 2024-03-11
Refine select nodelay

yizhuoliang:
## Description

Before this, `select()` on inet sockets experience lock contention on `SocketHandles` and degrades the performance of nginx. So now we added domain and system raw fd into `SocketDest`, so `select()` won't need to acquire sockhandle locks for every fd now. Note that the `domain` is set at the point of socket creation, but the `rawfd` field is set at listen, connect, and accept. The fundamental reason we can do this is that, both domain and rawfd can't change once it is set.

Also changed the `options` field of a scokethandle to be `socket_options` and `tcp_options`, as we are handling two different levels of socket options. Thus changing `getsockopt` and `setsockopt`.

Also because `select()` on inet sockets don't access sockhanldes, we added checks at `recv_syscall` to see if the `INPROGRESS` state is already changed to `CONNECTED`.

### Type of change


src/interface/comm.rs
```diff
@@ -366,3 +369,95 @@ impl Drop for Socket {
...
+
+    pub fn is_empty(&self) -> bool {
+        let fd_array: &[u8] = unsafe {
+            std::slice::from_raw_parts(&self.0 as *const _ as *const u8, mem::size_of::<libc::fd_set>())
+        };
+        fd_array.iter().all(|&byte| byte == 0)
+    }
+
+    // for each fd, if kernel_fds turned it on, then self will turn the corresponding tranlated fd on
+    pub fn set_from_kernelfds_and_translate(&mut self, kernel_fds: &FdSet, nfds: i32, rawfd_lindfd_tuples: &Vec<(i32, i32)>) {
```
rennergade, line 422: theres a lot of nesting here, maybe condense some of this to one liners for readability

rennergade (changes requested):
some initial thoughts

justyoyo123 (approved):

src/interface/comm.rs
yzhang71: Following Nick's suggestion, the nested loop can be modified as something like this:
```
use std::collections::HashMap;

pub fn set_from_kernelfds_and_translate(&mut self, kernel_fds: &FdSet, nfds: i32, rawfd_lindfd_tuples: &Vec<(i32, i32)>) {
    // Create a HashMap from the rawfd_lindfd_tuples for O(1) look-up
    let fd_map: HashMap<i32, i32> = rawfd_lindfd_tuples.iter().cloned().collect();
    for fd in 0..nfds {
        if kernel_fds.is_set(fd) {
            if let Some(lindfd) = fd_map.get(&fd) {
                self.set(*lindfd);
            }
        }
    }
}
```

yzhang71 (approved):
Except for the parts with the nested loops (see detailed comments), everything looks good to me.

src/interface/comm.rs
rennergade: this is probably not a great idea because I believe creating a hashmap is slow, so even though the algorithmic complexity may be better N should never be large enough where it surpasses that allocation cost.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1229,32 +1221,32 @@ impl Cage {
...
-        let new_writefds =  new_writes_chunk.as_mut_ptr();
+        let new_readfds = &mut interface::FdSet::new();
+        let new_writefds =  &mut interface::FdSet::new();
         loop { //we must block manually
             // 1. iterate thru readfds
-            if readfds.is_some() {
-                let res = self.select_readfds(nfds, readfds.unwrap(), new_readfds, &mut retval);
+            if let Some(readfds_ref) = readfds.as_ref() {
+                let res = self.select_readfds(nfds, readfds_ref, new_readfds, &mut retval);
                 if res != 0 {return res}
```
Yaxuan-w, line 1230: We may need to handle error cases here?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1378,28 +1339,44 @@ impl Cage {
...
             }
         }
+
+        // do the kernel_select for inet sockets
+        if !kernel_inet_fds.is_empty() {
+            let kernel_ret;
+            // note that this select call always have timeout = 0, so it doesn't block
+            
+            kernel_ret = interface::kernel_select(highest_raw_fd + 1, Some(kernel_inet_fds), None, None);
+            if kernel_ret < 0 {return kernel_ret}
```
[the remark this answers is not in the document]
Yaxuan-w: I didn't fully understand what is 

> we may have increased the return number earlier and at this point don't want to return an error

Does that mean we should return `syscall_error` when kernel_ret < 0?

Yaxuan-w (approved):
The overall implementation logic makes sense to me! Great work!!

src/safeposix/syscalls/net_calls.rs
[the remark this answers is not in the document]
rennergade: I meant we may have found a ready descriptor that isnt a kernel socket before we do the internal select. If thats the case we still want to return that information.

rennergade (changes requested):
looks great, minor change request then good to merge

src/interface/comm.rs
```diff
@@ -2,10 +2,13 @@
 // //
 // //
 
+use crate::interface;
 use std::mem::size_of;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::fs::read_to_string;
 use std::str::from_utf8;
+use std::os::unix::io::{AsRawFd, RawFd};
```
rennergade, line 10: I think we don't actually need to import this. Can't we just use ints?

rennergade (changes requested):
One last thing, I think we dont need to import the RawFD type

src/interface/comm.rs
```diff
@@ -2,10 +2,13 @@
...
+use std::mem;
```
rennergade, line 11: also we import mem:size_of above, maybe combine and just import what you need?

rennergade (changes requested):

src/interface/comm.rs
yizhuoliang: nice catch

rennergade (approved):
Awesome. Great job!

Issue #223, opened 2024-05-03
Let's make file I/O more efficient


JustinCappos:
Currently there is a lot of unneeded code in our file system handling which is left over from RepyV2's file interface.  

One good thing to address to start with is to remove a lot of unneeded code in [Emulfile](https://github.com/Lind-Project/safeposix-rust/blob/a938eea07e02556b5f6cacfeeeaac27eb5e9766d/src/interface/file.rs#L102).  Since this is only being called by RustPOSIX with names like linddata.XXX, etc. and operates on a directory it is assumed to control, it really doesn't need most of this checking.  I propose this function is changed to only open the file descriptor and set up the object without making any other system calls.

One complication for this plan is how [lseek](https://github.com/Lind-Project/safeposix-rust/blob/pipe-return-test/src/safeposix/syscalls/fs_calls.rs#L935) relies on filesize, which is set in this function.  One option would be to simply lseek everytime this is called, but this would be inefficient because most lseeks in a program are likely within the file bounds.  

To have an efficient solution for this problem without recording the actual file size, I propose that filesize is changed to something like sizeisatleast and for lseek to be called to check if you seek to a position greater than sizeisatleast.  This way, a minimal number of lseek operations need to be performed on a file descriptor, while one will not seek past the end of file during a call to write, read, etc.

Of course ftruncate and truncate also need to be modified to update sizeisatleast.  fstat also should likely be modified in this case since it reads the actual file size from disk.

rennergade:
I believe [this PR](https://github.com/Lind-Project/safeposix-rust/pull/225) addressed a lot of this, but we still could definitely further clean-up file.rs

PR #225, opened 2024-05-04
Open/Close Performance/Style Fixes

rennergade:
## Description

Fixes # (issue)
This removes a lot of cruft in the interface open that was leftover from the RePy transition and was unesccesary as well as decreasing the performance of open. It also fixes a minor bug in logging in close that was also making some performance losses.

### Type of change

## How Has This Been Tested?
- RustPOSIX test suite
- Main Lind test suite
- All LAMP/IPC tests

## Checklist:

### DIFF
--- a/src/interface/file.rs
+++ b/src/interface/file.rs
@@ -31,8 +31,12 @@ pub fn removefile(filename: String) -> std::io::Result<()> {
     Ok(())
 }
 
-pub fn openfile(filename: String) -> std::io::Result<EmulatedFile> {
-    EmulatedFile::new(filename)
+pub fn openfile(filename: String, filesize: usize) -> std::io::Result<EmulatedFile> {
+    EmulatedFile::new(filename, filesize)
+}
+
+pub fn openmetadata(filename: String) -> std::io::Result<EmulatedFile> {
+    EmulatedFile::new_metadata(filename)
 }
 
 #[derive(Debug)]
@@ -49,7 +53,13 @@ pub fn pathexists(filename: String) -> bool {
 
 impl EmulatedFile {
 
-    fn new(filename: String) -> std::io::Result<EmulatedFile> {
+    fn new(filename: String, filesize: usize) -> std::io::Result<EmulatedFile> {
+
+        let f = OpenOptions::new().read(true).write(true).create(true).open(filename.clone()).unwrap();
+        Ok(EmulatedFile {filename: filename, fobj: Some(Arc::new(Mutex::new(f))), filesize: filesize})
+    }
+
+    fn new_metadata(filename: String) -> std::io::Result<EmulatedFile> {
 
         let f = OpenOptions::new().read(true).write(true).create(true).open(filename.clone()).unwrap();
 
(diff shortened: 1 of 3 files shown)


JustinCappos:
@Yaxuan-w You should take a look at this and see what you think.  This is meant to address #223 .

src/interface/file.rs
```diff
@@ -20,30 +20,8 @@ use std::ffi::c_void;
...
-    return names;
-}
 
 pub fn removefile(filename: String) -> std::io::Result<()> {
-    let openfiles = &OPEN_FILES;
-
-    if openfiles.contains(&filename) {
-        panic!("FileInUse");
-    }
-
```
JustinCappos, line 46: I'm not certain this isn't needed.  On Linux this is probably okay.  On Windows, I think this will error since you can't remove an open file.  Certainly, you could do something better then get the whole file list.

From what I see, we only use removefile internally, for our own use.  So are you able to just kill the function altogether?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1828,7 +1828,7 @@ impl Cage {
                     panic!("Somehow a normal file with an fd was truncated but there was no file object in rustposix?");
                 } else {
                     let sysfilename = format!("{}{}", FILEDATAPREFIX, inodenum);
-                    tempbind = interface::openfile(sysfilename, true).unwrap();
+                    tempbind = interface::openfile(sysfilename, filesize).unwrap(); // create new file
```
JustinCappos, line 1831: Should this be length or ulength?

JustinCappos (commented):
I don't yet feel qualified to review this.  I've added a few comments.  I'd suggest @Yaxuan-w takes a look...

As I understand it, because you either create, truncate, or read the metadata for a file, you should always know its exact size.

src/safeposix/syscalls/fs_calls.rs
rennergade: this is just creating a new blank file, the actual truncating is done further down.

JustinCappos: If it's creating a blank file, should filesize be 0?  Why does it read from the inode on L1813?

rennergade: Right I said that wrong. It's opening a file that we know is of length filesize from the inode. The truncating based on length/ulength is done further down.

src/interface/file.rs
rennergade: That make sense. We'll need to audit this code further in the upcoming PRs. For now the management we do in RustPOSIX regarding linkcount should be sane.

src/interface/file.rs
```diff
@@ -53,84 +31,45 @@ pub fn removefile(filename: String) -> std::io::Result<()> {
...
 
-        Ok(EmulatedFile {filename: filename, abs_filename: absolute_filename, fobj: Some(Arc::new(Mutex::new(f))), filesize: filesize as usize})
-
+        Ok(EmulatedFile {filename: filename, fobj: Some(Arc::new(Mutex::new(f))), filesize: filesize as usize})
     }
 
+
     pub fn close(&self) -> std::io::Result<()> {
-        OPEN_FILES.remove(&self.filename);
         Ok(())
```
Yaxuan-w, line 73: Our intention in retaining these is to avoid extensive changes to the logic in fs_calls.rs at this stage, correct?

PR #228, opened 2024-05-05
Removing toy benchmark, adding comparison with native system call times

JustinCappos:
## Description

Fixes # (issue)

 

### Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

## Checklist:


benches/rustposix-criterion.rs
```diff
@@ -46,14 +36,20 @@ fn get_dummy_cage() -> Cage {
 
 pub fn basic_rustposix_benchmark(c: &mut Criterion) {
 
-    // First, just do a basic benchmark to show the runner works...
-    c.bench_function("fib 20", |b| b.iter(|| fibonacci(black_box(20))));
+    // I'm not setting this up properly because I think I don't need that much
+    // state.  I just want to call system calls.
     let a = get_dummy_cage();
```
rennergade: In our tests we do lindrustinit/finalize to setup and shutdown, this will create an initial cage. That may work here as well unless you're trying to test for a specific state. Not partial to either, just pointing out the difference.

rennergade (commented):
I like this a lot. Seems self-explanatory and very useful. I didn't see any necessary changes and just left a minor suggestion.

Yaxuan-w (approved):
Code is easy to understand and test makes sense to me

yzhang71 (approved):
LGTM!

JustinCappos:
> I left a minor suggestion.

I added a fix to address this in my latest commit.  I also added a few more tests.  

Could someone take a look and if it's okay, go ahead and merge?  Please be warned, that I really don't get some aspects of Rust or Lind file system initialization + finalization so, please double check for bugs...

benches/rustposix-criterion.rs
```diff
@@ -1,68 +1,157 @@
...
+    }
+
+    // For comparison let's time the native OS...
+    group.bench_function("Native OS kernel read+write", |b| b.iter(|| 
+        {
+            unsafe{
+                let _ = libc::write(fd,CString::new("Well, hello there!!!").unwrap().as_ptr() as *const c_void,20);
+                let mut read_buffer = sizecbuf(20);
+                libc::lseek(fd,0,SEEK_SET);
+                cage.read_syscall(fd,read_buffer.as_mut_ptr(), 20);
```
Yaxuan-w: If we want to test Native OS, would using `libc::read(fd, read_buffer.as_mut_ptr(), 20)` makes more sense here?

JustinCappos:
```suggestion
                libc::read(fd,read_buffer.as_mut_ptr() as *mut c_void, 20);
```

JustinCappos: Thanks, fixed!

benches/rustposix-criterion.rs
```diff
@@ -1,68 +1,157 @@
-use criterion::{black_box, criterion_group, criterion_main, Criterion};
+/* Benchmarks for the microvisor implementation.  In general, I'm not doing
+ * results checking / assertations to avoid adding bias to the results.  */
+
+
+// I hate allowing this, but this is apparently a known issue for a lot of
```
rennergade: There are some helper functions in tests/mod.rs that I believe would help you do these more safely.

benches/rustposix-criterion.rs
```diff
@@ -1,68 +1,157 @@
...
+
+    unsafe {
+        fd = libc::open(CString::new("/tmp/foo").unwrap().as_ptr(),O_CREAT | O_TRUNC | O_WRONLY,S_IRWXA);
+    }
+
+    // For comparison let's time the native OS...
+    group.bench_function("Native OS kernel read+write", |b| b.iter(|| 
+        {
+            unsafe{
+                let _ = libc::write(fd,CString::new("Well, hello there!!!").unwrap().as_ptr() as *const c_void,20);
```
rennergade: I think in general declaring the strings outside of the unsafe blocks/function signature would also help with safety/clarity

benches/rustposix-criterion.rs
JustinCappos: I think I addressed this.  I use those functions now.

benches/rustposix-criterion.rs
JustinCappos: I tried what you recommended and it doesn't remove the need to have: #![allow(temporary_cstring_as_ptr)]

To me, it doesn't matter one way or the other, but I'm not enough in touch with what the right style is to really judge.

`fd = libc::open(CString::new("/tmp/foo").unwrap().as_ptr(),O_CREAT | O_TRUNC | O_WRONLY,S_IRWXA);`

versus

```
    let filename = CString::new("/tmp/foo").unwrap().as_ptr();
    unsafe {
        fd = libc::open(filename,O_CREAT | O_TRUNC | O_WRONLY,S_IRWXA);
    }
```

seems pretty similar to me.

JustinCappos: Okay, I've updated the PR to move to the str2cbuf(...) function instead everywhere.  (Loosely related.)

JustinCappos:
@rennergade This should be ready now.  Please squash and merge if you think it's ready.

rennergade (approved):
Awesome this looks great.

PR #231, opened 2024-05-07
Add RustPOSIX CI Action

rennergade:
## Description

Mock up for a simple RustPOSIX only CI to run `cargo test --lib` on changes/PRs.

### Type of change

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,22 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "develop" ]
+  pull_request:
+    branches: [ "develop" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


JustinCappos:
Looks good to me!

JustinCappos (approved):

.github/workflows/rust.yml
```diff
@@ -0,0 +1,22 @@
...
+    branches: [ "develop" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
```
naveensrinivasan, line 16: The rust toolchain has to be installed.

rennergade: Interesting! My assumption was since this is github hosted and this was basically their default template that was handled. Is that incorrect?

naveensrinivasan: This could be a good option https://github.com/actions-rust-lang/setup-rust-toolchain

PR #232, opened 2024-05-08
parameterized r/w

JustinCappos:
## Description

Adds parameters to read and write calls in the file system, varying the buffer sizes.  

### Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Run "cargo bench TF02" and "cargo bench TF03" to test

## Checklist:

### DIFF
--- a/benches/fs_read_write.rs
+++ b/benches/fs_read_write.rs
@@ -92,9 +92,10 @@ pub fn run_benchmark(c: &mut Criterion) {
     for buflen in [1,64,1024,65536].iter() {
 
         let fd: c_int;
-
+        let c_str = CString::new("/tmp/foo").unwrap();
+        let path = c_str.into_raw() as *const i8;
         unsafe {
-            fd = libc::open(tests::str2cbuf("/tmp/foo"),O_CREAT | O_TRUNC | O_WRONLY,S_IRWXA);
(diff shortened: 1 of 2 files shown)


benches/fs_read_write.rs
```diff
@@ -33,68 +35,101 @@ pub fn run_benchmark(c: &mut Criterion) {
...
-            cage.read_syscall(fd,read_buffer.as_mut_ptr(), 20);
-        }
-    ));
+    // First do this for Lind
 
-    cage.close_syscall(fd);
-    cage.unlink_syscall("foo");
+    // Reduce the time to reduce disk space needed and go faster.  
+    // Default is 5s...
+    group.measurement_time(Duration::from_secs(2));
```
rennergade, line 52: If we're going to configure these groups the same for every benchmark it might make sense to factor something out for reuse.

rennergade (commented):
This looks great. I don't see anything wrong with the implementation. Left a minor possible suggestion.

Yaxuan-w (approved):
The overall logic makes sense to me, and submit a pr to fix type mismatch error

benches/fs_read_write.rs
JustinCappos: We're not.  It is just for these functions.  (And there is already a [place where the refactored bits go](https://github.com/Lind-Project/safeposix-rust/blob/develop/benches/global_criterion_settings.rs).)

PR #233, opened 2024-05-08
Select/Poll fixes for performance/signal handling

rennergade:
## Description
Makes select faster by yielding instead of sleeping
Properly checks for signal in Poll()

### Type of change

## How Has This Been Tested?

Rust Suite
Lind Suite
LAMP tests


Yaxuan-w (approved):
Slightly changes and logic is easy to understand. LGTM so approved.

PR #235, opened 2024-05-08
Update Rust CI

rennergade:
## Description
Update Rust CI to build with nightly

Fixes # (issue)

Hopefully this fixes CI build issue by using nightly. Open for comments.

### Type of change


JustinCappos (approved):
I'm not the best to review this as I don't fully understand this workflow, but because we're hoping to fix something which is already broken, I think it's sane to approve and merge.

PR #239, opened 2024-05-10
clarifing octal and fixing NONBLOCK bug

JustinCappos:
## Description

Fixes incorrect constants due to confusion primarily about how Rust handles Octal values.  This is partially related to [the clippy issue on the main tracker](https://github.com/Lind-Project/lind_project/issues/373#issuecomment-2102774715).

In Rust, an octal must be written as 0oXXXX, instead of C where you can write 0XXXX.  This is the heart of most of this confusion.

While fixing this, I also noticed an error with the SOCK_NONBLOCK flag.  Did I mention we need more testing?  😁 

### Type of change

## How Has This Been Tested?

I don't have a complete environment running locally so didn't test.  I did double check the constants in both glibc and NaCl to make sure they are correct.

## Checklist:


rennergade (approved):
Great catch. This should save us some headaches in the future and doesn't introduce any new weird behavior.

Thanks!

Issue #242, opened 2024-05-11
Benchmarks have errors / problems...


JustinCappos:
When examining the benchmarks for rustposix, I noticed a few odd measurements that led me to find a few bugs in the benchmark scripts.

The overall problem is that return values were not being adequately validated.  (Naively, I thought this was a good idea to limit the overhead of this checking in the benchmark.)  This was masking bugs.  

A few examples: 

In many benchmarks, the file was opened write-only.  This means that reads were faster than writes, with Lind far outperforming the kernel because it would error more quickly.

In the (common) case where writes are slower than reads, this means that a benchmark that writes into a file and then reads from it, can easily read to the end of the file and keep requesting bytes.  This means iterations at the end will just measure the size of a 0 byte read, regardless of the actual read buffer size.

So, as a result, I will fix those issues in PRs and also update the style guidelines to require assertions in almost all cases so that these sorts of errors can be minimized in the future.

Issue #241, opened 2024-05-11
lseek wraps around!


JustinCappos:
The implementation of lseek has a bug where the data type (i32) isn't large enough and so it will wrap around on large files.

See https://github.com/Lind-Project/safeposix-rust/commit/678ffb29918a8667d7aa0a33d18bf07056cb887b#diff-97744cd8def1974ce0c338f5dedaa20ce129fd3b1a07b75cddff8eeff3e96b0eR75-R76 for an example.

This should be changed to i64, which will not overflow in practice, at least with current storage sizes.   Note that Rust's [std::io implementation uses an u64](https://doc.rust-lang.org/std/io/trait.Seek.html).  POSIX's type is system dependent (because, of course it is), but is [usually a i64](https://man7.org/linux/man-pages/man2/lseek.2.html).  Note that POSIX also will return -1 and set errno to EOVERFLOW if you go past the end and would wrap around.

PR #243, opened 2024-05-11
Fix benchmarks

JustinCappos:
## Description

Partial fix for #242 

This fixes the two example problems listed in #242.  However, #241 is not addressed here (which is noted in the code along with instructions about how to change the code after fixing this bug).

Furthermore, a bug still exists leading to some memory corruption issues.  (Notice the XXXXXXX in the output below is indicative that a string with the test name was overwritten by the [construction of deststring](https://github.com/Lind-Project/safeposix-rust/blob/fix_benchmarks/benches/fs_read_write.rs#L53-L55).)

I would appreciate help to track down the memory corruption issue.

`Compare fs:write+read+lseek/TF03:Lind write+read+lseek/1                        
                        time:   [1.0006 µs 1.0080 µs 1.0200 µs]
                        change: [+2.9263% +3.7612% +4.6903%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) high mild
  7 (7.00%) high severe
Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/64: Warming                                                                                 Benchmarking XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX: Collecti                                                                                Benchmarking XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX: Analyzin                                                                                XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                        time:   [1.0222 µs 1.0229 µs 1.0236 µs]
                        change: [+3.1119% +4.1536% +5.9263%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe
Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/1024: Warmin                                                                                Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/1024: Collec                                                                                Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/1024: Analyz                                                                                Compare fs:write+read+lseek/TF03:Lind write+read+lseek/1024
                        time:   [1.0780 µs 1.0958 µs 1.1209 µs]
                        change: [+6.9239% +21.318% +44.944%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 21 outliers among 100 measurements (21.00%)
  4 (4.00%) high mild
  17 (17.00%) high severe
Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/65536: Warmi                                                                                Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/65536: Colle                                                                                Benchmarking Compare fs:write+read+lseek/TF03:Lind write+read+lseek/65536: Analy                                                                                Compare fs:write+read+lseek/TF03:Lind write+read+lseek/65536
                        time:   [5.4057 µs 5.4267 µs 5.4614 µs]
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  8 (8.00%) high severe
error: bench failed, to rerun pass '--bench fs_read_write_seek'

Caused by:
  process didn't exit successfully: '/home/test/safeposix-rust/target/release/deps/fs_read_write_seek-adecf1c62a8ef04d --bench' (signal: 11, SIGSEGV: invalid memory reference)
`

### Type of change

## How Has This Been Tested?

Run on docker on my mac using cargo bench

## Checklist:

### DIFF
--- a/benches/fs_open_close.rs
+++ b/benches/fs_open_close.rs
@@ -43,8 +43,8 @@ pub fn run_benchmark(c: &mut Criterion) {
     group.bench_function("TF01: Lind open+close", |b| {
         b.iter(|| {
             let fd = cage.open_syscall("foo", O_CREAT | O_TRUNC | O_WRONLY, S_IRWXA);
-            assert!(fd>2); // Ensure we didn't get an error or an odd fd
-            assert_eq!(cage.close_syscall(fd),0);  // close the file w/o error
+            assert!(fd > 2); // Ensure we didn't get an error or an odd fd
+            assert_eq!(cage.close_syscall(fd), 0); // close the file w/o error
         })
     });
 
@@ -56,8 +56,8 @@ pub fn run_benchmark(c: &mut Criterion) {
                 O_CREAT | O_TRUNC | O_WRONLY,
                 S_IRWXA,
             );
-            assert!(fd>2); // Ensure we didn't get an error or an odd fd
-            assert_eq!(libc::close(fd),0);  // close the file w/o error
+            assert!(fd > 2); // Ensure we didn't get an error or an odd fd
+            assert_eq!(libc::close(fd), 0); // close the file w/o error
         })
     });
     group.finish();
--- a/benches/fs_open_close.rs
+++ b/benches/fs_open_close.rs
@@ -43,7 +43,8 @@ pub fn run_benchmark(c: &mut Criterion) {
     group.bench_function("TF01: Lind open+close", |b| {
         b.iter(|| {
             let fd = cage.open_syscall("foo", O_CREAT | O_TRUNC | O_WRONLY, S_IRWXA);
-            cage.close_syscall(fd);
+            assert!(fd>2); // Ensure we didn't get an error or an odd fd
+            assert_eq!(cage.close_syscall(fd),0);  // close the file w/o error
         })
     });
 
@@ -55,7 +56,8 @@ pub fn run_benchmark(c: &mut Criterion) {
                 O_CREAT | O_TRUNC | O_WRONLY,
                 S_IRWXA,
             );
-            libc::close(fd);
+            assert!(fd>2); // Ensure we didn't get an error or an odd fd
+            assert_eq!(libc::close(fd),0);  // close the file w/o error
         })
     });
     group.finish();
--- a/benches/fs_read_write.rs
+++ b/benches/fs_read_write.rs
@@ -177,7 +177,6 @@ pub fn run_benchmark(c: &mut Criterion) {
             // reset the file position
             libc::lseek(fd, 0, SEEK_SET);
         }
-        println!("FILE LENGTH: {}", file_length);
 
         // My current position when reading...
         let mut pos = 0;
@@ -197,7 +196,6 @@ pub fn run_benchmark(c: &mut Criterion) {
                     if file_length <= pos {
                         libc::lseek(fd, 0, SEEK_SET);
                         pos = 0;
-                        print!(".");
                     }
                     assert_eq!(
                         libc::read(fd, read_buffer.as_mut_ptr() as *mut c_void, *buflen),
--- a/benches/fs_read_write.rs
+++ b/benches/fs_read_write.rs
@@ -54,7 +54,7 @@ pub fn run_benchmark(c: &mut Criterion) {
             &String::from_utf8(vec![b'X'; *buflen]).expect("error building string"),
         );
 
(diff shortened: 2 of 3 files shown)


JustinCappos:
This memory corruption issue happens even when I comment out all of the lind code and just use the [Native part of the benchmark](https://github.com/Lind-Project/safeposix-rust/blob/fix_benchmarks/benches/fs_read_write_seek.rs#L84-L134).  

I don't fully understand this issue, but I'm quite suspicious of [tests::size2cbuf](https://github.com/Lind-Project/safeposix-rust/blob/fix_benchmarks/src/tests/mod.rs#L84C1-L88C2) from our code.  [This discussion](https://www.reddit.com/r/rust/comments/7z9qv3/comment/dunffyg/) also seems to indicate the problem may be with how vectors are being used.

JustinCappos:
Note, the problem also occurs when I replace the file I/O, etc. with std::ptr::copy_nonoverlapping(deststring, read_buffer, *buflen);

A minimal example that replicates this is:

```
    for buflen in [1, 64, 1024, 65536].iter() {
        let deststring = tests::str2cbuf(
            &String::from_utf8(vec![b'X'; *buflen]).expect("error building string"),
        );
        let read_buffer = tests::sizecbuf(*buflen).as_mut_ptr();
        group.bench_with_input(
            BenchmarkId::new("TF03:Native write+read+lseek", buflen),
            buflen,
            |b, buflen| {
                b.iter(|| unsafe {
                    std::ptr::copy_nonoverlapping(deststring, read_buffer, *buflen);
                })
            },
        );
    }

```

JustinCappos:
It still crashes when replacing:
```let read_buffer = tests::sizecbuf(*buflen).as_mut_ptr();```
with
```
let mut c: Vec<u8> = Vec::new();
let mut read_buffer: *mut u8 = c.as_mut_ptr();
```

**Comment by JustinCappos** [comment]
Okay, this fixes it:

```
        let mut actual_read_buffer = [0u8;65536];
        let mut read_buffer: *mut u8 = actual_read_buffer.as_mut_ptr();
```

My understanding is that this is because the bytes are actually allocated here.  

As I understand it, my previous attempt with Vec.new() doesn't work because [Rust doesn't allocate memory for Vectors until it is needed](https://doc.rust-lang.org/nomicon/vec/vec-alloc.html).  

However, I'm still confused why https://github.com/Lind-Project/safeposix-rust/blob/fix_benchmarks/src/tests/mod.rs#L84C1-L88C2 doesn't work.  From [what I can see](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#1139), it seems like this should still be a valid part of memory to access.  Is this falling out of scope and being garbage collected when it shouldn't?  

(I also need to get far enough through the Rust Book so I understand why a Box may do this.  To me, this just seems like a malloced area on the heap now...)

**Comment by JustinCappos** [comment]
Here is a minimal test case which shows the problem:

```
pub fn sizecbuf<'a>(size: usize) -> Box<[u8]> {
    let v = vec![0u8; size];
    v.into_boxed_slice()
}

fn main() {
    
    // Iterate to trigger the bug...
    for buflen in [1, 16, 64, 256, 1024, 4096, 65536].iter() {
        let src_vec: *mut u8 = vec![b'X'; *buflen].as_mut_ptr();

        let dest_buffer = sizecbuf(*buflen).as_mut_ptr();

// Alternative that also has the problem
/*        let v = vec![0u8; *buflen];
        let mut new_v:Box<[u8]> = v.into_boxed_slice();
        let mut dest_buffer = new_v.as_mut_ptr();  */

// Alternative which avoids the issue
//       let mut actual_dest_buffer = [0u8;65536];
//        let dest_buffer: *mut u8 = actual_dest_buffer.as_mut_ptr();


        for _ in 0..10 {
            unsafe {
               std::ptr::copy_nonoverlapping(src_vec, dest_buffer, *buflen);
            }
        }

        // Having a println in here makes it trigger more reliably
        println!("hi");
    }

}
```

So, I know how to fix this, but I'd like to understand why this is an issue.  I would have thought 'Box'ing would allocate memory on the heap and was a safe way to do this.

PR #247, opened 2024-05-19
Socket syscall

rupeshkoushik07:
## Description
This is an example pull request showing how to add comments for syscalls, in this case, for the socket_syscall function.
unit tests are updated by @yashaswi2000 
Fixes # (issue)

 

### Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -74,14 +74,36 @@ impl Cage {
         }
         0
     }
-//Creates a new socket, ensuring the requested domain, socket type, and protocol are supported by SafePOSIX and also
-//validates the requested communication domain, socket type, and protocol permitting only combinations that are known
+
+/// ## `socket_syscall`
+///
+/// ### Description
+/// This function creates a new socket, ensuring the requested domain, socket type, 
+/// and protocol are supported by SafePosix. 
+/// It validates the requested communication domain, socket type, and protocol, permitting only combinations that are known 
+/// to be safe and secure.
+///
+/// ### Function Arguments
+/// * `domain`: The communication domain for the socket. Supported values are 
+///   `PF_INET` (Internet Protocol) and `PF_UNIX` (Unix domain sockets).
+/// * `socktype`: The socket type. Supported values are `SOCK_STREAM` (stream sockets) and `SOCK_DGRAM` (datagram sockets).
+/// * `protocol`: The protocol to use for communication. This defaults to TCP for stream sockets 
+///   (`SOCK_STREAM`) and UDP for datagram sockets (`SOCK_DGRAM`).
+///
+/// ### Returns
+/// * The new file descriptor representing the socket on success.
+///
+/// ### Errors
+/// * `EOPNOTSUPP(95)`: If an unsupported combination of domain, socket type, or protocol is requested.
+/// * `EINVAL(22)`: If an invalid combination of flags is provided.
     pub fn socket_syscall(&self, domain: i32, socktype: i32, protocol: i32) -> i32 {
         let real_socktype = socktype & 0x7; //get the type without the extra flags, it's stored in the last 3 bits
         let nonblocking = (socktype & SOCK_NONBLOCK) != 0; // Checks if the socket should be non-blocking.
-        //Check blocking status for storage in the file descriptor, we'll need this for calls that don't access the kernel socket, unix sockets, and properly directing kernel calls for recv and accept
+        //Check blocking status for storage in the file descriptor, we'll need this for calls that don't access the kernel 
+        //socket, unix sockets, and properly directing kernel calls for recv and accept
         let cloexec = (socktype & SOCK_CLOEXEC) != 0;
-        // Checks if the 'close-on-exec' flag is set. This flag ensures the socket is automatically closed if the current process executes another program, preventing unintended inheritance of the socket by the new program.
+        // Checks if the 'close-on-exec' flag is set. This flag ensures the socket is automatically closed if the current
+        // process executes another program, preventing unintended inheritance of the socket by the new program.
        
         // additional flags are not supported
         // filtering out any socktypes with unexpected flags set.
@@ -95,7 +117,8 @@ impl Cage {
                 "Invalid combination of flags"
             );
         }
-        //SafePOSIX intentionally supports only a restricted subset of socket types . This is to make sure that applications not creating other socket types which may lead to security issues. 
+        //SafePOSIX intentionally supports only a restricted subset of socket types . This is to make sure that 
+        // applications not creating other socket types which may lead to security issues. 
         //By using the match statement, SafePOSIX ensures that only these approved socket types are allowed.
         match real_socktype {// Handles different socket types SOCK_STREAM or SOCK_DGRAM in this cases
             SOCK_STREAM => {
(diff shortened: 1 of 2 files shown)


src/safeposix/syscalls/net_calls.rs
```diff
@@ -74,13 +74,15 @@ impl Cage {
         }
         0
     }
-
+//Creates a new socket, ensuring the requested domain, socket type, and protocol are supported by SafePOSIX and also
+//validates the requested communication domain, socket type, and protocol permitting only combinations that are known
     pub fn socket_syscall(&self, domain: i32, socktype: i32, protocol: i32) -> i32 {
         let real_socktype = socktype & 0x7; //get the type without the extra flags, it's stored in the last 3 bits
-        let nonblocking = (socktype & SOCK_NONBLOCK) != 0;
+        let nonblocking = (socktype & SOCK_NONBLOCK) != 0; // Checks if the socket should be non-blocking.
```
JustinCappos, line 103: Why do we check this though?  What is the purpose of this check?  Why do we care if this is set?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -74,13 +74,15 @@ impl Cage {
         }
         0
     }
-
+//Creates a new socket, ensuring the requested domain, socket type, and protocol are supported by SafePOSIX and also
+//validates the requested communication domain, socket type, and protocol permitting only combinations that are known
     pub fn socket_syscall(&self, domain: i32, socktype: i32, protocol: i32) -> i32 {
         let real_socktype = socktype & 0x7; //get the type without the extra flags, it's stored in the last 3 bits
```
JustinCappos, line 102: What if there are other bits set here?  what happens?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -74,13 +74,15 @@ impl Cage {
...
-        let nonblocking = (socktype & SOCK_NONBLOCK) != 0;
+        let nonblocking = (socktype & SOCK_NONBLOCK) != 0; // Checks if the socket should be non-blocking.
         let cloexec = (socktype & SOCK_CLOEXEC) != 0;
+        // Checks if the 'close-on-exec' flag is set. This flag ensures the socket is automatically closed if the current process executes another program, preventing unintended inheritance of the socket by the new program.
 
-        match real_socktype {
+        match real_socktype {// Handles different socket types SOCK_STREAM or SOCK_DGRAM in this cases
```
JustinCappos, line 125: I can see that the comment says what it does, but why do we check this?  What is the purpose?

JustinCappos (changes requested):
A few comments in the text below to try to help make the comments better.

Where are the unit tests and benchmarks for this?

src/tests/networking_tests.rs
```diff
@@ -1299,26 +1299,27 @@ pub mod net_tests {
...
+        assert!(sockfd > 0, "Expected a valid file descriptor, got error");
+
         let sockfd2 = cage.socket_syscall(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+        assert!(sockfd2 > 0, "Expected a valid file descriptor, got error");
 
         let sockfd3 = cage.socket_syscall(AF_INET, SOCK_DGRAM, 0);
+        assert!(sockfd3 > 0, "Expected a valid file descriptor, got error");
+        
         let sockfd4 = cage.socket_syscall(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+        assert!(sockfd4 > 0, "Expected a valid file descriptor, got error");
```
JustinCappos, line 1356: A few more things to test: 

Should the sockets be the same or different?  

Can you check any properties of them?

Can you test to ensure that (at least a few) types of incorrect requests for sockets lead to errors of the right types too?

JustinCappos (changes requested):

src/safeposix/syscalls/net_calls.rs
rupeshkoushik07: If there are other bits set here it can result in socktype value that don't correspond to any defined socket type or flag.

src/safeposix/syscalls/net_calls.rs
rupeshkoushik07: SafePOSIX intentionally supports only a restricted subset of socket types . This is to make sure that applications not creating other socket types which may lead to security issues. By using the match statement, SafePOSIX ensures that only these approved socket types are allowed.

src/safeposix/syscalls/net_calls.rs
rupeshkoushik07: Few applications require blocking sockets (where they can wait for an operation to complete) and non-blocking sockets(where they can't afford to wait ).
This is a necessary check; SafePOSIX wouldn't know the application's intended socket behaviour, and the application might fall into a busy-wait loop, resulting in excessive CPU resources.

JustinCappos: Okay, but why does SafePOSIX need this information?  The kernel will have it because the flags will be passed down.  Why do we also check it?

src/safeposix/syscalls/net_calls.rs
JustinCappos: okay, please add a comment to clarify this.

JustinCappos (approved):
Looks good!  Thanks for the changes.

src/safeposix/syscalls/net_calls.rs
rupeshkoushik07: Relying solely on the kernel to handle non-blocking behaviour without considering the specific context of the sandboxed environment can undermine the control, security, and efficiency that SafePOSIX is designed to provide.

By performing its own SOCK_NONBLOCK check, SafePOSIX ensures consistent behaviour and maintains its configured operation—this added layer of control and security safeguards against potential issues within its sandboxed environment.

rupeshkoushik07: ill update this in the comments

JustinCappos: >By performing its own SOCK_NONBLOCK check, SafePOSIX ensures consistent behaviour and maintains its configured operation—this added layer of control and security safeguards against potential issues within its sandboxed environment.

Okay, be sure to explain how RustPOSIX does this.  (Note, I'm not sure what you're saying here is true.  It may be, but it also may not be.  It's fine to say you don't know why something happens.  As you go through more of the code, you may figure it out and then update your comment...)

rupeshkoushik07: I wasn't precise before. I've now understood why SafePOSIX requires this and have made the necessary updates.

JustinCappos (approved):
Looks good!

rennergade:
I think it makes sense to merge this and just keep a link to it as reference in the documentation issue rather than keeping it open. 

@rupeshkoushik07 can you update this for the newer RustDoc and we'll get it approved?

rupeshkoushik07:
I have updated the comments @rennergade

rennergade (changes requested):
Need to update errors/panics

rennergade:
lets also change the name of this PR to socket comments or something like that so its not confusing once we merge

rennergade:
This looks good and up to date and IDT we need this as a non-merged example anymore. Going to merge now.

PR #249, opened 2024-05-21
readat/writeat

rennergade:
… call

## Description
change read and write to use readat and writeat, removing excess seek call

Fixes # (issue)
Minor performance improvements

### Type of change

## How Has This Been Tested?

Rust suite/lamp tests


Yaxuan-w (approved):
`read_at` and `write_at` will read data directly from the specified offset of the file. Change to use them will avoid using `seek()` and make the code more clear. The overall logic makes sense to me so approved.

JustinCappos (changes requested):
This is also quite confusing with readat calling read_at, etc.  More comments about why you are doing what you are doing / code clarity would be a big help here.

JustinCappos (approved):

PR #256, opened 2024-06-03
added more tests for socketpair syscall and fixed an issue with prior socketpair test

qianxichen233:
## Description

Fixes # (issue)

This PR added more tests for the socketpair syscall, and fixed an issue for prior socketpair test that could stuck the entire test in some cases, especially, when a test inside a non-main thread failed, the thread will panic and the main thread will never be able to return from recv_syscall as it is waiting message from an already dead thread. Fix involves creating dedicated thread for each peer and using non-blocking thread join method. Besides the fix of the old test, below is the new tests cased being added:
1. socketpair_syscall called with SOCK_NONBLOCK flag set
2. socketpair_syscall called with SOCK_CLOEXEC flag set
3. socketpair_syscall called with bad inputs to test for error code

### Type of change

## How Has This Been Tested?

This PR is a test itself, and currently all the tests were passed.

## Checklist:


src/tests/networking_tests.rs
```diff
@@ -1518,22 +1520,33 @@ pub mod net_tests {
             );
         });
 
-        assert_eq!(
-            cage.send_syscall(socketpair.sock1, str2cbuf("test"), 4, 0),
-            4
-        );
+        let cage3 = cage.clone();
```
Yaxuan-w, line 1586: I'm curious why we use `clone()` here instead of `fork_syscall()`?

Yaxuan-w (commented):
Great job! Tests are comprehensive. LGTM with a minor question

src/tests/networking_tests.rs
qianxichen233: We are not really creating a new process here, the `cage` object here is `Arc<Cage>` instead of `Cage`. I need to clone this object so that I can `move` it to the scope of the new thread and use it.

src/tests/networking_tests.rs
```diff
@@ -8,28 +8,30 @@ pub mod net_tests {
     use std::sync::{Arc, Barrier};
 
     pub fn net_tests() {
-        ut_lind_net_bind();
```
JustinCappos: Was this due to 'cargo fmt' or did you do this for some reason?

Also, I can see the writev test was removed.  Why?

src/tests/networking_tests.rs
```diff
@@ -1518,22 +1520,33 @@ pub mod net_tests {
...
+            let str2 = cbuf2str(&buf2);
+            assert_eq!(str2, "Socketpair Test");
+        });
 
-        let mut buf2 = sizecbuf(15);
         loop {
-            let result = cage.recv_syscall(socketpair.sock1, buf2.as_mut_ptr(), 15, 0);
-            if result != -libc::EINTR {
-                break; // if the error was EINTR, retry the syscall
+            if thread_2.is_finished() {
```
JustinCappos: Why do you check this and have separate if statements for each?  Why not just join both?  Does the order matter for some reason?

Join seems to block (https://doc.rust-lang.org/std/thread/index.html), so I'm not sure why it needs to work like this...

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1557,156 @@ pub mod net_tests {
...
+
+        assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
+    pub fn ut_lind_net_socketpair_closexec() {
+        lindrustinit(0);
+        let cage = interface::cagetable_getref(1);
+        let mut socketpair = interface::SockPair::default();
+        // try with closexec flag
```
JustinCappos:
```suggestion
        // try with cloexec flag
```

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1557,156 @@ pub mod net_tests {
...
+        assert_eq!(
+            Cage::socketpair_syscall(cage.clone(), AF_UNIX, 472810394, 0, &mut socketpair),
+            -(Errno::EOPNOTSUPP as i32)
+        );
+
+        assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
+    pub fn ut_lind_net_socketpair_closexec() {
```
JustinCappos: I'd recommend changing this to cloexec also...

JustinCappos (changes requested):
Looks good overall!  A few minor comments / request for changes...

src/tests/networking_tests.rs
qianxichen233: Yeah, the order of join matters here. Let's say if I do not care about the order of join, and join thread 1 first, then thread 2. Then if thread 2 failed the test and panic, while thread 1 is still waiting for message from thread 2 in recv_syscall, it would cause the same issue as I mentioned before (entire test will stuck). So I am using non-blocking join method to join these two threads (or at least the first returned thread), so that if any of the thread failed the test and panic, thread.is_finished() will return true, then when I try to join this panic thread, it will return an Error so the test could be aborted immediately

src/tests/networking_tests.rs
qianxichen233: sorry this is a mistake and I will fix it later
The main reason for the weird thing here is that when I create my own branch, I seems to forked from the old develop branch that missed some latest commits, including the rustfmt commit at May 11. And after I finished my branch and try to make a pull request, I noticed the issue and try to rebase from the latest develop branch. And when I resolve the conflict for `net_tests` function, I did not look very closely and thought my `net_tests` was newest so I overwrote it with my change, which left it like this

src/tests/networking_tests.rs
JustinCappos: Okay, please explain this in the comments in the test...

yashaswi2000:
Good work on the tests! , Needs more documentation on what is intended and how we are testing. Missing comments on the original Socketpair syscall in net_calls.rs, and potential error checking improvements there.

src/tests/networking_tests.rs
JustinCappos: If you've fixed these, please "resolve conversation"  If all are done and you're ready for another review, please add a comment or similar to say so.

src/tests/networking_tests.rs
qianxichen233: Had a discussion with Nick this week and I decided to migrate my old solution that uses non-blocking join method to panic hook method, which is much more convenient and clean.
I've also added some comments to socketpair_syscall

qianxichen233:
@davidge20 Could you review my socketpair PR?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2834,6 +2834,40 @@ impl Cage {
...
+    ///                2. SOCK_DGRAM
+    ///                      Supports datagrams (connectionless, unreliable messages of a fixed maximum length).
+    ///                The type argument serves a second purpose: in addition to specifying a socket type, it may include the bitwise
+    ///                OR of any of the following values, to modify the behavior of the socket:
+    ///                1. SOCK_NONBLOCK
+    ///                      Set the O_NONBLOCK file status flag on the open file description referred to by the new file
+    ///                      descriptor.
+    ///                2. SOCK_CLOEXEC
+    ///                      Set the close-on-exec flag on the new file descriptor.
+    /// * `protocol` - The protocol specifies a particular protocol to be used with the socket. Currently only support the default protocol.
```
davidge20: please mention the "default protocol"

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2861,6 +2896,7 @@ impl Cage {
             );
```
davidge20, line 3168: please explicitly explain the bit-wise logic on the else if segment and why its needed

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2834,6 +2834,40 @@ impl Cage {
         }
     }
 
+    /// ## ------------------SOCKETPAIR SYSCALL------------------
+    /// ### Description
+    /// The `socketpair_syscall()` call creates an unnamed pair of connected sockets in the specified domain, of the specified type, and using
+    /// the optionally specified protocol.
+    /// The file descriptors used in referencing the new sockets are returned in sv[0] and sv[1]. The two sockets are
```
davidge20: while the man page does say the new sockets are in sv[0] and sv[1], it seems to be worth changing that to sv.sock1 and sv.sock2 as thats our implementation

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2861,6 +2896,7 @@ impl Cage {
             );
         }
 
+        // get the flags
         let nonblocking = (socktype & SOCK_NONBLOCK) != 0;
         let cloexec = (socktype & SOCK_CLOEXEC) != 0;
```
davidge20, line 3179: below i see the following comment:

"// we need to increment the refcount of the sockets we created
  // reason: in bind_inner_socket, we added entries to the inode table"
  
  please add some details to the comment. specifically what is a refcount of the sockets. I think this would make it clearer as to why we need to increment it if we add those entries to the inode table

src/tests/mod.rs
```diff
@@ -72,6 +74,18 @@ mod main_tests {
...
     }
+
+    fn set_panic_hook() {
+        let orig_hook = std::panic::take_hook();
+        std::panic::set_hook(Box::new(move |panic_info| {
+            // this hook would be triggered whenever a panic occurs
+            // good for test cases that panicked inside the non-main thread
+            // so the trace information could be printed immediately
+            // instead of raising the error when the thread is joined, which might
+            // never happen and left the test blocking forever in some test cases
```
davidge20, line 100: great explanation here!

src/tests/networking_tests.rs
```diff
@@ -1492,6 +1495,10 @@ pub mod net_tests {
     }
 
     pub fn ut_lind_net_socketpair() {
```
davidge20, line 1551: good additions to the test!

src/tests/networking_tests.rs
```diff
@@ -1502,13 +1509,15 @@ pub mod net_tests {
...
             loop {
                 let result = cage2.recv_syscall(socketpair.sock2, buf.as_mut_ptr(), 10, 0);
                 if result != -libc::EINTR {
                     break; // if the error was EINTR, retry the syscall
                 }
             }
+            // check if received message is correct
             assert_eq!(cbuf2str(&buf), "test\0\0\0\0\0\0");
 
             interface::sleep(interface::RustDuration::from_millis(30));
```
davidge20: please state why its necessary to sleep here for some amount of time

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
         lindrustfinalize();
     }
 
+    pub fn ut_lind_net_socketpair_bad_input() {
+        // test for error cases of socketpair
+        lindrustinit(0);
+        let cage = interface::cagetable_getref(1);
+        let mut socketpair = interface::SockPair::default();
+
+        // test for unspported domain
```
davidge20: please state the unsupported domains explicitly

also, lets correct the typo to "unsupported" throughout

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+            -(Errno::EOPNOTSUPP as i32)
+        );
+
+        // test for unspported socktype
+        assert_eq!(
+            Cage::socketpair_syscall(cage.clone(), AF_UNIX, SOCK_DGRAM, 0, &mut socketpair),
+            -(Errno::EOPNOTSUPP as i32)
+        );
+
+        // test for unspported protocol
```
davidge20: what is the unsupported protocol of 1?

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+        lindrustinit(0);
+        let cage = interface::cagetable_getref(1);
+        let mut socketpair = interface::SockPair::default();
+        // try with cloexec flag
+        assert_eq!(
+            Cage::socketpair_syscall(cage.clone(), AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, &mut socketpair),
+            0
+        );
+
+        let mut uselessstatdata = StatData::default();
```
davidge20, line 1699: please explain the choice for StatData

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+
+        // check if the file descriptor exists
```
davidge20, line 1701: so the file descriptor exists if any sort of data is returned from fstat?

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+            Cage::socketpair_syscall(cage.clone(), AF_UNIX, 472810394, 0, &mut socketpair),
+            -(Errno::EOPNOTSUPP as i32)
+        );
+
+        assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
+    pub fn ut_lind_net_socketpair_cloexec() {
+        // this test is used for testing socketpair when cloexec flag is set
```
davidge20, line 1677: please mention the expected behavior of the test

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+            0
+        );
+
+        let mut uselessstatdata = StatData::default();
+
+        // check if the file descriptor exists
+        assert_eq!(cage.fstat_syscall(socketpair.sock1, &mut uselessstatdata), -(Errno::EOPNOTSUPP as i32));
+        assert_eq!(cage.fstat_syscall(socketpair.sock2, &mut uselessstatdata), -(Errno::EOPNOTSUPP as i32));
+
+        assert_eq!(cage.exec_syscall(2), 0);
```
davidge20, line 1713: what is being executed here?

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+            newcage.fstat_syscall(socketpair.sock2, &mut uselessstatdata),
+            -(Errno::EBADF as i32)
+        );
+
+        assert_eq!(newcage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
+    pub fn ut_lind_net_socketpair_nonblocking() {
+        // this test is used for testing socketpair when nonblocking flag is set
```
davidge20, line 1734: please add an explanation of what the test is doing at a high level and the expected behavior upon success/failure as well

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+        assert_eq!(
+            Cage::socketpair_syscall(cage.clone(), AF_UNIX, SOCK_STREAM | SOCK_NONBLOCK, 0, &mut socketpair),
+            0
+        );
+
+        let cage2 = cage.clone();
+
+        let thread = interface::helper_thread(move || {
+            let mut buf = sizecbuf(10);
+            let mut counter = 0;
```
davidge20, line 1763: why is the counter necessary and why are we incrementing it?

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+
+            // receive the message
+            // since peer will sleep for 30ms before send the message
+            // some nonblocing returns are expected
```
davidge20: nonblocking*

and what is a nonblocking return?

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+            loop {
+                let result = cage2.recv_syscall(socketpair.sock2, buf.as_mut_ptr(), 10, 0);
+                if result == -(Errno::EAGAIN as i32) {
```
davidge20, line 1770: please add when EAGAIN occurs

src/tests/networking_tests.rs
```diff
@@ -1544,6 +1556,167 @@ pub mod net_tests {
...
+        thread.join().unwrap();
+        thread_2.join().unwrap();
+
+        assert_eq!(cage.close_syscall(socketpair.sock1), 0);
+        assert_eq!(cage.close_syscall(socketpair.sock2), 0);
+
+        assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
```
davidge20, line 1837: I learned a lot from reading through your tests! Great work overall!

- Just a note, I would appreciate more comments especially at the top of blocks of code that perform something. This would help me out a lot.

davidge20 (commented):
I learned a lot from reading through your tests! Great work overall!

Just a note, I would appreciate more comments especially at the top of blocks of code that perform something. This would help me out a lot.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3050,6 +3050,48 @@ impl Cage {
...
+    /// * `protocol` - The protocol specifies a particular protocol to be used with the
+    ///                socket. Currently only support the default protocol (IPPROTO_TCP).
+    /// * `sv` -  The file descriptors used in referencing the new sockets are returned
+    ///           in sv.sock1 and sv.sock2. The two sockets are indistinguishable.
+
+    /// ### Returns
+    /// On success, zero is returned. Otherwise, errors or panics are returned for
+    /// different scenarios.
+    ///
+    /// ### Errors and Panics
```
yashaswi2000: seperate errors and panics

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3077,6 +3125,7 @@ impl Cage {
             );
         }
 
+        // get the flags
```
yashaswi2000, line 3176: can we add error checking here, such that we error out for invalid combination of flags, something like this: 

`       if socktype & !(SOCK_NONBLOCK | SOCK_CLOEXEC | 0x7) != 0 {
            return syscall_error(
                Errno::EOPNOTSUPP,
                "socket",
                "Invalid combination of flags"
            );
        }`

src/tests/mod.rs
```diff
@@ -86,6 +88,17 @@ mod setup {
         thelock
     }
 
+    fn set_panic_hook() {
```
yashaswi2000, line 93: very interesting!. thanks for this!

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3080,22 +3080,68 @@ impl Cage {
...
+    /// * `protocol` - The protocol specifies a particular protocol to be used with the
+    ///                socket. Currently only support the default protocol (IPPROTO_TCP).
+    /// * `sv` -  The file descriptors used in referencing the new sockets are returned
+    ///           in sv.sock1 and sv.sock2. The two sockets are indistinguishable.
+
+    /// ### Returns
+    /// On success, zero is returned. Otherwise, errors or panics are returned for
+    /// different scenarios.
+    ///
+    /// ### Errors
```
yashaswi2000, line 3122: add a panic section after this, mentioning when can this function panic

yashaswi2000 (approved):
LGTM

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3080,22 +3080,68 @@ impl Cage {
...
+    /// ## ------------------SOCKETPAIR SYSCALL------------------
+    /// ### Description
+    /// The `socketpair_syscall()` call creates an unnamed pair of connected sockets
+    /// in the specified domain, of the specified type, and using the optionally
+    /// specified protocol.
+    /// The file descriptors used in referencing the new sockets are returned
+    /// in sv.sock1 and sv.sock2. The two sockets are indistinguishable.
+    /// ### Function Arguments
+    /// The `socketpair_syscall()` receives four arguments:
+    /// * `domain` -  The domain argument specifies a communication domain; this
```
rennergade, line 3092: I'm not sure if this is wrapping more than 80 chars. Regardless make sure to cargo fmt this before we merge.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3080,22 +3080,68 @@ impl Cage {
...
         sv: &mut interface::SockPair,
     ) -> i32 {
-        let newprotocol = if protocol == 0 { IPPROTO_TCP } else { protocol };
+        let newprotocol = if protocol == 0 { IPPROTO_TCP } else { protocol }; // only support protocol of 0 currently
+                                                                              // BUG: current implementation of socketpair creates two sockets and bind to an unique address.
+                                                                              // But according to standard, the sockets created from socketpair should not bind to any address
+
         // firstly check the parameters
+        // socketpair should always be a AF_UNIX TCP socket
+        // socket type is stored at the lowerest 3 bits
```
rennergade: "lowest"

rennergade: the type comment should go with the corresponding block below

rennergade (changes requested):
Some minor changes requested from Yash and I. Once those are addressed I think this is good to go!

src/safeposix/syscalls/net_calls.rs
qianxichen233: I did not find any place that might cause panic currently. But I just wonder is the panic something that should occur inside the syscalls? I just feel like whenever something goes wrong, we should always return a syscall error instead of letting it panic directly. A panic inside the syscall would look like a bug to me.

yashaswi2000: hmm that's something to think about. ideally I agree with your line of thought, but the purpose of panic is kind of to say something went wrong when it shouldn't. And it might not be related to the logic of the Syscall.  @rennergade your thoughts on this?

rennergade (approved):
Great! Approved.

Issue #246, opened 2024-05-16
Fix Rust Test Suite Bind Error


rennergade:
The RustPOSIX test suite requires you to wait between test runs because of trying to repeatedly bind to the same ports. We should be able to fix this by randomizing ports between runs, as well as handling bind errors gracefully. Let's implement this to make the tests easier to run.

rennergade:
This got fixed with Rupeshs recent PR. Thanks @rupeshkoushik07 !

PR #258, opened 2024-06-07
Random port

rupeshkoushik07:
## Description

Fixes # [246](https://github.com/Lind-Project/safeposix-rust/issues/246#event-12841048134)

### Type of change
Added random ports.

- [x] New feature (adds random ports)

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -65,7 +65,7 @@ mod main_tests {
         lindrustfinalize();
 
         println!("FS TESTS");
-        test_fs();
+        // test_fs();
 
         println!("NET TESTS");
         net_tests();
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -3,6 +3,7 @@
 mod fs_tests;
 mod ipc_tests;
 mod networking_tests;
+use rand::Rng;
 
 use crate::interface;
 use crate::safeposix::{cage::*, filesystem::*};
@@ -89,4 +90,10 @@ pub fn sizecbuf<'a>(size: usize) -> Box<[u8]> {
 
 pub fn cbuf2str(buf: &[u8]) -> &str {
     std::str::from_utf8(buf).unwrap()
+}
+
+pub fn generate_random_port() -> u16 {
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    rng.gen_range(49152..65535)
 }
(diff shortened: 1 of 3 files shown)


JustinCappos:
FYI: perhaps mark as draft until it's ready...

src/tests/mod.rs
```diff
@@ -90,3 +91,9 @@ pub fn sizecbuf<'a>(size: usize) -> Box<[u8]> {
 pub fn cbuf2str(buf: &[u8]) -> &str {
     std::str::from_utf8(buf).unwrap()
 }
+
+pub fn generate_random_port() -> u16 {
```
rennergade, line 103: add a comment here quickly explaining why were randomizing ports and in what situations we do so

src/tests/networking_tests.rs
```diff
@@ -1083,15 +1084,16 @@ pub mod net_tests {
         assert_eq!(cage.listen_syscall(serversockfd, 4), 0);
 
         // allocate spaces for fd_set bitmaps
-        let inputs = &mut interface::FdSet::new();
+        let master_set = &mut interface::FdSet::new();
```
rennergade, line 1089: this looks good and is a very helpful change. can you leave a comment explaining how the sets work?

rennergade (changes requested):
Looks great. Needs two more comments I think.

src/tests/mod.rs
JustinCappos: This is useful, but doesn't guarantee these are unused.  There are ways to do this: https://superuser.com/questions/885414/linux-command-get-unused-port

I think random is better than fixed, but checking for a port to be unused may be better than either...

JustinCappos: To expand on this, you have a ~2.5% chance of a collision now (just from your own tests), and if we get to 50 tests, it'll be ~7.5%.  This is enough that we will frequently see errors when testing and so will be conditioned that false alarms are expected.  I'd prefer to sidestep this, if possible, especially if we're doing automated testing on every commit...

rupeshkoushik07: I initially underestimated the likelihood of collisions. However, your calculations demonstrate the necessity of checking for unused ports. I have updated the function accordingly.

JustinCappos (approved):
thanks for the fixes.  Looks good!

PR #259, opened 2024-06-07
Fcntl_syscall updates

ve1nard:
## Description

Fixes # (issue)

The following changes include more elaborate comments and new unit tests for fcntl_syscall. Moreover, it was proposed to use "if let" construct instead of "unwrap" to get an entry from a file descriptor table to avoid "panic!" when provided with a file descriptor outside of the allowed range.

### Type of change

- [ ] Small fix for obtaining a file descriptor table entry without "panic!"
- [ ] More detailed comments for fcntl_syscall
- [ ] New unit tests for fcntl_syscall

## How Has This Been Tested?
To run the tests, we need to run cargo test --lib command inside the safeposix-rust directory.

All the tests are present under this directory: lind_project/src/safeposix-rust/src/tests/fs_tests.rs

- Test A - `ut_lind_fs_fcntl_valid_args()`
- Test B - `ut_lind_fs_fcntl_invalid_args()`
- Test C - `ut_lind_fs_fcntl_invalid_fd()`
- Test D - `ut_lind_fs_fcntl_dup()`

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1901,101 +1901,115 @@ impl Cage {
 
     pub fn fcntl_syscall(&self, fd: i32, cmd: i32, arg: i32) -> i32 {
         //if the provided file descriptor is out of bounds, get_filedescriptor returns Err(),
-        //and using unwrap on Err() causes a thread to panick
-        //instead, I propose using the 'if let' construct to be able to report an error to 
-        //the user instead of simply panicking
+        //matching on which throws a 'Bad file number' error
         //otherwise, file descriptor table entry is stored in 'checkedfd'
-        if let Ok(checkedfd) = self.get_filedescriptor(fd) {        
-            let mut unlocked_fd = checkedfd.write();
-            //performing the specified command if the file descriptor entry is not empty
-            if let Some(filedesc_enum) = &mut *unlocked_fd {
-                //'flags' consists of bitwise-or'd access mode, file creation, and file status flags
-                //to retrieve a particular flag, it can bitwise-and'd with 'flags'
-                let flags = match filedesc_enum {
-                    Epoll(obj) => &mut obj.flags,
-                    Pipe(obj) => &mut obj.flags,
-                    Stream(obj) => &mut obj.flags,
-                    File(obj) => &mut obj.flags,
-                    //not clear why running F_SETFL on Socket type requires special treatment
-                    Socket(ref mut sockfdobj) => {
-                        if cmd == F_SETFL && arg >= 0 {
-                            let sock_tmp = sockfdobj.handle.clone();
-                            let mut sockhandle = sock_tmp.write();
+        let fd_table_entry_ptr = self.get_filedescriptor(fd);
+        match fd_table_entry_ptr {
+            Err(()) => {
+                syscall_error(Errno::EBADF, "fcntl", "File descriptor is out of range")
+            },
+            Ok(checkedfd) => {
+                let mut unlocked_fd = checkedfd.write();
+                //returning a 'Bad file number' error if the file descriptor entry is empty
+                //performing the specified command otherwise
+                let fd_table_entry = &mut *unlocked_fd;
+                match fd_table_entry {
+                    None => syscall_error(Errno::EBADF, "fcntl", "Invalid file descriptor"),
+                    Some(filedesc_enum) => {
+                        //'flags' consists of bitwise-or'd access mode, file creation, and file status flags
+                        //to retrieve a particular flag, it can bitwise-and'd with 'flags'
+                        let flags = match filedesc_enum {
+                            Epoll(obj) => &mut obj.flags,
+                            Pipe(obj) => &mut obj.flags,
+                            Stream(obj) => &mut obj.flags,
+                            File(obj) => &mut obj.flags,
+                            //not clear why running F_SETFL on Socket type requires special treatment
+                            Socket(ref mut sockfdobj) => {
+                                if cmd == F_SETFL && arg >= 0 {
+                                    let sock_tmp = sockfdobj.handle.clone();
+                                    let mut sockhandle = sock_tmp.write();
 
-                            if let Some(ins) = &mut sockhandle.innersocket {
-                                let fcntlret;
-                                if arg & O_NONBLOCK == O_NONBLOCK {
-                                    //set for non-blocking I/O
-                                    fcntlret = ins.set_nonblocking();
-                                } else {
-                                    //clear non-blocking I/O
-                                    fcntlret = ins.set_blocking();
-                                }
-                                if fcntlret < 0 {
-                                    match Errno::from_discriminant(interface::get_errno()) {
-                                        Ok(i) => {
-                                            return syscall_error(
-                                                i,
-                                                "fcntl",
-                                                "The libc call to fcntl failed!",
-                                            );
+                                    if let Some(ins) = &mut sockhandle.innersocket {
+                                        let fcntlret;
+                                        if arg & O_NONBLOCK == O_NONBLOCK {
+                                            //set non-blocking I/O
+                                            fcntlret = ins.set_nonblocking();
+                                        } else {
+                                            //set blocking I/O
+                                            fcntlret = ins.set_blocking();
                                         }
-                                        Err(()) => panic!("Unknown errno value from fcntl returned!"),
-                                    };
+                                        if fcntlret < 0 {
+                                            match Errno::from_discriminant(interface::get_errno()) {
+                                                Ok(i) => {
+                                                    return syscall_error(
+                                                        i,
+                                                        "fcntl",
+                                                        "The libc call to fcntl failed!",
+                                                    );
+                                                }
+                                                Err(()) => panic!("Unknown errno value from fcntl returned!"),
+                                            };
+                                        }
+                                    }
                                 }
+
+                                &mut sockfdobj.flags
                             }
-                        }
+                        };
 
-                        &mut sockfdobj.flags
-                    }
-                };
-
-                //matching the tuple
-                match (cmd, arg) {
-                    //because the arg parameter is not used in certain commands, it can be anything (..)
-                    //currently, O_CLOEXEC is the only defined file descriptor flag, thus only this flag is
-                    //masked when using F_GETFD or F_SETFD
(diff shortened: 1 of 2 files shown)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1900,85 +1900,102 @@ impl Cage {
...
-                                        return syscall_error(
-                                            i,
-                                            "fcntl",
-                                            "The libc call to fcntl failed!",
-                                        );
-                                    }
-                                    Err(()) => panic!("Unknown errno value from fcntl returned!"),
-                                };
+        //if the provided file descriptor is out of bounds, get_filedescriptor returns Err(),
+        //and using unwrap on Err() causes a thread to panick
```
JustinCappos:
```suggestion
        //and using unwrap on Err() causes a thread to panic
```

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1900,85 +1900,102 @@ impl Cage {
...
+        //instead, I propose using the 'if let' construct to be able to report an error to 
+        //the user instead of simply panicking
+        //otherwise, file descriptor table entry is stored in 'checkedfd'
+        if let Ok(checkedfd) = self.get_filedescriptor(fd) {
```
JustinCappos: So I like the philosophy here, but I think you want to reverse the if statement behavior so that you error out quickly, rather than continually nesting your if statement.  This way one can much more easily read the code.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1900,85 +1900,102 @@ impl Cage {
...
-                        *flags |= O_CLOEXEC;
-                    } else {
-                        *flags &= !O_CLOEXEC;
+                //matching the tuple
+                match (cmd, arg) {
+                    //because the arg parameter is not used in certain commands, it can be anything (..)
+                    //currently, O_CLOEXEC is the only defined file descriptor flag, thus only this flag is
+                    //masked when using F_GETFD or F_SETFD
+                    (F_GETFD, ..) => *flags & O_CLOEXEC,
+                    // set the flags but make sure that the flags are valid
```
JustinCappos: I'm not sure how this achieves your goal.  Can you say more?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1900,85 +1900,102 @@ impl Cage {
...
+                        *flags |= arg;
+                        0
+                    }
+                    (F_DUPFD, arg) if arg >= 0 => self._dup2_helper(&filedesc_enum, arg, false),
+                    //TO DO: implement. this one is saying get the signals
+                    (F_GETOWN, ..) => {
+                        0 //TO DO: traditional SIGIO behavior
+                    }
+                    (F_SETOWN, arg) if arg >= 0 => {
+                        0 //this would return the PID if positive and the process group if negative,
```
JustinCappos: Would it set this or return it?  Why do we do nothing?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1900,85 +1900,102 @@ impl Cage {
...
+                    (F_GETFL, ..) => {
+                        //for get, we just need to return the flags
+                        *flags & !O_CLOEXEC
+                    }
+                    (F_SETFL, arg) if arg >= 0 => {
+                        *flags |= arg;
+                        0
+                    }
+                    (F_DUPFD, arg) if arg >= 0 => self._dup2_helper(&filedesc_enum, arg, false),
+                    //TO DO: implement. this one is saying get the signals
```
JustinCappos: Which is saying to get the signals?  What signals?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1900,85 +1900,102 @@ impl Cage {
...
+                        0 //TO DO: traditional SIGIO behavior
+                    }
+                    (F_SETOWN, arg) if arg >= 0 => {
+                        0 //this would return the PID if positive and the process group if negative,
+                        //either way do nothing and return success
+                    }
+                    _ => syscall_error(
+                        Errno::EINVAL,
+                        "fcntl",
+                        "Arguments provided do not match implemented parameters",
```
JustinCappos: Can we list anything else useful like the arguments themselves?

src/tests/fs_tests.rs
```diff
@@ -437,35 +440,104 @@ pub mod fs_tests {
...
-
-        //checking if the flags are updated...
+        //changing the file access mode to read-only, enabling the 
+        //O_NONBLOCK file status flag, and checking if they were correctly set
+        assert_eq!(cage.fcntl_syscall(filefd, F_SETFL, O_RDONLY | O_NONBLOCK), 0);
         assert_eq!(cage.fcntl_syscall(filefd, F_GETFL, 0), 2048);
 
+        //when provided with 'F_GETFD' or 'F_GETFL' command, 'arg' should be ignored, thus even
+        //negative arg values should produce nomal behavior
+        //However, testing results in two errors
```
JustinCappos: Okay, great!  So this test will fail now, but after we fix the code later it will pass, right?

JustinCappos (changes requested):
The testing is good.  I have a major style comment in the code itself, but much of it looks sound.

src/safeposix/syscalls/fs_calls.rs
ve1nard: If I write something like "if let Err() = self.get_filedescriptor(fd)", then in the "else", I will have to do pattern matching once again to extract checkedfd, which seems redundant. Should I use match instead?

src/safeposix/syscalls/fs_calls.rs
ve1nard: F_GETFD command should return file descriptor flags only, which means that  the returned value should not include access mode flags and file status flags, and because O_CLOEXEC is the only implemented file descriptor as of now, we can get it by bit-wise and'ing. Same logic applies to F_SETFD: because O_CLOEXEC is the only existing file descriptor flag, we need to check only it.

src/safeposix/syscalls/fs_calls.rs
ve1nard: When I was talking to Yash, he suggested me that I should not focus on things that are not yet implemented, and that is why I left this part with F_SETOWN and F_GETOWN commands untouched. Should I work on them?

src/tests/fs_tests.rs
ve1nard: This is my bad: everything works now and the test passes. I just forgot to delete this part of the comment.

src/safeposix/syscalls/fs_calls.rs
ve1nard: Thank you! I will include a (cmd, arg) tuple in the error message.

src/safeposix/syscalls/fs_calls.rs
JustinCappos: That would be an option.  

I'm not a Rust expert, but code like [this](https://github.com/JustinCappos/fdtables/blob/main/src/dashmaparrayglobal.rs#L812C14-L882C1) looks more "correct" to me.  I'm very open to hearing from others, but certainly in Python, C, etc. it is preferred to do it in the style I've written.

yashaswi2000:
Please add documentation about the syscall, its arguments, return values and manpage link above the function definition.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1898,87 +1898,126 @@ impl Cage {
     }
 
     //------------------------------------FCNTL SYSCALL------------------------------------
+    
+    //fcntl performs operations, like returning or setting file status flags,
```
JustinCappos, line 1902: Looks great!

@yashaswi2000 , do you want to have the rustdoc changes be done in these edits or a different set of PRs?  I do think that for new PRs, we likely should include them...

JustinCappos (approved):

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1898,87 +1898,126 @@ impl Cage {
...
+                            File(obj) => &mut obj.flags,
+                            //not clear why running F_SETFL on Socket type requires special treatment
+                            Socket(ref mut sockfdobj) => {
+                                if cmd == F_SETFL && arg >= 0 {
+                                    let sock_tmp = sockfdobj.handle.clone();
+                                    let mut sockhandle = sock_tmp.write();
+
+                                    if let Some(ins) = &mut sockhandle.innersocket {
+                                        let fcntlret;
+                                        if arg & O_NONBLOCK == O_NONBLOCK {
```
JustinCappos: Deep nesting.  Is there a way we can avoid this?

rennergade: No not easily. The rustfmt makes it looks worse than it is.

src/safeposix/syscalls/fs_calls.rs
yashaswi2000: I will refactor these first 3 PRs. and any new ones coming in will follow the template I will share with them.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1898,87 +1898,126 @@ impl Cage {
...
-                                };
-                            }
-                        }
-                    }
+                                }
 
-                    &mut sockfdobj.flags
-                }
-            };
+                                &mut sockfdobj.flags
```
rennergade: remove extra whitespace

src/safeposix/syscalls/fs_calls.rs
JustinCappos: I think clippy will suggest a way to change this pattern:
```
       match fd_table_entry_ptr {
            Err(()) => {
                syscall_error(Errno::EBADF, "fcntl", "File descriptor is out of range")
            },
            Ok(checkedfd) => {
                let mut unlocked_fd = checkedfd.write();
                ...
```
into an `if let Ok(checkedfd) = fd_table_entry_ptr {...}` in at least some cases.  Here, it would unfortunately move the error condition to the end, so while this would reduce indentation, it isn't a 100% win...

rennergade: This seems fine as it for now but may be part of a later refactor.

yashaswi2000 (approved):
LGTM!

rennergade (approved):
Approved!

PR #262, opened 2024-06-13
Bind comment updates

davidge20:
## Description

Commented out bind_syscall and all functions related to it

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -12,6 +12,7 @@ use crate::safeposix::filesystem::*;
 use crate::safeposix::net::*;
 
 impl Cage {
+    //Initializes a socket description and sets the necessary flags
     fn _socket_initializer(
         &self,
         domain: i32,
@@ -21,6 +22,10 @@ impl Cage {
         cloexec: bool,
         conn: ConnState,
     ) -> SocketDesc {
+        //For blocking sockets, operation wait until completed.
+        //For non-blocking sockets, operations return immediately, even if the requested operation 
+        //is not completed.
+        //cloexec flag closes the fd pointing to the socket on the execution of a new program
         let flags = if nonblocking { O_NONBLOCK } else { 0 } | if cloexec { O_CLOEXEC } else { 0 };
 
         let sockfd = SocketDesc {
@@ -36,29 +41,45 @@ impl Cage {
         return sockfd;
     }
 
+    //Find an available file descriptor index in the File Descriptor Table
+    //If available, insert the socket fd into the File Descriptor Table at the 
+    //available index
     fn _socket_inserter(&self, sockfd: FileDescriptor) -> i32 {
         let (fd, guardopt) = self.get_next_fd(None);
+        //Indicate an error in the fd and propogate the error forward 
         if fd < 0 {
             return fd;
         }
         let fdoption = &mut *guardopt.unwrap();
-        let _insertval = fdoption.insert(sockfd);
+        //Insert the sockfd as the FileDescriptor inside fdoption
+        let _insertval = fdoption.insert(sockfd); 
         return fd;
     }
 
+    //An implicit bind refers to the automatic binding of a socket to an address 
+    //and port by the system, without an explicit call to the bind() function by the
+    //programmer. This typically happens when the socket is used for client-side 
+    //operations, such as when it initiates a connection to a server.
     fn _implicit_bind(&self, sockhandle: &mut SocketHandle, domain: i32) -> i32 {
         if sockhandle.localaddr.is_none() {
+            //Assign a new local address to the socket handle
             let localaddr = match Self::assign_new_addr(
                 sockhandle,
                 domain,
+                //The SO_RESUEPORT bit within the protocol encodes rebind ability
+                //The rebind ability of a socket refers to whether a socket can be 
+                //re-bound to an address and port that it was previously bound to, 
+                //especially after it has been closed or if the binding has been reset. 
                 sockhandle.protocol & (1 << SO_REUSEPORT) != 0,
             ) {
                 Ok(a) => a,
                 Err(e) => return e,
             };
 
+            //Bind the address to the socket handle
             let bindret = self.bind_inner_socket(sockhandle, &localaddr, true);
 
+            //If an error occurs during binding,
             if bindret < 0 {
                 match Errno::from_discriminant(interface::get_errno()) {
                     Ok(i) => {
@@ -159,35 +180,49 @@ impl Cage {
 
     //creates a sockhandle if none exists, otherwise this is a no-op
     pub fn force_innersocket(sockhandle: &mut SocketHandle) {
+        //If innersocket is not available, then insert a socket into innersocket
         if let None = sockhandle.innersocket {
+            //Create a new socket
             let thissock =
                 interface::Socket::new(sockhandle.domain, sockhandle.socktype, sockhandle.protocol);
 
+            //Loop through socket options and check which ones are set
             for reuse in [SO_REUSEPORT, SO_REUSEADDR] {
+                //If socket option is not set, continue to next socket option
                 if sockhandle.socket_options & (1 << reuse) == 0 {
                     continue;
                 }
+
+                //Otherwise, set the socket option in the new socket
                 let sockret = thissock.setsockopt(SOL_SOCKET, reuse, 1);
+                //Failure occured upon creation of socket
                 if sockret < 0 {
                     panic!("Cannot handle failure in setsockopt on socket creation");
                 }
             }
 
+            //Insert a socket into the innersocket value
             sockhandle.innersocket = Some(thissock);
         };
     }
 
     //we assume we've converted into a RustSockAddr in the dispatcher
+    //bind a name to a socket - assign the address specified by localaddr to the 
+    //socket referred to by the file descriptor fd
     pub fn bind_syscall(&self, fd: i32, localaddr: &interface::GenSockaddr) -> i32 {
         self.bind_inner(fd, localaddr, false)
     }
 
+    //Based on the domain of the socket, bind the socket with a local address and 
+    //insert a cloned local address in the socket handle
+    //return the result of the binding operation
     fn bind_inner_socket(
         &self,
         sockhandle: &mut SocketHandle,
         localaddr: &interface::GenSockaddr,
         prereserved: bool,
     ) -> i32 {
+        //The family of the local address must match the domain of the socket handle
         if localaddr.get_family() != sockhandle.domain as u16 {
             return syscall_error(
                 Errno::EINVAL,
@@ -196,6 +231,7 @@ impl Cage {
             );
         }
 
+        //If the socket is already binded to an address, exit with error
         if sockhandle.localaddr.is_some() {
             return syscall_error(
                 Errno::EINVAL,
@@ -206,6 +242,7 @@ impl Cage {
 
         let mut newsockaddr = localaddr.clone();
 
+        //Bind socket based on domain type 
         let res = match sockhandle.domain {
             AF_UNIX => self.bind_inner_socket_unix(sockhandle, &mut newsockaddr),
             AF_INET | AF_INET6 => {
@@ -221,6 +258,7 @@ impl Cage {
         res
     }
 
+    //bind_syscall implementation in the case that the socket's domain is unix
     fn bind_inner_socket_unix(
         &self,
         sockhandle: &mut SocketHandle,
@@ -232,8 +270,11 @@ impl Cage {
         if path.len() == 0 {
             return syscall_error(Errno::ENOENT, "bind", "given path was null");
         }
+        //true path is normalized path of the path to a unix socket
         let truepath = normpath(convpath(path), self);
 
+        //returns tuple consisting of inode number of file (if it exists), and 
+        //inode number of parent (if it exists)
         match metawalkandparent(truepath.as_path()) {
             //If neither the file nor parent exists
             (None, None) => {
@@ -244,18 +285,25 @@ impl Cage {
                 let filename = truepath.file_name().unwrap().to_str().unwrap().to_string(); //for now we assume this is sane, but maybe this should be checked later
 
                 //this may end up skipping an inode number in the case of ENOTDIR, but that's not catastrophic
+                //FS_METADATA contains information about the file system
                 let newinodenum = FS_METADATA
                     .nextinode
-                    .fetch_add(1, interface::RustAtomicOrdering::Relaxed); //fetch_add returns the previous value, which is the inode number we want
+                    .fetch_add(1, interface::RustAtomicOrdering::Relaxed); 
+                    //fetch_add returns the previous value, which is the inode number we want,
+                    //while incrementing the nextinode value as well
                 let newinode;
 
+                //Pattern match the Directory Inode of the parent
                 if let Inode::Dir(ref mut dir) =
                     *(FS_METADATA.inodetable.get_mut(&pardirinode).unwrap())
                 {
+                    //Add file type flags and user read,write,execute permission flags ***
                     let mode = (dir.mode | S_FILETYPEFLAGS as u32) & S_IRWXA;
+                    //Add file type constant of a socket
                     let effective_mode = S_IFSOCK as u32 | mode;
 
                     let time = interface::timestamp(); //We do a real timestamp now
+                    //Create a new inode for the file of the socket
                     newinode = Inode::Socket(SocketInode {
                         size: 0,
                         uid: DEFAULT_UID,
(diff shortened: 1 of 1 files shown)


src/safeposix/syscalls/net_calls.rs
```diff
@@ -21,6 +22,10 @@ impl Cage {
         cloexec: bool,
         conn: ConnState,
     ) -> SocketDesc {
+        //For blocking sockets, operation wait until completed.
+        //For non-blocking sockets, operations return immediately, even if the requested operation 
+        //is not completed.
+        //cloexec flag closes the fd pointing to the socket on the execution of a new program
```
JustinCappos: Why do we need to track these flags?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -36,29 +41,45 @@ impl Cage {
         return sockfd;
     }
 
+    //Find an available file descriptor index in the File Descriptor Table
+    //If available, insert the socket fd into the File Descriptor Table at the 
+    //available index
     fn _socket_inserter(&self, sockfd: FileDescriptor) -> i32 {
         let (fd, guardopt) = self.get_next_fd(None);
+        //Indicate an error in the fd and propogate the error forward
```
JustinCappos: When would this occur?  Why?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -36,29 +41,45 @@ impl Cage {
...
+    //available index
     fn _socket_inserter(&self, sockfd: FileDescriptor) -> i32 {
         let (fd, guardopt) = self.get_next_fd(None);
+        //Indicate an error in the fd and propogate the error forward 
         if fd < 0 {
             return fd;
         }
         let fdoption = &mut *guardopt.unwrap();
-        let _insertval = fdoption.insert(sockfd);
+        //Insert the sockfd as the FileDescriptor inside fdoption
```
JustinCappos: Why?  What are we doing here?  

(We can assume the reader knows Rust well enough to know they are inserting  a sockfd inside the fdoption datastructure)

src/safeposix/syscalls/net_calls.rs
```diff
@@ -36,29 +41,45 @@ impl Cage {
...
     fn _implicit_bind(&self, sockhandle: &mut SocketHandle, domain: i32) -> i32 {
         if sockhandle.localaddr.is_none() {
+            //Assign a new local address to the socket handle
             let localaddr = match Self::assign_new_addr(
                 sockhandle,
                 domain,
+                //The SO_RESUEPORT bit within the protocol encodes rebind ability
+                //The rebind ability of a socket refers to whether a socket can be 
+                //re-bound to an address and port that it was previously bound to, 
+                //especially after it has been closed or if the binding has been reset.
```
JustinCappos, line 76: Okay, why do we set this?  

It's slightly odd that SO_REUSEPORT is used to leftshift a bit.  Why is this done?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +180,49 @@ impl Cage {
 
     //creates a sockhandle if none exists, otherwise this is a no-op
     pub fn force_innersocket(sockhandle: &mut SocketHandle) {
+        //If innersocket is not available, then insert a socket into innersocket
```
JustinCappos: What is an innersocket?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +180,49 @@ impl Cage {
...
         if let None = sockhandle.innersocket {
+            //Create a new socket
```
JustinCappos: I can see we're creating a socket, but why?

Also, can this operation fail?   What happens if it does?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +180,49 @@ impl Cage {
...
             let thissock =
                 interface::Socket::new(sockhandle.domain, sockhandle.socktype, sockhandle.protocol);
 
+            //Loop through socket options and check which ones are set
```
JustinCappos, line 201: This is a weird little bit of code.  I think you want to explain that because you need to make separate calls to set different socket options (i.e. you can't set multiple flags at once) you loop here.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -196,6 +231,7 @@ impl Cage {
             );
         }
 
+        //If the socket is already binded to an address, exit with error
```
JustinCappos:
```suggestion
        //If the socket is already bound to an address, exit with error
```

src/safeposix/syscalls/net_calls.rs
```diff
@@ -206,6 +242,7 @@ impl Cage {
 
         let mut newsockaddr = localaddr.clone();
 
+        //Bind socket based on domain type
```
JustinCappos, line 320: Why do you need to do this check?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -244,18 +285,25 @@ impl Cage {
                 let filename = truepath.file_name().unwrap().to_str().unwrap().to_string(); //for now we assume this is sane, but maybe this should be checked later
 
                 //this may end up skipping an inode number in the case of ENOTDIR, but that's not catastrophic
+                //FS_METADATA contains information about the file system
                 let newinodenum = FS_METADATA
                     .nextinode
-                    .fetch_add(1, interface::RustAtomicOrdering::Relaxed); //fetch_add returns the previous value, which is the inode number we want
+                    .fetch_add(1, interface::RustAtomicOrdering::Relaxed);
```
JustinCappos, line 369: Interesting.  What does Relaxed do here?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -268,26 +316,35 @@ impl Cage {
                         mtime: time,
                     });
 
+                    //Insert new inode number of file to the same file system
```
JustinCappos: I can't understand what you mean here.  Can you explain in more detail?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -268,26 +316,35 @@ impl Cage {
...
                     dir.linkcount += 1;
                 } else {
+                    //Parent dictory inode does not exist in inode table of file system
                     return syscall_error(
                         Errno::ENOTDIR,
                         "bind",
                         "unix domain socket path made socket address child of non-directory file",
                     );
                 }
+                //Insert unix info into socket handle
```
JustinCappos, line 413: Why?  What is this doing?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -268,26 +316,35 @@ impl Cage {
...
                     inode: newinodenum,
                 });
 
+                //Insert path to socket file into a set
                 NET_METADATA.domsock_paths.insert(truepath);
+                //Insert the file inode num and inode as key-value pair into
+                //file system inode table
                 FS_METADATA.inodetable.insert(newinodenum, newinode);
             }
+            //File already exists
```
JustinCappos: What is this doing?!?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -296,35 +353,47 @@ impl Cage {
...
         } else {
+            //Reserve a new local port num
             let localout = NET_METADATA._reserve_localport(
                 newsockaddr.addr(),
                 newsockaddr.port(),
                 sockhandle.protocol,
                 sockhandle.domain,
                 intent_to_rebind,
             );
+            //If reserving a local port causes an error, return the error num
```
JustinCappos: You could possibly do the next few lines as a match statement instead.  It would make it clearer that your last unwrap is safe...

I'll note, it is slightly weird that the code here really just wants the inner result regardless of if it is an err or not...

src/safeposix/syscalls/net_calls.rs
```diff
@@ -337,27 +406,40 @@ impl Cage {
...
+    //localaddr: reference to the GenSockaddr enum that hold the address
+    //prereserved: bool that describes whether the address and port have 
+    //             been set aside or designated for specific purposes
     pub fn bind_inner(
         &self,
         fd: i32,
         localaddr: &interface::GenSockaddr,
         prereserved: bool,
     ) -> i32 {
+        //chekedfd is an atomic reference count of a lock on the fd
```
JustinCappos:
```suggestion
        //checkedfd is an atomic reference count of a lock on the fd
```

Of a lock?  Or of the number of locks?  Or what?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -337,27 +406,40 @@ impl Cage {
...
         let checkedfd = self.get_filedescriptor(fd).unwrap();
-        let mut unlocked_fd = checkedfd.write();
+        //returns a write lock once no other writers or readers have access to the lock
+        let mut unlocked_fd = checkedfd.write(); //why isn't .unwrap() used here?
```
JustinCappos: Great idea to add a comment saying what you don't know!!!  😄

src/safeposix/syscalls/net_calls.rs
```diff
@@ -337,27 +406,40 @@ impl Cage {
...
         if let Some(filedesc_enum) = &mut *unlocked_fd {
             match filedesc_enum {
+                //*unlocked_fd is in the format of Option<T>, where T is of type Socket(&mut SocketDesc)
                 Socket(ref mut sockfdobj) => {
+                    //Clone the socket handle
```
JustinCappos, line 515: Why do we clone this?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -337,27 +406,40 @@ impl Cage {
...
                     let sock_tmp = sockfdobj.handle.clone();
+                    //Access write gaurd for socket handle
```
JustinCappos:
```suggestion
                    //Access write guard for socket handle
```

Does this access the write guard?  How?

JustinCappos (changes requested):
Overall, it's a good first pass.  At a high level:

1) focus on why the code does something instead of what.  For example, why did you insert an item into a data structure?

2) Feel free to say when you are not sure.  This is totally normal and something I do when writing my own code.  Don't be shy to do this more often.

3) if you want to cite a source go ahead.  For example, if you read a manpage to understand something or looked at an article, it's okay to drop a link, especially if the concept is complicated.  This might be helpful for some of the setsockopt activities and related flags.

src/safeposix/syscalls/net_calls.rs
davidge20: I found a source that explains the importance of  the SO_RESUSEPORT socket option.

https://lwn.net/Articles/542629/

I was wondering if some post like this is valid to cite since I'm not exactly sure of the credentials of the site but it seems legit

JustinCappos: Sure.  You should cite whatever you looked at.

JustinCappos:
Please ping me for a review after you've had a chance to address all the comments.

davidge20:
> Please ping me for a review after you've had a chance to address all the comments.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -12,6 +12,7 @@ use crate::safeposix::filesystem::*;
 use crate::safeposix::net::*;
 
 impl Cage {
+    //Initializes a socket description and sets the necessary flags
```
rennergade: should be "socket file descriptor"

src/safeposix/syscalls/net_calls.rs
```diff
@@ -36,29 +44,49 @@ impl Cage {
...
     fn _socket_inserter(&self, sockfd: FileDescriptor) -> i32 {
         let (fd, guardopt) = self.get_next_fd(None);
+        //In the case that no fd is available from the call to get_next_fd,
+        //fd is set to -ENFILE = -23 and the error is propagated forward
         if fd < 0 {
             return fd;
         }
         let fdoption = &mut *guardopt.unwrap();
-        let _insertval = fdoption.insert(sockfd);
+        //Insert the sockfd as the FileDescriptor inside fdoption
```
rennergade: The option doesn't say anything about whats inside it, it just is None or Some(). By inserting into it indicates that there is now a fd at that index

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +187,64 @@ impl Cage {
...
+        //The innersocket is an Option wrapped around the raw sys fd
+        //Depending on domain, innersocket may or may not be necessary
+        //Ex: Unix sockets do not have a kernal fd
+        //If innersocket is not available, then create a socket and insert its
+        //fd into innersocket
         if let None = sockhandle.innersocket {
+            //Create a new socket, as no sockhandle exists
+            //Socket creation rarely fails except for invalid parameters or extremely low-resources conditions
+            //Upon failure, process will panic
+            // ** What domains does lind satisfy???? ** //
```
rennergade: we handle ipv4, ipv6, and unix

src/safeposix/syscalls/net_calls.rs
```diff
@@ -244,18 +310,31 @@ impl Cage {
                 let filename = truepath.file_name().unwrap().to_str().unwrap().to_string(); //for now we assume this is sane, but maybe this should be checked later
 
                 //this may end up skipping an inode number in the case of ENOTDIR, but that's not catastrophic
+                //FS_METADATA contains information about the file system
                 let newinodenum = FS_METADATA
                     .nextinode
-                    .fetch_add(1, interface::RustAtomicOrdering::Relaxed); //fetch_add returns the previous value, which is the inode number we want
+                    .fetch_add(1, interface::RustAtomicOrdering::Relaxed); 
+                    //fetch_add returns the previous value, which is the inode number we want,
```
rennergade: this is good for your understanding but im not sure this needs a lenghty comment, but ill defer to Justin

rennergade (changes requested):
Requested some minor changes. Also should add some tests, bind is used in most of the net tests in the test suite already, but worth adding some explicit edge cases.

davidge20:
@JustinCappos pinging for further review after updates

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +187,64 @@ impl Cage {
...
+
+                //Otherwise, set the socket option in the new socket
+                //The level argument specifies the protocol level at which the option resides
+                //In our case, we are setting options at the socket level
+                //To learn more about setsockopt https://man7.org/linux/man-pages/man3/setsockopt.3p.html
                 let sockret = thissock.setsockopt(SOL_SOCKET, reuse, 1);
+                //Failure occured upon setting a socket option
+                //Possible failures can be read at the man page linked above
+                //
+                // **Possibly add errors instead of having a single panick for all errors ??? ** //
```
JustinCappos:
```suggestion
                // TODO: Possibly add errors instead of having a single panic for all errors ??? ** //
```

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +187,64 @@ impl Cage {
...
 
     //we assume we've converted into a RustSockAddr in the dispatcher
+    //bind a name to a socket - assign the address specified by localaddr to the 
+    //socket referred to by the file descriptor fd
     pub fn bind_syscall(&self, fd: i32, localaddr: &interface::GenSockaddr) -> i32 {
         self.bind_inner(fd, localaddr, false)
     }
 
+    //Based on the domain of the socket, bind the socket with a local address and 
+    //insert a cloned local address in the socket handle
```
JustinCappos: But what is an "inner" socket?  What is the purpose of this?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -159,35 +187,64 @@ impl Cage {
...
             }
 
+            //Insert the socket file descriptor into the innersocket value
             sockhandle.innersocket = Some(thissock);
         };
     }
 
     //we assume we've converted into a RustSockAddr in the dispatcher
+    //bind a name to a socket - assign the address specified by localaddr to the 
+    //socket referred to by the file descriptor fd
```
JustinCappos: The bind_inner function seems pointless.  Why have it?

src/safeposix/syscalls/net_calls.rs
JustinCappos: I'd say the comment can be shortened, but something is needed here.  I'm actually more interested in *why* relaxed ordering was used instead of what relaxed ordering does...

JustinCappos (approved):
It's close.  Please fix the things I mentioned (if something isn't clear, let me know) and then have Yash or Nick merge...

src/safeposix/syscalls/net_calls.rs
rennergade: I'm pretty sure this is a remnant from a refactor at some point.

src/safeposix/syscalls/net_calls.rs
rennergade: We just went with the fastest ordering guaranteed to be atomic since we probably don't need to be more restrictive for a counter.

https://stackoverflow.com/questions/30407121/which-stdsyncatomicordering-to-use

rennergade (approved):
I think this is good to go. Approved!

JustinCappos (approved):
LGTM!

PR #264, opened 2024-06-14
Open_syscall comments and tests

namanlalitnyu:
## Description

Fixes # (issue)
The following changes include the tests and comments in the code for the "open_syscall" file system call under RustPosix.
The tests were added to cover all the possible scenarios that might happen when calling the file system_call `open_syscall`. 

### Type of change
- [ ]  This change just contains the tests for an existing file system call.

## How Has This Been Tested?
Inorder to run the tests, we need to run `cargo test --lib` command inside the `safeposix-rust` directory.

All the tests are present under this directory: `lind_project/src/safeposix-rust/src/tests/fs_tests.rs`

- Test A - `ut_lind_fs_open_empty_directory()`
- Test B - `ut_lind_fs_open_nonexisting_parentdirectory_and_file()`
- Test C - `ut_lind_fs_open_existing_parentdirectory_and_nonexisting_file()`
- Test D - `ut_lind_fs_open_existing_file_without_flags()`
- Test E -  `ut_lind_fs_open_existing_file_with_flags()`
- Test F -  `ut_lind_fs_open_create_new_file_and_check_link_count()`
- Test G -  `ut_lind_fs_open_existing_file_with_o_trunc_flag()`
- Test H -  `ut_lind_fs_open_new_file_with_s_ifchar_flag()`

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -11,31 +11,34 @@ use crate::safeposix::net::NET_METADATA;
 use crate::safeposix::shm::*;
 
 impl Cage {
-    //------------------------------------OPEN SYSCALL------------------------------------
-    // Description
-    // The open_syscall() creates an open file description that refers to a file and a file descriptor that refers to that open file description.
-    // The file descriptor is used by other I/O functions to refer to that file.
-    // There are generally two cases which occur when this function is called. 
-    // Case 1: If the file to be opened doesn't exist, then a new file is created at the given location and a new file descriptor is created.
-    // Case 2: If the file already exists, then a few conditions are checked and based on them, file is updated accordingly.
+    /// ## ------------------OPEN SYSCALL------------------
+    /// ### Description
+    /// The open_syscall() creates an open file description that refers to a file and a file descriptor that refers to that open file description.
+    /// The file descriptor is used by other I/O functions to refer to that file.
+    /// There are generally two cases which occur when this function is called. 
+    /// Case 1: If the file to be opened doesn't exist, then a new file is created at the given location and a new file descriptor is created.
+    /// Case 2: If the file already exists, then a few conditions are checked and based on them, file is updated accordingly.
     
-    // Function Arguments
-    // The open_syscall() receives three arguments:
-    // 1. Path - This argument points to a pathname naming the file.
-    //           For example: "/parentdir/file1" represents a file which will be either opened if exists or will be created at the given path.
-    // 2. Flags - This argument contains the file status flags and file access modes which will be alloted to the open file description.
-    //            The flags are combined together using a bitwise-inclusive-OR and the result is passed as an argument to the function.
-    //            Some of the most common flags used are: O_CREAT | O_TRUNC | O_RDWR | O_EXCL | O_RDONLY | O_WRONLY, with each representing a different file mode.
-    // 2. Mode - This represents the permission of the newly created file. 
-    //           The general mode used is "S_IRWXA": which represents the read, write, and search permissions on the new file. 
+    /// ### Function Arguments
+    /// The open_syscall() receives three arguments:
+    /// 1. Path - This argument points to a pathname naming the file.
+    ///           For example: "/parentdir/file1" represents a file which will be either opened if exists or will be created at the given path.
+    /// 2. Flags - This argument contains the file status flags and file access modes which will be alloted to the open file description.
+    ///            The flags are combined together using a bitwise-inclusive-OR and the result is passed as an argument to the function.
+    ///            Some of the most common flags used are: O_CREAT | O_TRUNC | O_RDWR | O_EXCL | O_RDONLY | O_WRONLY, with each representing a different file mode.
+    /// 3. Mode - This represents the permission of the newly created file. 
+    ///           The general mode used is "S_IRWXA": which represents the read, write, and search permissions on the new file. 
     
-    // Return Values
-    // Upon successful completion of this call, a file descriptor is returned which points the file which is opened.
-    // Otherwise, an error with a proper errorNumber and errorMessage is returned based on the different scenarios.
-    //
-    // Tests
-    // All the different scenarios for open_syscall() are covered and tested in the "fs_tests.rs" file under "open_syscall_tests" section.
-    //
+    /// ### Return Values
+    /// Upon successful completion of this call, a file descriptor is returned which points the file which is opened.
+    /// Otherwise, an error with a proper errorNumber and errorMessage is returned based on the different scenarios.
+    ///
+    /// ### Tests
+    /// All the different scenarios for open_syscall() are covered and tested in the "fs_tests.rs" file under "open_syscall_tests" section.
+    ///
+    /// for more detailed description of all the commands and return values, see 
+    /// [open(2)](https://man7.org/linux/man-pages/man2/open.2.html)
+    ///
 
     // This function is used to create a new File Descriptor Object and return it.
     // This file descriptor object is then inserted into the File Descriptor Table of the associated cage in the open_syscall() function
@@ -62,7 +65,8 @@ impl Cage {
 
         // Fetch the next file descriptor and its lock write guard to ensure the file can be associated with the file descriptor 
         let (fd, guardopt) = self.get_next_fd(None);
-        // Return an error when no file descriptor is available
+        // The above function returns a valid file descriptor, but incase there is any error related to the filedescriptor processing
+        // or if the file descriptor is invalid, the return value is always an error with value < 0, so we check the same in the code below.
         if fd < 0 {
             return syscall_error(
                 Errno::ENFILE,
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -12,9 +12,34 @@ use crate::safeposix::shm::*;
 
 impl Cage {
     //------------------------------------OPEN SYSCALL------------------------------------
+    // Description
+    // The open_syscall() creates an open file description that refers to a file and a file descriptor that refers to that open file description.
+    // The file descriptor is used by other I/O functions to refer to that file.
+    // There are generally two cases which occur when this function is called. 
+    // Case 1: If the file to be opened doesn't exist, then a new file is created at the given location and a new file descriptor is created.
+    // Case 2: If the file already exists, then a few conditions are checked and based on them, file is updated accordingly.
+    
+    // Function Arguments
+    // The open_syscall() receives three arguments:
+    // 1. Path - This argument points to a pathname naming the file.
+    //           For example: "/parentdir/file1" represents a file which will be either opened if existed or will be created at the given path.
+    // 2. Flags - This argument contains the file status flags and file access modes which will be alloted to the open file description.
+    //            The flags are combined together using a bitwise-inclusive-OR and the result is passed as an argument to the function.
+    //            Some of the most common flags used are: O_CREAT | O_TRUNC | O_RDWR | O_EXCL | O_RDONLY | O_WRONLY, with each representing a different file mode.
+    // 2. Mode - This represents the permission of the newly created file. 
+    //           The general mode used is "S_IRWXA": which represents the read, write, and search permissions on the new file. 
+    
+    // Return Values
+    // Upon successful completion of this call, a file descriptor is returned which points the file which is opened.
+    // Otherwise, an error with a proper errorNumber and errorMessage is returned based on the different scenarios.
+    //
+    // Tests
+    // All the different scenarios for open_syscall() are covered and tested in the "fs_tests.rs" file under "open_syscall_tests" section.
+    //
 
+    // This function is used to create a new File Descriptor Object and return it.
+    // This file descriptor object is then inserted into the File Descriptor Table of the associated cage in the open_syscall() function
     fn _file_initializer(&self, inodenum: usize, flags: i32, size: usize) -> FileDesc {
-        //insert file descriptor into self.filedescriptortableable of the cage
         let position = if 0 != flags & O_APPEND { size } else { 0 };
         let allowmask = O_RDWRFLAGS | O_CLOEXEC;
         FileDesc {
@@ -26,21 +51,34 @@ impl Cage {
     }
 
     pub fn open_syscall(&self, path: &str, flags: i32, mode: u32) -> i32 {
-        //Check that path is not empty
+        // Check that the given input path is not empty
         if path.len() == 0 {
             return syscall_error(Errno::ENOENT, "open", "given path was null");
         }
+        
+        // Retrieve the absolute path from the root directory. The absolute path is then used to validate directory paths
+        // while navigating through subdirectories and creating a new file or open existing file at the given location.
         let truepath = normpath(convpath(path), self);
 
+        // Fetch the next file descriptor and its lock write guard to ensure the file can be associated with the file descriptor 
         let (fd, guardopt) = self.get_next_fd(None);
+        // Return an error when no file descriptor is available
         if fd < 0 {
-            return fd;
+            return syscall_error(
+                Errno::ENFILE,
+                "open_helper",
+                "no available file descriptor number could be found",
+            );
         }
+        // File Descriptor Write Lock Guard
         let fdoption = &mut *guardopt.unwrap();
 
+        // Walk through the absolute path which returns a tuple consisting of inode number of file (if it exists), and inode number of parent (if it exists)
         match metawalkandparent(truepath.as_path()) {
-            //If neither the file nor parent exists
+            // Case 1: When neither the file directory nor the parent directory exists
             (None, None) => {
+                // O_CREAT flag is used to create a file if it doesn't exist.
+                // If this flag is not present, then a file can not be created and error is returned.
                 if 0 == (flags & O_CREAT) {
                     return syscall_error(
                         Errno::ENOENT,
@@ -48,11 +86,13 @@ impl Cage {
                         "tried to open a file that did not exist, and O_CREAT was not specified",
                     );
                 }
+                // O_CREAT flag is set but the path doesn't exist, so return an error with a different message string.
                 return syscall_error(Errno::ENOENT, "open", "a directory component in pathname does not exist or is a dangling symbolic link");
             }
 
-            //If the file doesn't exist but the parent does
+            // Case 2: When the file doesn't exist but the parent directory exists
             (None, Some(pardirinode)) => {
+                // Check if O_CREAT flag is not present, then a file can not be created and error is returned.
                 if 0 == (flags & O_CREAT) {
                     return syscall_error(
                         Errno::ENOENT,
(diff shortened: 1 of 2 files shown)


namanlalitnyu:
Completed with the comments in the sys_call and added a few tests for it. But, still working on the remaining ones and will update the PR within some time once the other tests are completed.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -12,9 +12,34 @@ use crate::safeposix::shm::*;
...
+    // The open_syscall() creates an open file description that refers to a file and a file descriptor that refers to that open file description.
+    // The file descriptor is used by other I/O functions to refer to that file.
+    // There are generally two cases which occur when this function is called. 
+    // Case 1: If the file to be opened doesn't exist, then a new file is created at the given location and a new file descriptor is created.
+    // Case 2: If the file already exists, then a few conditions are checked and based on them, file is updated accordingly.
+    
+    // Function Arguments
+    // The open_syscall() receives three arguments:
+    // 1. Path - This argument points to a pathname naming the file.
+    //           For example: "/parentdir/file1" represents a file which will be either opened if existed or will be created at the given path.
```
JustinCappos:
```suggestion
    //           For example: "/parentdir/file1" represents a file which will be either opened if it exists or will be created at the given path.
```

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,33 +51,48 @@ impl Cage {
...
 
     pub fn open_syscall(&self, path: &str, flags: i32, mode: u32) -> i32 {
-        //Check that path is not empty
+        // Check that the given input path is not empty
         if path.len() == 0 {
             return syscall_error(Errno::ENOENT, "open", "given path was null");
         }
+        
+        // Retrieve the absolute path from the root directory. The absolute path is then used to validate directory paths
+        // while navigating through subdirectories and creating a new file or open existing file at the given location.
```
JustinCappos, line 76: Does this also follow symlinks?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,33 +51,48 @@ impl Cage {
...
         let truepath = normpath(convpath(path), self);
 
+        // Fetch the next file descriptor and its lock write guard to ensure the file can be associated with the file descriptor 
         let (fd, guardopt) = self.get_next_fd(None);
+        // Return an error when no file descriptor is available
         if fd < 0 {
```
JustinCappos: Will the same error value be returned for all error types?  should you check for a certain error return value and panic if you get something else?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,33 +51,48 @@ impl Cage {
...
+                "no available file descriptor number could be found",
+            );
         }
+        // File Descriptor Write Lock Guard
         let fdoption = &mut *guardopt.unwrap();
 
+        // Walk through the absolute path which returns a tuple consisting of inode number of file (if it exists), and inode number of parent (if it exists)
         match metawalkandparent(truepath.as_path()) {
-            //If neither the file nor parent exists
+            // Case 1: When neither the file directory nor the parent directory exists
```
JustinCappos: It's sort of weird to structure it like this because if the parent director doesn't exist, then the file clearly won't.  Can you just check to make sure the parent dir exists and return an error otherwise?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -61,62 +101,77 @@ impl Cage {
...
-                let filename = truepath.file_name().unwrap().to_str().unwrap().to_string(); //for now we assume this is sane, but maybe this should be checked later
-
+                // Error is thrown when the input flags contain S_IFCHR flag representing a special character file.
                 if S_IFCHR == (S_IFCHR & flags) {
                     return syscall_error(Errno::EINVAL, "open", "Invalid value in flags");
                 }
 
-                let effective_mode = S_IFREG as u32 | mode;
-
+                // Check for the condition if the mode bits are correct and have the required permissions to create a directory
```
JustinCappos: I'm pretty confused about what is happening here.  What is S_FILETYPEFLAGS?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -61,62 +101,77 @@ impl Cage {
...
+                // Create a new inode of type "File" representing a file and set the required attributes
                 let newinode = Inode::File(GenericInode {
-                    size: 0,
+                    size: 0, 
                     uid: DEFAULT_UID,
                     gid: DEFAULT_GID,
                     mode: effective_mode,
-                    linkcount: 1,
-                    refcount: 1,
+                    linkcount: 1, // because when a new file is created, it has a single hard link, which is the directory entry that points to this file's inode.
```
JustinCappos: What is a linkcount vs a refcount?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -126,39 +181,57 @@ impl Cage {
...
+                        //This is a special case when the input flags contain "O_TRUNC" flag, 
+                        //This flag truncates the file size to 0, and the mode and owner are unchanged
+                        // and is only used when the file exists and is a regular file
                         if O_TRUNC == (flags & O_TRUNC) {
-                            // We only do this to regular files, otherwise O_TRUNC is undefined
-                            //close the file object if another cage has it open
+                            // Close the existing file object and remove it from the FileObject Hashtable using the inodenumber
                             let entry = FILEOBJECTTABLE.entry(inodenum);
                             if let interface::RustHashEntry::Occupied(occ) = &entry {
+                                // Get the entry for the current file associated with the inodeNumber and close the opened it
```
JustinCappos: typo in comment.

Also, why do we close it?  What is the reason?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -126,39 +181,57 @@ impl Cage {
...
                                 occ.get().close().unwrap();
                             }
-                            // resize it to 0
+
+                            // Reset the size of the file to be 0
```
JustinCappos:
```suggestion
```

Can likely omit since this is obvious from the code.

JustinCappos (changes requested):
Good first cut.  A few minor changes requested.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -12,9 +12,34 @@ use crate::safeposix::shm::*;
...
+    // All the different scenarios for open_syscall() are covered and tested in the "fs_tests.rs" file under "open_syscall_tests" section.
+    //
 
+    // This function is used to create a new File Descriptor Object and return it.
+    // This file descriptor object is then inserted into the File Descriptor Table of the associated cage in the open_syscall() function
     fn _file_initializer(&self, inodenum: usize, flags: i32, size: usize) -> FileDesc {
-        //insert file descriptor into self.filedescriptortableable of the cage
         let position = if 0 != flags & O_APPEND { size } else { 0 };
         let allowmask = O_RDWRFLAGS | O_CLOEXEC;
         FileDesc {
```
ve1nard, line 61: Why out of all file creation flags only access mode and cloexec flags are stored in the FileDesc struct?

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: I checked the implementation for this function call (`get_next_fd`) and whenever there is success, it returns a value > 0. But in the remaining cases, including the error, we always get the same negative value returned from this function. So, that's why I think this implementation works fine.
I'll update the comment in the code for better clarity.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -61,62 +101,77 @@ impl Cage {
...
-
+                // Error is thrown when the input flags contain S_IFCHR flag representing a special character file.
                 if S_IFCHR == (S_IFCHR & flags) {
                     return syscall_error(Errno::EINVAL, "open", "Invalid value in flags");
                 }
 
-                let effective_mode = S_IFREG as u32 | mode;
-
+                // Check for the condition if the mode bits are correct and have the required permissions to create a directory
                 if mode & (S_IRWXA | S_FILETYPEFLAGS as u32) != mode {
```
ve1nard: Here you check if the only set mode bits are S_IRWXA or S_FILETYPEFLAGS. Why are they the only sane mode flags?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -61,62 +101,77 @@ impl Cage {
...
                     return syscall_error(Errno::EPERM, "open", "Mode bits were not sane");
-                } //assert sane mode bits
+                }
 
+                let filename = truepath.file_name().unwrap().to_str().unwrap().to_string(); //for now we assume this is sane, but maybe this should be checked later
                 let time = interface::timestamp(); //We do a real timestamp now
+                let effective_mode = S_IFREG as u32 | mode;
```
ve1nard: Why is a regular file type flag added to the effective mode flags?

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: In the current implementation, I can see that the FileObjectTable contains entries about the `open` files in the system. So, when the TRUNC flag is set, we close the existing opened file and remove its entry from the FileObjectTable. Then the size is set to 0, file is inserted back to the FileObjectTable in the open state.

src/tests/fs_tests.rs
```diff
@@ -1339,4 +1344,50 @@ pub mod fs_tests {
...
+        assert_eq!(cage.open_syscall(path, S_IFCHR | O_CREAT, S_IRWXA), -(Errno::EINVAL as i32));
+        
+        // Check for error when parent directory exists but file doesn't exist and mode bits are invalid
+        let invalid_mode = 0o77777;
+        assert_eq!(cage.open_syscall(path, O_CREAT, invalid_mode), -(Errno::EPERM as i32));
+
+        assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
```
ve1nard, line 1397: should you add a test case for when both parent directory and file exist?

src/safeposix/syscalls/fs_calls.rs
rennergade: If you just want clarity the tuple check in match statement would just be (.., None) since the second field is the parent and the first field doesn't matter. Thought metawalkparent will always return (None, None) if the parent doesn't exist since like you said if we don't find a parent we obviously don't have the file.

src/tests/fs_tests.rs
namanlalitnyu: Yes, will be adding the remaining tests in an updated review soon.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: When creating a new file, we need to specify its type. S_IFREG is the flag for a regular file, so it’s added to the mode to indicate that the new file being created is a regular file.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: These flags represent the permissions and file type respectively, which are the core attributes of a file's mode.
Any other bits set in the mode would be invalid or unsupported, hence, the sanity check ensures only these valid bits are used.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: This flag represents a bitmask that can be used to extract the file type information from a file's mode. 
In this code, we are checking whether the mode bits are sane by ensuring that only valid file permission bits (S_IRWXA) and file type bits (S_FILETYPEFLAGS) are set. Else, we return the error.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: While creating a new FileDescriptor, there are two important things that need to be present:
1. `Access Mode`:- The given flags determine whether the file is opened for reading, writing, or both.
2. `O_CLOEXEC` - This flag indicates that the file descriptor should be automatically closed during an exec family function. It’s needed for managing file descriptors across different processes, ensuring that they do not unintentionally remain open.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: @rennergade can you please share your thoughts on this?

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: LinkCount - It refers to the hard links pointing to an inode, indicates the number of directory entries referring to this inode. 
ReferenceCount - It refers to the number of active file descriptors open, that refer to that file.

src/safeposix/syscalls/fs_calls.rs
rennergade: We have no concept of symlinks so no

src/safeposix/syscalls/fs_calls.rs
JustinCappos: In general, it's a good idea to assert or similar in a situation like this because if they change that function later, your code would behave in unexpected ways.

src/safeposix/syscalls/fs_calls.rs
JustinCappos: Where is this defined?  Is this a new thing in Lind?  I can't find it with google...

src/safeposix/syscalls/fs_calls.rs
JustinCappos: Why can multiple active file descriptors refer to a file?  What is the purpose?  (Ask @rennergade for help if you're not sure.)

JustinCappos:
@namanlalitnyu A few minor github things...

1) feel free to resolve conversations after you think you've addressed something.  

2) it's better to address a question with a comment in the code (when this makes sense) instead of only on the issue.  When we read the code later, we've lost the q/a.  

This isn't a huge problem here (you've done a nice job!), but I wanted to bring them up since they are top-of-mind)

src/safeposix/syscalls/fs_calls.rs
rennergade: So this is stuff that was mostly ported from Lind-Repy, it seems to be from one of your [commits like 12 years ago. ](https://github.com/Lind-Project/nacl_repy/blame/af9037e60816ff712edf8c8196db45fac30bbed2/seattlelib/lind_fs_constants.py#L61)

It seems like these are  "all of the above" constants?

JustinCappos: Ah, good to know!  I should have commented it better before and we should comment it now.

namanlalitnyu:
> @namanlalitnyu A few minor github things...
> 
> 1. feel free to resolve conversations after you think you've addressed something.
> 2. it's better to address a question with a comment in the code (when this makes sense) instead of only on the issue.  When we read the code later, we've lost the q/a.
> 
> This isn't a huge problem here (you've done a nice job!), but I wanted to bring them up since they are top-of-mind)

Sure professor, thanks for the suggestions. Will do that!

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Had a discussion with Nick related to this. 
So, this case can happen as if we duplicate a fd, then we have multiple fds pointing to a file. If we fork, then we have two fds pointing to a file. 
This is needed to keep a track of the references, though in the Inode Structure, we only store LinkCount for a particular file. Reference Count is not stored as it only persists on an individual run.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Understood, have added matching checks for it. Can you please verify if this is the expected way? Or any other thoughts on the approach would be great.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,158 +59,225 @@ impl Cage {
...
                     }
                 }
 
-                let _insertval =
-                    fdoption.insert(File(self._file_initializer(inodenum, flags, size)));
+                // Once all the updates are done, the file descriptor value is returned
+                fd 
+            },
+            // If the file descriptor is invalid, the return value is always an error with value -23 (ENFILE).
+            -23 => {
```
yashaswi2000: can you use the Errno ENUM here instead of hardcoding it. so that its clear what we are expecting out of the function. and maybe the order of erroring out can be different like this can be checked before the happy case. so we error out quicker.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,158 +59,225 @@ impl Cage {
...
+                return syscall_error(
+                    Errno::ENFILE,
+                    "open_helper",
+                    "no available file descriptor number could be found",
+                );
+            },
+            // Return a generic error when there is some other issue fetching the file descriptor.
+            _ => {
```
yashaswi2000, line 281: should we panic in this case? instead of throwing an error

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,158 +59,225 @@ impl Cage {
...
-                        "tried to create a file as a child of something that isn't a directory",
-                    );
-                }
-                FS_METADATA.inodetable.insert(newinodenum, newinode);
-                log_metadata(&FS_METADATA, pardirinode);
-                log_metadata(&FS_METADATA, newinodenum);
+                                    // Remove the previous file and add a new one of 0 length
+                                    if let interface::RustHashEntry::Occupied(occ) = entry {
+                                        occ.remove_entry();
+                                    }
```
rennergade, line 218: remove this extra whitespace

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -26,158 +59,225 @@ impl Cage {
...
-                let mut inodeobj = FS_METADATA.inodetable.get_mut(&inodenum).unwrap();
-                match *inodeobj {
-                    Inode::File(ref mut f) => {
-                        if O_TRUNC == (flags & O_TRUNC) {
-                            // We only do this to regular files, otherwise O_TRUNC is undefined
-                            //close the file object if another cage has it open
-                            let entry = FILEOBJECTTABLE.entry(inodenum);
-                            if let interface::RustHashEntry::Occupied(occ) = &entry {
-                                occ.get().close().unwrap();
+                                // Doubt: Why are we removing the previous entry from the FileObjectTable and then inserting a new file with size 0.
```
rennergade: This is  hacky way to truncate since a new file will be size 0 instead of actually adjusting the filesize and pointer. I would comment that thats what were doing but also label it as a possible bug.

rennergade (changes requested):
Two minor changes, should be able to get this merged today.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: I checked, and there is not any extra whitespace here; it is just because of the indentation as the code section has moved inside the matching block.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Updated the check with the Enum value and also changed the ordering for it.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Yeah, panic seems a better choice as it is a case when some execution error occurs. Updated the code for it.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Updated the comment.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -11,11 +11,41 @@ use crate::safeposix::net::NET_METADATA;
...
+    ///            The flags are combined together using a bitwise-inclusive-OR and the result is passed as an argument to the function.
+    ///            Some of the most common flags used are: O_CREAT | O_TRUNC | O_RDWR | O_EXCL | O_RDONLY | O_WRONLY, with each representing a different file mode.
+    /// 3. Mode - This represents the permission of the newly created file. 
+    ///           The general mode used is "S_IRWXA": which represents the read, write, and search permissions on the new file. 
+    
+    /// ### Return Values
+    /// Upon successful completion of this call, a file descriptor is returned which points the file which is opened.
+    /// Otherwise, an error with a proper errorNumber and errorMessage is returned based on the different scenarios.
+    ///
+    /// for more detailed description of all the commands and return values, see
```
JustinCappos, line 47: We don't cover all the cases of open though, do we?  What errors are returned and which are not?

JustinCappos (approved):
Looks fine after the return values / panics in the comment block is updated.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: No, we don't cover all the errors. Manpage has a lot of them mentioned which are remaining.

namanlalitnyu:
> Looks fine after the return values / panics in the comment block is updated.

Updated the comments with error values and panics in the comment block.

yashaswi2000 (approved):
LGTM!

rennergade:
I think this looks ready. Good job!

Issue #248, opened 2024-05-21
Intermittent Semaphore Test Errors in suite


rennergade:
We seem to be seeing some errors in some of the ut_lind_fs_sem_* functions intermittently. We need to diagnose the problem and fix it. I will post a bug log next time I see it.

rennergade:
```
---- tests::main_tests::tests stdout ----
FS TESTS
thread '<unnamed>' panicked at src/tests/fs_tests.rs:1118:13:
assertion `left == right` failed
  left: 0
 right: 1
```

here is the bug, it seems were failing this assertion on line 1118 of fs_tests.rs

PR #268, opened 2024-06-20
Fixes for integration tests of sem_syscalls

yashaswi2000:
## Description
This is a fix for [#248] issue. where we see intermittent test failures which uses semaphores. and performs actions like
- sem_wait
- sem_post
- sem_timed_wait
- sem_try_wait

### The tests effected are:
- ut_lind_fs_sem_fork()
- ut_lind_fs_sem_trytimed()
- ut_lind_fs_sem_test()

### Causes:

* The first kind of issue was assertion failures when sem_getvalue_syscall is performed outside of the critical section of parent/child process. which causes nondeterministic outputs since once out of critical section the busy waiting thread/process will go into its critical section and modify the semaphore value. 

* The second type is performing sem_try_wait call to enter into critical section, and always asserting a success from the call. sem_try_wait is different from sem_wait in the sense there is no busy waiting and thus if another thread/process is in the critical section, it just returns false. 

* The third kind of issue is performing sem_destroy before the child process exits its critical section. which causes nondeterministic behaviour, leading to child's sem_post call to throw error. parent process has to wait for all the threads/process to exit before destroying semaphore and shared memory.

Fixes # (issue)

 

### Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/tests/fs_tests.rs
+++ b/src/tests/fs_tests.rs
@@ -1166,27 +1166,29 @@ pub mod fs_tests {
             assert_eq!(cage1.sem_post_syscall(shmatret as u32), 0);
             cage1.exit_syscall(EXIT_SUCCESS);
         });
+
         //Parent processes
-        let thread_parent = interface::helper_thread(move || {
-            // Parents waits for the semaphore
-            assert_eq!(cage.sem_wait_syscall(shmatret as u32), 0);
-            assert_eq!(cage.sem_getvalue_syscall(shmatret as u32), 0);
-            interface::sleep(interface::RustDuration::from_millis(100));
-            // Parents release the semaphore
-            assert_eq!(cage.sem_post_syscall(shmatret as u32), 0);
-            assert_eq!(cage.sem_getvalue_syscall(shmatret as u32), 1);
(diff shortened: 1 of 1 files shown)


src/tests/fs_tests.rs
```diff
@@ -1210,40 +1212,41 @@ pub mod fs_tests {
...
+        
+        // wait for the child to exit before destroying the semaphore.
         thread_child.join().unwrap();
-        thread_parent.join().unwrap();
+
+        // Destroy the semaphore
+        assert_eq!(cage.sem_destroy_syscall(shmatret as u32), 0);
+        // mark the shared memory to be rmoved
+        let shmctlret2 = cage.shmctl_syscall(shmid, IPC_RMID, None);
+        assert_eq!(shmctlret2, 0);
```
JustinCappos, line 1243: Do you need to do this cleanup or does the exit syscall do it?  It might be useful to say either way in the comments here...

JustinCappos (approved):
Looks good to me!

rennergade (approved):

PR #269, opened 2024-06-20
Updates to Test structure for better visibility.

yashaswi2000:
## Description
Current structure of tests runs all the tests as part of 3 big functions and the tests are opaque to the rust test framework. 
it's hard to understand what is happening under the hood while tests are being run. This PR addresses this issue by refactoring the Test structure and adds synchronisation as tests in rust are run in parallel by default.

Fixes # (issue)

 

### Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/tests/fs_tests.rs
+++ b/src/tests/fs_tests.rs
@@ -7,7 +7,6 @@ pub mod fs_tests {
     use crate::safeposix::syscalls::fs_calls::*;
     use crate::safeposix::{cage::*, dispatcher::*, filesystem};
     use libc::c_void;
-    use crate::tests::setup::TESTMUTEX;
     use std::fs::OpenOptions;
     use std::os::unix::fs::PermissionsExt;
 
--- a/src/tests/fs_tests.rs
+++ b/src/tests/fs_tests.rs
@@ -14,12 +14,9 @@ pub mod fs_tests {
     #[test]
     pub fn ut_lind_fs_simple() {
 
-        //acquiring a lock on TESTMUTEX prevents other tests from running concurrently
-        let mut _thelock = TESTMUTEX.lock().unwrap();
-        //creates the file system if it doesnt exist yet.
-        setup::test_setup();
+        //acquiring a lock on TESTMUTEX prevents other tests from running concurrently, and also performs clean env setup
+        let _thelock = setup::lock_and_init();
 
-        lindrustinit(0);
         let cage = interface::cagetable_getref(1);
 
         assert_eq!(cage.access_syscall("/", F_OK), 0);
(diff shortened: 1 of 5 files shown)


src/tests/fs_tests.rs
```diff
@@ -1083,7 +1243,13 @@ pub mod fs_tests {
         lindrustfinalize();
     }
 
+    #[test]
     pub fn ut_lind_fs_shm() {
+        //acquiring a lock on TESTMUTEX prevents other tests from running concurrently
+        let mut _thelock = TESTMUTEX.lock().unwrap();
+        //creates the file system if it doesnt exist yet.
+        setup::test_setup();
```
JustinCappos: Would it be better to put this in a single "lock_and_init" function?  That way if you want to add something else, you don't need to go and modify the code everywhere?

(I would think that `lindrustinit(0);` would also be part of this...)

src/tests/ipc_tests.rs
```diff
@@ -301,17 +309,25 @@ pub mod ipc_tests {
...
     pub fn ut_lind_ipc_socketpair() {
+        //acquiring a lock on TESTMUTEX prevents other tests from running concurrently
+        let mut _thelock = TESTMUTEX.lock().unwrap();
+        //creates the file system if it doesnt exist yet.
+        setup::test_setup();
+
         lindrustinit(0);
         let cage = interface::cagetable_getref(1);
-        let mut socketpair = interface::SockPair::default();
+        let socketpairsafe = Arc::new(Mutex::new(interface::SockPair::default()));
```
JustinCappos: Why does this need to change as part of this?  (I'm curious.  Likely a minor comment about this will suffice...

src/tests/mod.rs
```diff
@@ -9,19 +9,33 @@ use std::net::{TcpListener, UdpSocket};
 use crate::interface;
 use crate::safeposix::{cage::*, filesystem::*};
 
+#[allow(unused_parens)]
```
JustinCappos, line 12: Why do we need this?

JustinCappos (commented):
I like the change!  A few style suggestions...

src/tests/fs_tests.rs
yashaswi2000: we could try something like that, but then it will be like a semaphore ig rather than a lock. since lock is tied to scope. that wouldn't work as it will go out of scope after lock_and_init call?

JustinCappos: Okay, I don't know enough about this, but that makes sense.  

What about combining setup::test_setup(); and lindrustinit(0); into a helper?

yashaswi2000: will explore the helper function idea!

src/tests/ipc_tests.rs
yashaswi2000: ohh, before we fixed the random ports issue. i was trying a few things to make this stable. but turned out irrelevant. reverting back as this change isnt needed.

src/tests/mod.rs
yashaswi2000: we have a couple of places that throw - "warning: unnecessary parentheses around assigned value". I think it was used to suppress such warning.

JustinCappos:
Okay, you might look to see if those are useless or appropriate.  If
appropriate (or even just not making the code a lot harder to read) it's
probably best to make clippy happy...

Yaxuan-w (approved):
Great work! Overall logic LGTM so approved.

src/tests/fs_tests.rs
yashaswi2000: implemented a lock_and_init function as suggested.

JustinCappos (approved):
Looks good!  Thanks for the update!

rennergade:
Cool, looks good to me too!

PR #271, opened 2024-06-21
Add tests and comments for select_syscall

qianxichen233:
## Description

This PR added more tests for the select syscall, and marked some TODO and BUG in existing select syscalls. Tests being added including:
1. test for select monitoring on five different file descriptors:
- Regular file
- AF_INET server socket waiting for two clients
- AF_INET server socket's connection file descriptor with clients
- AF_UNIX server socket's connection file descriptor with a client
- Pipe
2. test for select working with pipe writefd when pipe is full
3. test for select working with AF_UNIX socket pipe writefd when pipe is full
- **This test currently will fail since select did not check for this situation. BUG is marked in select_writefds and this tests is currently commented out**
4. test for timeout field of select, two sub-tests were conducted here:
- select when timeout could expire
- select when all arguments were None except for timeout
5. test for error case of select
- **test for sending signal to select is marked as TODO since sending signals using kill_syscall is currently not supported in raw safeposix environment**

Marked BUG including:
1. under select_writefds, AF_UNIX socket condition, socket's send pipe is not checked for writing, causing one of the test fail
2. under select_writefds, sockets are always considered to be writable, which is not true since socket could block in cases like the kernel send buffer is full.

Marked TODO including:
1. select for exceptfds: we currently don't implement exceptfds and possibly could if necessary
2. select test with signals are marked as to do since sending signals using kill_syscall is currently not supported in raw safeposix environment

Besides, more comments were added to select_syscalls and its sub-functions.

Other noteworthy things:
1. Some tests used non-necessary barriers to make sure some codes are running in order, this is because there are some unfixed issues (pipe reference counter, etc) that could cause the tests fail. Using extra barrier is a workaround to skip these unexpected unfixed issues and make the test work normally.

### Type of change

## How Has This Been Tested?

This PR is addition of new tests and currently all tests were passed.

## Checklist:


src/safeposix/syscalls/net_calls.rs
```diff
@@ -1859,6 +1859,38 @@ impl Cage {
...
     }
 
+    //------------------------------------SELECT SYSCALL------------------------------------
+    // Description
+    // select() allows a program to monitor multiple file descriptors, waiting until one or more of the file descriptors become "ready"
+    // for some class of I/O operation (e.g., input possible).  A file descriptor is considered ready if it is possible to perform a
+    // corresponding I/O operation (e.g., read(), or a sufficiently small write()) without blocking.
+    
+    // Function Arguments
+    // The open_syscall() receives three arguments:
```
rennergade: *select

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1859,6 +1859,38 @@ impl Cage {
...
+    //               write may still block.
+    //               After select() has returned, writefds will be cleared of all file descriptors except for those that are ready for writing.
+    // 4. exceptfds - currently not supported, only the validity of the fds will be checked
+    // 5. timeout - The timeout argument is a RustDuration structure that specifies the interval that select() should block
+    //              waiting for a file descriptor to become ready.  The call will block until either:
+    //              •  a file descriptor becomes ready;
+    //              •  the call is interrupted by a signal handler; or
+    //              •  the timeout expires.
+
+    // Return Values
```
rennergade: I believe you're also supposed to add possible errors

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1859,6 +1859,38 @@ impl Cage {
         }
     }
 
+    //------------------------------------SELECT SYSCALL------------------------------------
```
JustinCappos: Needs some minor format changes to work well with Rustdoc.  See @yashaswi2000 's PR for tips...

JustinCappos (commented):
I'll take a longer look later after these comments are addressed.

rennergade:
General comment: your tests are very good and thorough but the commenting needs a bit of work. Try to re-read the code line by line and ask yourself, "if this was my first time looking at this, would it be obvious to why this thing is happening?", if the answer is no add a comment and tell us WHY that thing is happening.

src/safeposix/syscalls/net_calls.rs
qianxichen233: fixed

src/safeposix/syscalls/net_calls.rs
qianxichen233: Thanks, just updated the comments

qianxichen233:
@namanlalitnyu Could you please review my PR?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1859,6 +1859,40 @@ impl Cage {
...
+    /// ## ------------------SELECT SYSCALL------------------
+    /// ### Description
+    /// The `select_syscall()` allows a program to monitor multiple file descriptors, waiting until one or more of the file descriptors become "ready"
+    /// for some class of I/O operation (e.g., input possible).  A file descriptor is considered ready if it is possible to perform a
+    /// corresponding I/O operation (e.g., `read_syscall()`) without blocking.
+    
+    /// ### Function Arguments
+    /// The `select_syscall()` receives five arguments:
+    /// * `nfds` - This argument should be set to the highest-numbered file descriptor in any of the three sets, plus 1.  The
+    ///           indicated file descriptors in each set are checked, up to this limit.
```
namanlalitnyu: Adding the feedback that I also received for Rust Documentation. 
1. We need to wrap the comments to be <80 characters wide.
2. The maximum number of tabs can be 4, so we need to remove the extra whitespace here.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1867,6 +1901,7 @@ impl Cage {
         exceptfds: Option<&mut interface::FdSet>,
```
namanlalitnyu, line 2154: Can you explain a bit about `FdSet` in the form of comments? Currently, not able to understand it without looking at its implementation.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1859,6 +1859,40 @@ impl Cage {
...
+    ///               After select() has returned, writefds will be cleared of all file descriptors except for those that are ready for writing.
+    /// * `exceptfds` - currently not supported, only the validity of the fds will be checked
+    /// * `timeout` - The timeout argument is a RustDuration structure that specifies the interval that select() should block
+    ///              waiting for a file descriptor to become ready.  The call will block until either:
+    ///              •  a file descriptor becomes ready;
+    ///              •  the call is interrupted by a signal handler; or
+    ///              •  the timeout expires.
+
+    /// ### Returns
+    /// On success, select() and return the number of file descriptors contained in the two returned descriptor sets (that
```
namanlalitnyu: There is a small typo in the `return` comments.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1939,6 +1975,8 @@ impl Cage {
         return retval;
     }
 
+    // This function is used to select on readfds specifically
+    // It monitors every readfds to check if it is ready to read
```
namanlalitnyu: Should we include the comments for this function and other similar ones based on the Rust Documentation format to get some understanding about them, i.e., `///`?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1955,6 +1993,7 @@ impl Cage {
                 continue;
```
namanlalitnyu, line 2251: Based on the `if` block, shouldn't the comment for this check be changed to something like this `Ignore the fd not in readfds"?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1955,6 +1993,7 @@ impl Cage {
...
             }
 
+            // get the FileDescriptor Object from fd
             let checkedfd = self.get_filedescriptor(fd).unwrap();
             let unlocked_fd = checkedfd.read();
             if let Some(filedesc_enum) = &*unlocked_fd {
```
namanlalitnyu, line 2259: Can you explain a bit about what this condition is doing with some comments, currently not able to understand its use case.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1901,7 +1936,7 @@ impl Cage {
             }
 
             // 3. iterate thru exceptfds
-            // currently we don't really do select on execptfds, we just check if those fds are valid
+            // TODO: we currently don't implement exceptfds and possibly could if necessary
             if let Some(exceptfds_ref) = exceptfds.as_ref() {
                 for fd in 0..nfds {
                     // find the bit and see if it's on
```
davidge20, line 2196: what does this mean? what bit are we talking about? and why are we doing this ?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1966,6 +2005,8 @@ impl Cage {
                                 let sock_tmp = sockfdobj.handle.clone();
```
namanlalitnyu, line 2266: Can you add some comments about what each socket domain is (AF_UNIX, AF_INET, etc), and what are we trying to achieve inside each of the match blocks?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1966,6 +2005,8 @@ impl Cage {
...
                                 let sockhandle = sock_tmp.read();
                                 if sockhandle.state == ConnState::INPROGRESS {
+                                    // if connection state is INPROGRESS
+                                    // then check if the new connection
```
namanlalitnyu: I think the comment is incomplete here

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1966,6 +2005,8 @@ impl Cage {
...
                                     let remotepathbuf = normpath(
                                         convpath(sockhandle.remoteaddr.unwrap().path()),
                                         self,
```
namanlalitnyu, line 2283: What is the significance of this object `dsconnobj`?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2033,6 +2078,7 @@ impl Cage {
                     }
 
                     Pipe(pipefdobj) => {
+                        // check if the pipe has anything to read
                         if pipefdobj.pipe.check_select_read() {
                             new_readfds.set(fd);
                             *retval += 1;
```
davidge20, line 2365: why are we adding 1 to the retval here?

namanlalitnyu:
> @namanlalitnyu Could you please review my PR?

@qianxichen233, I added a few comments. Can you please address them and add more comments in the code to the remaining functions to get a better understanding of their implementation?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2062,6 +2108,8 @@ impl Cage {
         return 0;
```
davidge20, line 2391: what is update_readfds_from_kernel_select doing ? when does it fail with error?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2062,6 +2108,8 @@ impl Cage {
...
     }
 
+    // This function is used to select on writefds specifically
+    // It monitors every writefds to check if it is ready to write
     fn select_writefds(
```
davidge20, line 2400: what is the function doing if the fd is in writefds vs if it is not in writefds? there seems to be a bit more nuance than the comments above the function.

src/safeposix/syscalls/net_calls.rs
qianxichen233: Exceptfds is one of the argument of `select_syscall` that monitors exceptions specifically, and I think the comment already answered your question, that we do not support exceptfds currently, and this is a to-do that we might support it in the future if necessary

src/safeposix/syscalls/net_calls.rs
qianxichen233: `select` should return the number of fds that are ready as mentioned in comment, so each time we found a fd that is ready, we add 1 to the return value

src/safeposix/syscalls/net_calls.rs
qianxichen233: I don't quite understand what you mean. The `fds` in the argument is just all the writefds we need to monitor

davidge20: I think overall the description needs more details. For example, what exactly is "if it is ready to write" and how does it accomplish it.

qianxichen233: The answer of "when is a file descriptor ready to write/read" is already mentioned in the comments of `select_syscall` under Description. And for "how does it accomplish", the answer would be different for different file descriptor type, and is answered at the corresponding location

davidge20: Ahhh, thank you for pointing that out. Still a bit confused about what is going on in the case if the socket domain is AF_Unix.

qianxichen233: For AF_UNIX, what we are doing are two things (as mentioned at the corresponding if branch)
1. if the socket is in LISTEN state, that would mean the socket is trying to get some connections. So we read the accept table to see if there are any pending connections. If there is, the socket is marked as readable
2. if the socket is in CONNECTED state, the would mean the socket is already connected to a peer. If this case, we just check if peer has send anything to the pipe. So if the pipe is not empty, the socket is marked as readable
3. There is actually a third case that the socket is in INPROGRESS state, that would happen when the socket is calling `connect` with nonblocking flag. This feature is currently marked as BUG (maybe TODO as well) so you do not need to care about this right now

src/safeposix/syscalls/net_calls.rs
namanlalitnyu: @qianxichen233 Can you please reformat this code with `cargo fmt` command as looks like this section might need a bit of formatting.

src/safeposix/syscalls/net_calls.rs
namanlalitnyu: It would be great if you could add the same explanation in the comments, so we can understand it later once reading the code.

src/safeposix/syscalls/net_calls.rs
namanlalitnyu: This explanation looks good and can be added in the comments for `AF_UNIX` socket for other readers to get a better and quick understanding.

davidge20 (approved):
Looks good to me!

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2075,6 +2075,40 @@ impl Cage {
...
+    ///              •  the call is interrupted by a signal handler; or
+    ///              •  the timeout expires.
+
+    /// ### Returns
+    /// On success, select() and return the number of file descriptors contained in the two returned descriptor sets (that
+    /// is, the total number of bits that are set in readfds, writefds). The return value may be zero if the timeout expired
+    /// before any file descriptors became ready.
+    /// Otherwise, errors or panics are returned for different scenarios.
+    ///
+    /// ### Errors and Panics
```
rupeshkoushik07: Handle errors and panics separately

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2098,6 +2098,40 @@ impl Cage {
...
```
rennergade: This should be separated? Is there no situations where it panics?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2106,6 +2140,7 @@ impl Cage {
         exceptfds: Option<&mut interface::FdSet>,
         timeout: Option<interface::RustDuration>,
     ) -> i32 {
+        // nfds should be in the range
```
rennergade: what range? why?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2141,8 +2176,7 @@ impl Cage {
             }
 
             // 3. iterate thru exceptfds
-            // currently we don't really do select on execptfds, we just check if those fds
-            // are valid
+            // TODO: we currently don't implement exceptfds and possibly could if necessary
```
rennergade: and -> but

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2191,11 +2228,14 @@ impl Cage {
         let mut inet_info = SelectInetInfo::new();
 
         for fd in 0..nfds {
-            // check if current i is in readfd
+            // ignore file descriptors that is not in the set
```
rennergade: that are not in the set

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2180,6 +2215,8 @@ impl Cage {
         return retval;
     }
 
+    /// This function is used to select on readfds specifically
+    /// It monitors every readfds to check if it is ready to read
```
rennergade: "This function monitors all readfds to check if they are ready to read"

rennergade: What kinds of fds can read fds be? native sockets? in memory pipes? what else?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2191,11 +2228,14 @@ impl Cage {
...
 
         for fd in 0..nfds {
-            // check if current i is in readfd
+            // ignore file descriptors that is not in the set
             if !readfds.is_set(fd) {
                 continue;
             }
 
+            // try to get the FileDescriptor Object from fd number
+            // if the fd exists, do further process based on the file descriptor type
```
rennergade: *processing

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2204,9 +2244,19 @@ impl Cage {
                         let mut newconnection = false;
                         match sockfdobj.domain {
                             AF_UNIX => {
+                                // acquire the read access of the sockethandle
```
rennergade: sockethandle lock with read access

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2305,6 +2367,8 @@ impl Cage {
         return 0;
     }
 
+    /// This function is used to select on writefds specifically
```
rennergade, line 2394: similar comment as I had for readfds

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2313,23 +2377,33 @@ impl Cage {
         retval: &mut i32,
     ) -> i32 {
         for fd in 0..nfds {
-            // check if current i is in writefds
+            // ignore file descriptors that is not in the set
```
rennergade: are not

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2313,23 +2377,33 @@ impl Cage {
...
+            // try to get the FileDescriptor Object from fd number
+            // if the fd exists, do further process based on the file descriptor type
+            // otherwise, raise an error
             let checkedfd = self.get_filedescriptor(fd).unwrap();
             let unlocked_fd = checkedfd.read();
             if let Some(filedesc_enum) = &*unlocked_fd {
                 match filedesc_enum {
                     Socket(ref sockfdobj) => {
-                        // check if we've made an in progress connection first
+                        // acquire the read access of the sockethandle
```
rennergade: sockethandle lock with read access

src/safeposix/syscalls/net_calls.rs
```diff
@@ -2355,30 +2433,34 @@ impl Cage {
                             }
                         }
 
+                        // nonblocing AF_INET/AF_INET6 socket connection now established
```
rennergade: non-blocking

rennergade (changes requested):
A few cosmetic changes and need a bit more depth. Also remember to fmt this one.

qianxichen233:
@rennergade I just resolved these comments, could you check again?

rennergade (approved):
Ok great. Approved!

PR #275, opened 2024-06-22
chmod_syscall() updates

ve1nard:
## Description

Fixes # (issue)

The following changes include more elaborate comments and new unit tests for `chmod_syscall` and `fchmod_syscall`.

### Type of change

- [ ] More detailed comments for `chmod_syscall` and `fchmod_syscall`
- [ ] Shifted checking the validity of mode bits from `chmod_syscall()` to the `_chmod_helper()` helper function since it used by both `chmod_syscall()` and `fchmod_syscall()`
- [ ] Removed the unnecessary check for file type flags that allowed the file type to be changed
- [ ] New unit tests for `chmod_syscall` and `fchmod_syscall`

## How Has This Been Tested?

To run the tests, we need to run cargo test --lib command inside the safeposix-rust directory.

All the tests are present under this directory: lind_project/src/safeposix-rust/src/tests/fs_tests.rs

- Test A - `lut_lind_fs_chmod_valid_args()`
- Test B - `ut_lind_fs_chmod_invalid_args()`
- Test C - `ut_lind_fs_fchmod()`

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -2263,34 +2263,29 @@ impl Cage {
 
     /// ### Description
     ///
-    /// The `chmod_syscall()` changes a file's mode bits that consist of read, write, and execute file permission bits.
-    /// Changing set-user-ID, set-group-ID, and sticky bits is currently not supported.
+    /// The `_chmod_helper()` is a helper function used by both `chmod_syscall()` and `fchmod_syscall()` 
+    /// to change mode bits that consist of read, write, and execute file permission bits of a file
+    /// specified by an inode obtained from the corresponding caller syscall.
     ///
     /// ### Arguments
     ///
-    /// The `chmod_syscall()` accepts two arguments:
-    /// * `path` - pathname of the file whose mode bits we are willing to change (symbolic links are currently 
-    /// not supported). If the pathname is relative, then it is interpreted relative to the current working directory 
-    /// of the calling process.
+    /// The `_chmod_helper()` accepts two arguments:
+    /// * `inodenum` - an inode of a file whose mode bits we are willing to change obtained from the caller syscall.
     /// * `mode` - the new file mode, which is a bit mask created by bitwise-or'ing zero or more valid mode bits.
     /// Some of the examples of such bits are `S_IRUSR` (read by owner), `S_IWUSR` (write by owner), etc.
     ///
     /// ### Returns
     /// 
     /// Upon successful completion, zero is returned.
-    /// In case of a failure, -1 is returned, and `errno` is set depending on the error, e.g. EACCES, ENOENT, etc.
+    /// In case of a failure, an error is returned, and `errno` is set depending on the error, e.g. EACCES, ENOENT, etc.
     ///
     /// ### Errors and Panics
     ///
-    /// Currently, only two errors are supposrted:
-    /// * `EINVAL` - the value of the mode argument is invalid 
-    /// * `ENOENT` - a component of path does not name an existing file
+    /// Currently, only one error is supported:
+    /// * `EINVAL` - the value of the mode argument is invalid.
     /// Other errors, like `EFAULT  , `ENOTDIR`, etc. are not supported.
     ///
-    /// There are no cases where this syscall panics.
-    ///
-    /// To learn more about the syscall, valid mode bits, and error values, see
-    /// [chmod(2)](https://man7.org/linux/man-pages/man2/chmod.2.html)
+    /// There are no cases where this helper function panics.
 
     pub fn _chmod_helper(inodenum: usize, mode: u32) -> i32 {
         //S_IRWXA is a result of bitwise-or'ing read, write, and execute or search permissions for the file owner, group owners, 
@@ -2337,6 +2332,37 @@ impl Cage {
         }
     }
 
+    /// ### Description
+    ///
+    /// The `chmod_syscall()` changes a file's mode bits that consist of read, write, and execute file permission bits.
+    /// Changing set-user-ID, set-group-ID, and sticky bits is currently not supported.
+    ///
+    /// ### Arguments
+    ///
+    /// The `chmod_syscall()` accepts two arguments:
+    /// * `path` - pathname of the file whose mode bits we are willing to change (symbolic links are currently 
+    /// not supported). If the pathname is relative, then it is interpreted relative to the current working directory 
+    /// of the calling process.
+    /// * `mode` - the new file mode, which is a bit mask created by bitwise-or'ing zero or more valid mode bits.
+    /// Some of the examples of such bits are `S_IRUSR` (read by owner), `S_IWUSR` (write by owner), etc.
+    ///
+    /// ### Returns
+    /// 
+    /// Upon successful completion, zero is returned.
+    /// In case of a failure, an error is returned, and `errno` is set depending on the error, e.g. EACCES, ENOENT, etc.
+    ///
+    /// ### Errors and Panics
+    ///
+    /// Currently, only two errors are supposrted:
+    /// * `EINVAL` - the value of the mode argument is invalid 
+    /// * `ENOENT` - a component of path does not name an existing file
+    /// Other errors, like `EFAULT  , `ENOTDIR`, etc. are not supported.
+    ///
+    /// There are no cases where this syscall panics.
+    ///
+    /// To learn more about the syscall, valid mode bits, and error values, see
+    /// [chmod(2)](https://man7.org/linux/man-pages/man2/chmod.2.html)
+
     pub fn chmod_syscall(&self, path: &str, mode: u32) -> i32 {
         //Convert the provided pathname into an absolute path without `.` or `..` components.
         let truepath = normpath(convpath(path), self);
@@ -2372,7 +2398,7 @@ impl Cage {
     /// ### Returns
     /// 
     /// Upon successful completion, zero is returned.
-    /// In case of a failure, -1 is returned, and `errno` is set depending on the error, e.g. EACCES, ENOENT, etc.
+    /// In case of a failure, an error is returned, and `errno` is set depending on the error, e.g. EACCES, ENOENT, etc.
     ///
     /// ### Errors and Panics
     ///
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -2261,7 +2261,27 @@ impl Cage {
         }
     }
 
-    //------------------------------------CHMOD HELPER FUNCTION------------------------------------
+    /// ### Description
+    ///
+    /// The `chmod_syscall()` changes a file's mode bits that consist of file permission bits (read, write, execute)
+    /// plus set-user-ID, set-group-ID, and sticky bits.
+    ///
+    /// ### Arguments
+    ///
+    /// The `chmod_syscall()` accepts two arguments:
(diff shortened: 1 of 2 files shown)


JustinCappos:
Let's ping one person at a time on reviews.  Please pick a peer who is also
getting started.  Then get Yash or Rupesh to review after that.  Then me or
Nick for the final pass.  Over time we will promote people up so fewer
reviews are needed...

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,12 +2261,51 @@ impl Cage {
...
+    ///
+    /// ### Arguments
+    ///
+    /// The `chmod_syscall()` accepts two arguments:
+    /// * `path` - pathname of the file whose mode bits we are willing to change (symbolic links are currently 
+    /// not supported). If the pathname is relative, then it is interpreted relative to the current working directory 
+    /// of the calling process.
+    /// * `mode` - the new file mode, which is a bit mask created by bitwise-or'ing zero or more valid mode bits.
+    /// Some of the examples of such bits are `S_IRUSR` (read by owner), `S_IWUSR` (write by owner), 
+    /// `S_ISUID` (set-user-ID), etc.
```
qianxichen233: Do we really have S_ISUID?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,12 +2261,51 @@ impl Cage {
...
     pub fn _chmod_helper(inodenum: usize, mode: u32) {
+        //getting a mutable reference to an inode struct that corresponds to the file whose mode bits we want to change
         let mut thisinode = FS_METADATA.inodetable.get_mut(&inodenum).unwrap();
+        //log is used to store all the changes made to the filesystem. After the cage is closed, all the collected
+        //changes are serialized and the state of the underlying filsystem is persisted. This allows us to avoid
+        //serializing and persisting filesystem state after every chmod_syscall().
         let mut log = true;
+        //This sanity check of the `mode` argument has already been done in `chmod_syscall()` before this helper funciton is called.
+        //Isn't it redundant to check the same thing here again?
         if mode & (S_IRWXA | (S_FILETYPEFLAGS as u32)) == mode {
```
qianxichen233: I don't think chmod allows for modifying the file type directly. And even it allows, I don't think the mode assignment statement is correct, since it does not clear the S_FILETYPEFLAGS bits before making a bitwise or to it, which might result in broken S_FILETYPEFLAGS bits.

src/safeposix/syscalls/fs_calls.rs
ve1nard: this is a general description of the syscall based on the standard. before that part, right in the beginning, I do mention that changing set-user-ID, set-group-ID, and sticky bits is not supported in the current implementation.

src/safeposix/syscalls/fs_calls.rs
ve1nard: There is no mentioning anywhere that chmod is used to change file type. It is used for changing access mode bits. S_FILETYPEFLAGS is only used as a sanity check to make sure that we are working with the right types of files. This is also mentioned in the comments

qianxichen233:
Yeah, but I did not see anywhere you filtered file type bits out, so user
can still pass file type bits to change the file type using chmod

ve1nard ***@***.***>于2024年6月23日 周日21:03写道：

src/safeposix/syscalls/fs_calls.rs
qianxichen233: I actually feel like modifying `if mode & (S_IRWXA | (S_FILETYPEFLAGS as u32)) == mode` to not include S_FILETYPEFLAGS in the first place might be a better choice, but I'll let more professional guys commenting about this
And I guess it might be good to add the test under `ut_lind_fs_chmod_invalid_args` to test for chmod with file type bits passed as argument

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,107 +2261,198 @@ impl Cage {
...
+    ///
+    /// To learn more about the syscall, valid mode bits, and error values, see
+    /// [fchmod(2)](https://linux.die.net/man/2/fchmod)
 
     pub fn fchmod_syscall(&self, fd: i32, mode: u32) -> i32 {
+        //BUG
+        //if the provided file descriptor is out of bounds, 'get_filedescriptor' returns Err(),
+        //unwrapping on which  produces a 'panic!'
+        //otherwise, file descriptor table entry is stored in 'checkedfd'
         let checkedfd = self.get_filedescriptor(fd).unwrap();
```
yashaswi2000, line 2449: Should we return an error or panic in this case?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,107 +2261,198 @@ impl Cage {
...
+            //changes to an inode are saved into the log for all file types except for Sockets
             if log {
                 log_metadata(&FS_METADATA, inodenum)
             };
         }
     }
 
-    //------------------------------------CHMOD SYSCALL------------------------------------
-
     pub fn chmod_syscall(&self, path: &str, mode: u32) -> i32 {
```
yashaswi2000, line 2388: can we add doc string to this function aswell

src/safeposix/syscalls/fs_calls.rs
ve1nard: Looking at the standard now, by design, even if you are not changing the file type, you are not supposed to pass file type bits in the mode argument. It is the task of the underlying syscall to figure out what the file type is and keep it the same after the access mode bits are changed. In other words, mode argument is constructed by or'ing file access mode bits only. I think removing S_FILETYPEFLAGS is a sensible idea.

src/safeposix/syscalls/fs_calls.rs
ve1nard: This is a bug that we have discussed with Nick. Currently, if the descriptor is out of bounds, it will panic. For now, we just marked it as a bug.

src/safeposix/syscalls/fs_calls.rs
ve1nard: can you please elaborate? are you suggesting to add an additional description just like for any other syscall? If yes, the description of this syscall is already there, it is just that it is located above the helper function.

yashaswi2000: ideally it should describe this function and link to helper function

ve1nard: where exactly should I specify that there is a helper function? Should I put it in the description block right after the general description of the syscall?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,107 +2261,187 @@ impl Cage {
...
+                sock_inode.mode = (sock_inode.mode & !S_IRWXA) | mode;
+                //Sockets only exist as long as the cages using them are running. After these cages are closed, no changes
+                //to sockets' inodes need to be persisted, thus using log is unnecessary.
+                log = false;
+            }
+            Inode::Dir(ref mut dir_inode) => {
+                dir_inode.mode = (dir_inode.mode & !S_IRWXA) | mode;
+            }
+        }
+        drop(thisinode);
```
namanlalitnyu: Why are we dropping this inode object in which we updated the `mode` for it?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,107 +2261,188 @@ impl Cage {
...
+                    general_inode.mode = (general_inode.mode & !S_IRWXA) | mode;
                 }
                 Inode::CharDev(ref mut dev_inode) => {
                     dev_inode.mode = (dev_inode.mode & !S_IRWXA) | mode;
                 }
                 Inode::Socket(ref mut sock_inode) => {
                     sock_inode.mode = (sock_inode.mode & !S_IRWXA) | mode;
+                    //Sockets only exist as long as the cages using them are running. After these cages are closed, no changes
+                    //to sockets' inodes need to be persisted, thus using log is unnecessary.
                     log = false;
```
namanlalitnyu, line 2328: If we are saying that no changes to `sockets` need to be persisted, then why are we updating the mode for Socket type inodes?  We can simply ignore the Socket types, as we are not updating the metadata.

ve1nard: "no changes need to be persisted" means that after the cage is closed, we don't need to know anything about its sockets and what happened to them. However, while the cage is still running, the changes need to be updated.

src/safeposix/syscalls/fs_calls.rs
ve1nard: addressed the question in the code comments. copying in this reply:
"the mutable reference to the inode has to be dropped because log_metadata will need to acquire an immutable reference to the same inode"

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,109 +2261,203 @@ impl Cage {
...
                 Inode::Dir(ref mut dir_inode) => {
                     dir_inode.mode = (dir_inode.mode & !S_IRWXA) | mode;
                 }
             }
+            //the mutable reference to the inode has to be dropped because log_metadata will need to acquire an immutable 
+            //reference to the same inode
             drop(thisinode);
+            //changes to an inode are saved into the log for all file types except for Sockets
             if log {
                 log_metadata(&FS_METADATA, inodenum)
```
namanlalitnyu: Ig `;` is missing at the end of this line.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,109 +2261,203 @@ impl Cage {
...
+    /// ### Description
+    ///
+    /// The `_chmod_helper()` is a helper function used by both `chmod_syscall()` and `fchmod_syscall()` 
+    /// to change mode bits that consist of read, write, and execute file permission bits of a file
+    /// specified by an inode obtained from the corresponding caller syscall.
+    ///
+    /// ### Arguments
+    ///
+    /// The `_chmod_helper()` accepts two arguments:
+    /// * `inodenum` - an inode of a file whose mode bits we are willing to change obtained from the caller syscall.
```
namanlalitnyu: We need to wrap this and other comments to be <80 chars wide.

ve1nard: fixed

src/safeposix/syscalls/fs_calls.rs
ve1nard: this brings no practical difference since neither log_metadata() nor _chmod_helper() return any values. nevertheless, added a ';' for completeness

namanlalitnyu:
LGTM!

namanlalitnyu (approved):

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -2261,109 +2261,227 @@ impl Cage {
...
+    /// bitwise-or'ing zero or more valid mode bits. Some of the examples of 
+    /// such bits are `S_IRUSR` (read by owner), `S_IWUSR` (write by owner), etc.
+    ///
+    /// ### Returns
+    /// 
+    /// Upon successful completion, zero is returned.
+    /// In case of a failure, an error is returned, and `errno` is set depending
+    /// on the error, e.g. EACCES, ENOENT, etc.
+    ///
+    /// ### Errors and Panics
```
yashaswi2000: please list these as two seperate sections

ve1nard: fixed

rennergade (approved):
This is a really good PR. All the comments are relevant and informative. 

Approved!

PR #279, opened 2024-06-27
Comments and Fixes for pipe() and the pipe.rs interface

rennergade:
## Description

Fixes # (issue)
Fixes issues with how blocking/nonblocking was handled in the pipe subsystem
Added many comments and a few tests

### Type of change

## How Has This Been Tested?

New tests in ipc_tests.rs

## Checklist:

- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works


src/interface/pipe.rs
```diff
@@ -1,8 +1,14 @@
...
 // Author: Nicholas Renner
-//
-// Pipes for SafePOSIX based on Lock-Free Circular Buffer
 
-#![allow(dead_code)]
+//! In-Memory Pipe Imlpementation for the RustPOSIX interface
+//!
+//! ## Pipe Module
+//!
+//! This module provides a method for in memory iPC between cages and is able to replicate both pipes and Unix domain sockets
```
JustinCappos: Could use about 2-3x more detail.  Are you using a ring buffer of pages or bytes?  How does this differ from normal pipes?  This is UDS also...  how does that plug in and what things are and are not implemented? , etc.

src/interface/pipe.rs
```diff
@@ -14,17 +20,42 @@ use std::slice;
...
 const O_RDONLY: i32 = 0o0;
 const O_WRONLY: i32 = 0o1;
 const O_RDWRFLAGS: i32 = 0o3;
 const PAGE_SIZE: usize = 4096;
 
-const CANCEL_CHECK_INTERVAL: usize = 1048576; // check to cancel pipe reads every 2^20 iterations
-
+// lets also define an interval to check for thread cancellations since we may be blocking in trusted space here
+// we're going to define this as 2^20 which should be ~1 sec according to the standard CLOCKS_PER_SEC definition, though it should be quite shorter on modern CPUs
+const CANCEL_CHECK_INTERVAL: usize = 1048576;
```
JustinCappos: I'm confused about this.  What is this actually for?  Are you busy waiting?  

(Also please line wrap the comments to 80 chars)

src/interface/pipe.rs
```diff
@@ -14,17 +20,42 @@ use std::slice;
...
+
+/// # Description
+/// Helper function to create pipe objects
+///
+/// # Arguments
+///
+/// * `size` - Size of the iPC construct (either pipe or Unix socket)
```
JustinCappos: in bytes?  

Is this interface public to the rest of Lind?

Why even have this function?

src/interface/pipe.rs
```diff
@@ -14,17 +20,42 @@ use std::slice;
...
+/// In-memory pipe struct
+///
+/// # Fields
+///
+/// * `write_end` - Reference to the write end of the pipe protected by a RWLock.
+/// * `read_end` - Reference to the read end of the pipe protected by a RWLock.
+/// * `refcount_write` - Count of open write references.
+/// * `refcount_read` - Count of open read references.
+/// * `eof` - Flag signifying the pipe has finished being written to.
+/// * `size` - Size of pipe buffer in bytes.
```
JustinCappos: These may or may not be useful as Rustdoc comments.  It feels like these are internals which don't matter to the caller.  They just need to know what behavior will happen when calling the code unless you mean to have them manipulate them directly...  (I see that refcount_write and refcount_read are pub, which feels wrong to me...)

src/interface/pipe.rs
```diff
@@ -36,6 +67,17 @@ pub struct EmulatedPipe {
...
+    /// Creates an in-memory pipe object
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - Size of the iPC construct (either pipe or Unix socket)
+    ///
+    /// # Returns
+    ///
+    /// EmulatedPipe object
+    ///
```
JustinCappos: Feels like you either need the one above or this, not both.

Can you end up with ENOMEM or similar if you pass a huge value as the size?

src/interface/pipe.rs
```diff
@@ -49,18 +91,41 @@ impl EmulatedPipe {
         }
     }
 
+    /// # Description
+    /// Setter for EOF flag
```
JustinCappos: Please explain why you only support setting EOF and not clearing it...

src/interface/pipe.rs
```diff
@@ -49,18 +91,41 @@ impl EmulatedPipe {
...
         self.eof.store(true, Ordering::Relaxed);
     }
 
+    /// # Description
+    /// Getter for write references
+    ///
+    /// # Returns
+    ///
+    /// Number of references to write end of the pipe
+    ///
```
JustinCappos: Why do you need this?  In what circumstances?  

(I have similar comments below for many other functions)

src/interface/pipe.rs
```diff
@@ -89,60 +168,127 @@ impl EmulatedPipe {
...
+    /// True if descriptor is ready for writing, false if it will block
+    ///
     pub fn check_select_write(&self) -> bool {
         let write_end = self.write_end.lock();
         let pipe_space = write_end.remaining();
 
-        return pipe_space != 0;
+        // Linux considers a pipe writeable if there is at least PAGE_SIZE (PIPE_BUF) remaining space (4096 bytes)
+        return (self.size - pipe_space) > PAGE_SIZE;
```
JustinCappos: Does this mean there isn't a fully allocated page buffer?  This will not necessarily be exactly the size you mention.

Note also, that I don't know we need to exactly mimic Linux's behavior.  Many of the corner cases just don't matter. in practice.  We can look at what POSIX requires and ensure we're compliant there.

src/interface/pipe.rs
```diff
@@ -89,60 +168,127 @@ impl EmulatedPipe {
...
+    ///
+    /// write_to_pipe accepts three arguments:
+    /// * `ptr` - a pointer to the data being written.
+    /// * `length` - the amount of bytes to attempt to write
+    /// * `nonblocking` - if this attempt to write is nonblocking
+    ///
+    /// ### Returns
+    ///
+    /// Upon successful completion, the amount of bytes written is returned.
+    /// In case of a failure, an error is returned to the calling syscall.
```
JustinCappos, line 195: What errors?  Please list and explain when...

src/interface/pipe.rs
```diff
@@ -89,60 +168,127 @@ impl EmulatedPipe {
...
-                continue;
-            }
-            // we write if the pipe is empty, otherwise we try to limit writes to 4096 bytes (unless whats leftover of this write is < 4096)
-            if remaining != self.size
-                && (length - bytes_written) > PAGE_SIZE
-                && remaining < PAGE_SIZE
-            {
-                continue;
+            // we loop until either more than a page of bytes is free in the pipe
+            // for why we wait for a free page of bytes, refer to the pipe man page about atomicity of writes or the pipe_write kernel implementation
```
JustinCappos: Can you write 2-3 more sentences to explain?  Others (and you from the future) will appreciate not having to reread man pages and / or kernel source code to try to reconstruct what you were thinking.

src/interface/pipe.rs
```diff
@@ -158,6 +304,7 @@ impl EmulatedPipe {
             unsafe {
                 assert!(!ptr.is_null());
                 let iovec = *ptr;
+                // lets conevrt this iovec into a slice and then extend our concatenated buffer
```
JustinCappos: What does this actually mean?

src/interface/pipe.rs
```diff
@@ -168,9 +315,34 @@ impl EmulatedPipe {
...
+    /// * `nonblocking` - if this attempt to read is nonblocking
+    ///
+    /// ### Returns
+    ///
+    /// Upon successful completion, the amount of bytes read is returned.
+    /// In case of a failure, an error is returned to the calling syscall.
+    ///
+    /// ### Errors
+    ///
+    /// * `EAGAIN` - Non-blocking is enabled and there is no data in the pipe.
```
JustinCappos: What if EOF is set?

JustinCappos (changes requested):
A bunch of comments on comments.  Would be happy to discuss more...

src/interface/pipe.rs
```diff
@@ -179,6 +351,9 @@ impl EmulatedPipe {
         let mut read_end = self.read_end.lock();
         let mut pipe_space = read_end.len();
         if nonblocking && (pipe_space == 0) {
```
qianxichen233, line 398: I think read with 0 length should always success even if the pipe space is 0. That's what I've read from Linux pipe implementation, and they checked length==0 at the beginning of both read_from_pipe and write_from_pipe.
And I think add a test testing write and read on 0 length might be good

src/interface/pipe.rs
```diff
@@ -89,60 +168,127 @@ impl EmulatedPipe {
...
+                    }
+                } else {
+                    // we yield here on a non-writable pipe to let other threads continue more quickly
+                    interface::lind_yield();
+                    continue;
+                }
             };
+
+            // lets read the minimum of the specified amount or whatever space we have
             let bytes_to_write = min(length, bytes_written as usize + remaining);
```
qianxichen233, line 282: Just want to point out that Linux is writing a maximum of one page at each iteration, instead of writing all bytes at the same time. Though I think this might not be an issue

src/interface/pipe.rs
rennergade: Yes we have to check for cancellation when we busy wait, this is in read specifically

src/interface/pipe.rs
qianxichen233: Looks like Linux also considers a pipe to be writable if there is no readers

src/interface/pipe.rs
qianxichen233: Linux pipe implementation does not have eof attributes, whenever it wants to see if the pipe is eof, it just checks if there is still any writers. I think Linux's approach might be better since it does not depend on external caller (in this case, close_syscall) to mark the pipe is in end of file state.

src/interface/pipe.rs
rennergade: Really? thats interesting. Can you link me to that?

I guess its for software that is selecting on a closed pipe to error out as quickly as possible.

src/interface/pipe.rs
rennergade: Yeah this makes sense, I'm going to try to factor this out.

src/interface/pipe.rs
qianxichen233: Its just inside the pipe_writable() under pipe.c, I'll paste it here:

```
static inline bool pipe_writable(const struct pipe_inode_info *pipe)
{
	unsigned int head = READ_ONCE(pipe->head);
	unsigned int tail = READ_ONCE(pipe->tail);
	unsigned int max_usage = READ_ONCE(pipe->max_usage);

	return !pipe_full(head, tail, max_usage) ||
		!READ_ONCE(pipe->readers);
}
```

rennergade: thanks!

src/interface/pipe.rs
rennergade: I dont understand what you mean? Don't I list the errors directly below?

src/interface/pipe.rs
rennergade: the manpage explanation is several paragraphs but I'll try to summarize and meet in the middle here

src/interface/pipe.rs
rennergade: Cool. As for ENOMEM, thats not a pipe specific error code, and it only takes up to usize. This constructor is also only called with our constants for pipe and domain socket buf sizes and we don't alllow for adjusting that.

src/interface/pipe.rs
rennergade: I'm not sure I understand your comment but having an empty page in the Linux buffer guarantees there to be at least 4096 bytes? I believe @qianxichen233 mentioned that this is what the posix requriement is.

qianxichen233: This requirements is from Linux man page, I think it is not a POSIX standard. From what I've read: "The Space for PIPE_BUF bytes column indicates whether the pipe has room to atomically write PIPE_BUF bytes without blocking. This is the criterion on which Linux considers a pipe ready for writing. Some other UNIX implementations use the same criterion; others consider a pipe writable if even a single byte can be written."

rennergade:
Thanks guys. I believe I addressed both of your comments if you could take another look at your convenience.

src/tests/ipc_tests.rs
```diff
@@ -408,4 +411,213 @@ pub mod ipc_tests {
...
+    #[test]
+    pub fn ut_lind_ipc_pipe_rw_zero() {
+        //acquiring a lock on TESTMUTEX prevents other tests from running concurrently,
+        // and also performs clean env setup
+        let _thelock = setup::lock_and_init();
+
+        let cage1 = interface::cagetable_getref(1);
+
+        // lets setup the pipe and fork
+        let mut pipefds = PipeArray {
```
JustinCappos, line 589: I'm a bit confused about what this is doing.  In general, I'm against over-commenting tests, but why are readfd and writefd -1?)

src/tests/ipc_tests.rs
```diff
@@ -408,4 +411,213 @@ pub mod ipc_tests {
         assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
         lindrustfinalize();
     }
+
+    #[test]
+    pub fn ut_lind_ipc_pipe2_nonblock() {
```
JustinCappos, line 418: there should be at least one test which does a bunch of common error conditions.  And also a test with a unis domain socket...

JustinCappos (changes requested):
Comments are much better now.  Would like to see a little more testing...

src/tests/ipc_tests.rs
rennergade: this is testing for EAGAIN which is really the only error condition besides EPIPE which we can't test on standalone RustPOSIX because it will trigger a signal.

rennergade: There's also a few Unix tests already, though David can flesh them out further when he gets to send/recv.

src/tests/ipc_tests.rs
rennergade: yeah i see that its unclear, just initializing here, added a comment

src/tests/ipc_tests.rs
rennergade: I guess technically the pipe() syscall can error with ENFILE but thats really more a `get_next_fd()` test than a pipe() test.

qianxichen233 (approved):

rennergade:
@JustinCappos I think this is good to go, can I get a final approval?

JustinCappos (approved):
Thanks for the change.  Loos good to merge...

PR #293, opened 2024-07-08
Updates to "read_syscall" and "pread_syscall".

namanlalitnyu:
## Description

Fixes # (issue)
The following changes include the tests and comments in the code for the "read_syscall" and "pread_syscall" file system calls under RustPosix.
The tests were added to cover all the possible scenarios that might happen when calling the file system_calls  `read_syscall` and `pread_syscall`.

### Type of change
- [x]  This change just contains the tests for an existing file system call.
- [x]  This change contains the minor code changes and comments for read_syscall and pread_syscall.
- [x]  This change contains code reformatting for existing file system calls. 

## How Has This Been Tested?
Inorder to run the tests, we need to run `cargo test --lib` command inside the `safeposix-rust` directory.

All the tests are present under this directory: `lind_project/src/safeposix-rust/src/tests/fs_tests.rs`

- Test A - `ut_lind_fs_read_write_only_fd()`
- Test B - `ut_lind_fs_read_from_directory()`
- Test C - `ut_lind_fs_read_from_epoll()`
- Test D - `ut_lind_fs_read_from_regular_file()`
- Test E - `ut_lind_fs_read_from_chardev_file()`
- Test F -  `ut_lind_fs_read_from_sockets()`
- Test G - `ut_lind_fs_read_from_pipe_blocking_mode()`
- Test H - `ut_lind_fs_read_from_pipe_nonblocking_mode()`
- Test I - `ut_lind_fs_pread_write_only_fd()`
- Test J - `ut_lind_fs_pread_from_file()`
- Test K - `ut_lind_fs_pread_from_directory()`
- Test L - `ut_lind_fs_pread_invalid_types()`

## Checklist:

### DIFF
--- a/src/interface/misc.rs
+++ b/src/interface/misc.rs
@@ -204,6 +204,9 @@ pub fn sigcheck() -> bool {
 }
 
 pub fn fillrandom(bufptr: *mut u8, count: usize) -> i32 {
+    // Potential Bug: The fillrandom function is reading from the /dev/urandom
+    // file, where it should read from "/dev/random" file. And, there should be
+    // a seperate function for reading from "/dev/random" file.
     let slice = unsafe { std::slice::from_raw_parts_mut(bufptr, count) };
     let mut f = std::fs::OpenOptions::new()
         .read(true)
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1493,6 +1493,7 @@ impl Cage {
                         // For `Socket` type inode, a panic is returned as socket type files are not
                         // supported.
                         Inode::Socket(_) => {
+                            // Read is not supported for Socket type inodes.
                             panic!("read(): Socket inode found on a filedesc fd.")
                         }
 
@@ -1741,13 +1742,14 @@ impl Cage {
 
     /// ### Description
     ///
-    /// The `_read_chr_file()` helper function reads from character devices by
-    /// matching the device number (DevNo) of the DeviceInode. It handles
-    /// `/dev/null`, `/dev/zero`, `/dev/random`, and `/dev/urandom` by
-    /// performing the appropriate actions for each device. If the device is
-    /// unsupported, it returns an error indicating that the operation is
-    /// not supported. This function is used for interacting with special
-    /// character files in a Unix-like filesystem.
+    /// The `_read_chr_file()` helper function is used by `read_syscall()` and
+    /// `pread_syscall()` for reading from character device type files.
+    /// It reads from character devices by matching the device number (DevNo)
+    /// of the DeviceInode. It handles `/dev/null`, `/dev/zero`, `/dev/random`,
+    /// and `/dev/urandom` by performing the appropriate actions for each
+    /// device. If the device is unsupported, it returns an error indicating
+    /// that the operation is not supported. This function is used for
+    /// interacting with special character files in a Unix-like filesystem.
     ///
     /// ### Function Arguments
     ///
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1471,14 +1471,11 @@ impl Cage {
                             // object is updated by adding the number of bytes read (bytesread).
                             // This ensures that the next read operation will start from the correct
                             // position.
-                            if let Ok(bytesread) = fileobject.readat(buf, count, position) {
-                                //move position forward by the number of bytes we've read
-                                normalfile_filedesc_obj.position += bytesread;
-                                bytesread as i32
-                            } else {
-                                0 //0 bytes read, but not an error value that
-                                  // can/should be passed to the user
-                            }
+                            let bytesread = fileobject.readat(buf, count, position as usize).unwrap();
+                            //move position forward by the number of bytes we've read
+                            normalfile_filedesc_obj.position += bytesread;
+                            // Return the number of bytes read.
+                            bytesread as i32
                         }
 
                         // For `CharDev` type inode, the reading happens from the Character Device
@@ -1518,6 +1515,12 @@ impl Cage {
                     Errno::EOPNOTSUPP,
                     "read",
                     "reading from stdin not implemented yet",
+                ),
+                // Reading from `Epoll` type file descriptors is not supported.
+                Epoll(_) => syscall_error(
+                    Errno::EINVAL,
+                    "read",
+                    "fd is attached to an object which is unsuitable for reading",
                 ),
                 // The `Pipe` type file descriptor handles read through blocking and non-blocking
                 // modes differently to ensure appropriate behavior based on the flags set on the
@@ -1576,12 +1579,6 @@ impl Cage {
                         return ret;
                     }
                 }
-                // Reading from `Epoll` type file descriptors is not supported.
-                Epoll(_) => syscall_error(
-                    Errno::EINVAL,
-                    "read",
-                    "fd is attached to an object which is unsuitable for reading",
-                ),
             }
         } else {
             syscall_error(Errno::EBADF, "read", "invalid file descriptor")
@@ -1674,13 +1671,11 @@ impl Cage {
                             let fileobject =
                                 FILEOBJECTTABLE.get(&normalfile_filedesc_obj.inode).unwrap();
 
-                            // Attempt to read from the file at the specified offset
-                            if let Ok(bytesread) = fileobject.readat(buf, count, offset as usize) {
-                                bytesread as i32
-                            } else {
-                                0 //0 bytes read, but not an error value that
-                                  // can/should be passed to the user
-                            }
+                            // `readat` function reads from file at specified offset into provided
+                            // C-buffer.
+                            let bytesread = fileobject.readat(buf, count, offset as usize).unwrap();
+                            // Return the number of bytes read.
+                            bytesread as i32
                         }
 
                         // For `CharDev` type inode, the reading happens from the Character Device
@@ -1790,6 +1785,7 @@ impl Cage {
             // reading from /dev/random fills the buffer with random bytes
             RANDOMDEVNO => interface::fillrandom(buf, count),
             // reading from /dev/urandom also fills the buffer with random bytes
+            // Note: This might have to be changed in future.
             URANDOMDEVNO => interface::fillrandom(buf, count),
             // for any device number not specifically handled above,
             // we return an error
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1422,11 +1422,8 @@ impl Cage {
     /// For more detailed description of all the commands and return values, see
     /// [read(2)](https://man7.org/linux/man-pages/man2/read.2.html)
     pub fn read_syscall(&self, fd: i32, buf: *mut u8, count: usize) -> i32 {
-        // Attempt to get the file descriptor; handle error if it does not exist
-        let checkedfd = match self.get_filedescriptor(fd) {
-            Ok(fd) => fd,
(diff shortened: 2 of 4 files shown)


src/safeposix/syscalls/sys_calls.rs
```diff
@@ -43,8 +43,6 @@ use crate::safeposix::filesystem::{decref_dir, metawalk, Inode, FS_METADATA};
 use crate::safeposix::net::NET_METADATA;
 use crate::safeposix::shm::SHM_METADATA;
 
-
-
 impl Cage {
```
namanlalitnyu, line 46: Changes for this file are only the formatting changes done using `cargo fmt`. No new changes have been made in this PR.

namanlalitnyu:
@ve1nard and @Anway-Agte, requesting your review of these changes. 
Thanks!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1409,68 +1481,101 @@ impl Cage {
                             }
```
Anway-Agte: Why is this not an error? does the read_dat function return a None type value only when there is nothing to be read?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1409,68 +1481,101 @@ impl Cage {
...
                         Inode::CharDev(char_inode_obj) => {
+                            // reads from character devices by matching the device number (DevNo) of
+                            // the DeviceInode.
                             self._read_chr_file(&char_inode_obj, buf, count)
                         }
 
+                        // For `Socket` type inode, a panic is returned as socket type files are not
+                        // supported.
                         Inode::Socket(_) => {
                             panic!("read(): Socket inode found on a filedesc fd.")
```
Anway-Agte, line 1493: A line within the panic function saying that read is not supported on sockets would be good here

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1409,68 +1481,101 @@ impl Cage {
                             }
                         }
 
+                        // For `CharDev` type inode, the reading happens from the Character Device
+                        // file, with each device type returning different results returned
+                        // from the `_read_chr_file` file.
                         Inode::CharDev(char_inode_obj) => {
+                            // reads from character devices by matching the device number (DevNo) of
+                            // the DeviceInode.
                             self._read_chr_file(&char_inode_obj, buf, count)
```
Anway-Agte, line 1487: How do we handle the result of the _read_chr_file function?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1518,36 +1682,52 @@ impl Cage {
...
                         Inode::Dir(_) => syscall_error(
                             Errno::EISDIR,
                             "pread",
                             "attempted to read from a directory",
                         ),
                     }
                 }
+                // Return an error for Sockets, as they do not support the concept of seeking to a
+                // specific offset because data arrives in a continuous stream from the network.
                 Socket(_) => syscall_error(
```
Anway-Agte, line 1705: I am a bit confused about the structure of these match statements - we initially get the fd object and check if it points to a FIle, a Socket or a Pipe - but within the File matching branch we again check if the inode points to a socket - are there cases where an fd matches to a type of File but the inode matches to the socket type?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1559,12 +1739,57 @@ impl Cage {
         }
     }
 
+    /// ### Description
+    ///
+    /// The `_read_chr_file()` helper function reads from character devices by
+    /// matching the device number (DevNo) of the DeviceInode. It handles
+    /// `/dev/null`, `/dev/zero`, `/dev/random`, and `/dev/urandom` by
```
Anway-Agte: What do these 4 cases signify?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1559,12 +1739,57 @@ impl Cage {
         }
     }
 
+    /// ### Description
+    ///
```
Anway-Agte, line 1739: A line stating that this function is used by the read_syscall would be great

Anway-Agte (changes requested):

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: These cases represent 4 different types of character devices that are currently supported by the system.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Yes, the initial match block checks for the different types of File Descriptors (Sockets, Streams, File, Pipe, Epoll), and in the inside block of File type FileDescriptor, we check for different types of inodes(File, CharDev, Directory, Sockets).  
So, both of them are different here.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Here, this function returns the number of bytes read into the buffer. And since for returning values in Rust, we just specify the return value, here we are just calling the function which returns the expected result. 
Will update this comment for better understanding.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: I referenced the man-pages for `read_syscall` and they have mentioned that if read is not successful, then we return 0 as the default value in the response, representing no bytes were read from the file. 
https://man7.org/linux/man-pages/man2/read.2.html

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: fixed.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: fixed.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Updated the comments with the description of the return value.

ve1nard (approved):
lgtm!

Anway-Agte (approved):

namanlalitnyu:
@yashaswi2000, Please can you review these changes.
Thanks!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1409,68 +1481,102 @@ impl Cage {
...
+                            // Received `EAGAIN` and no cancellation is requested, continue the loop
+                            // to try reading from the pipe again
+                            continue;
                         }
-                        return ret; // if we get here we can return
+                        // If the read was successful, return the result
+                        return ret;
                     }
                 }
+                // Reading from `Epoll` type file descriptors is not supported.
```
yashaswi2000: can we change the order to fail quick

src/safeposix/syscalls/fs_calls.rs
yashaswi2000: is that the EOF condition or if any failure for read_at should return 0? shouldnt we be doing -1 if it failed due to unknown reason?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1559,12 +1741,58 @@ impl Cage {
...
         match inodeobj.dev {
-            NULLDEVNO => 0, //reading from /dev/null always reads 0 bytes
+            // reading from /dev/null always reads 0 bytes indicating an end-of-file condition.
+            NULLDEVNO => 0,
+            // reading from /dev/zero fills the buffer with zeroes
             ZERODEVNO => interface::fillzero(buf, count),
+            // reading from /dev/random fills the buffer with random bytes
             RANDOMDEVNO => interface::fillrandom(buf, count),
+            // reading from /dev/urandom also fills the buffer with random bytes
             URANDOMDEVNO => interface::fillrandom(buf, count),
```
yashaswi2000, line 1788: so there is no difference between random and urandom?

yashaswi2000 (commented):
looks almost done. great job with the tests!

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Updated the description and original function with comments about a possible bug.

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Updated both the code references as per the discussion.

src/safeposix/syscalls/fs_calls.rs
rennergade: I think for a match statement this shouldn't matter? I may be wrong.

namanlalitnyu: Yes, I think the same but its better to have things which fail together, so that the code looks more readable. Have updated it though in the latest commit.

yashaswi2000: > Think of a match expression as being like a coin-sorting machine: coins slide down a track with variously sized holes along it, and each coin falls through the first hole it encounters that it fits into. In the same way, values go through each pattern in a match, and at the first pattern the value “fits,” the value falls into the associated code block to be used during execution.
- Rust book - [link](https://doc.rust-lang.org/book/ch06-02-match.html)

just mentioning here for clarity over the match workings

yashaswi2000 (approved):
LGTM!

src/interface/misc.rs
```diff
@@ -204,6 +204,9 @@ pub fn sigcheck() -> bool {
 }
 
 pub fn fillrandom(bufptr: *mut u8, count: usize) -> i32 {
```
rennergade, line 206: I think this comment should clarify that for now we don't care about the blocking/entropy aspect of /dev/random

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1385,111 +1446,209 @@ impl Cage {
...
+                        // from the `_read_chr_file` file.
                         Inode::CharDev(char_inode_obj) => {
+                            // reads from character devices by matching the device number (DevNo) of
+                            // the DeviceInode.
                             self._read_chr_file(&char_inode_obj, buf, count)
                         }
 
+                        // For `Socket` type inode, a panic is returned as socket type files are not
+                        // supported.
                         Inode::Socket(_) => {
```
rennergade, line 1492: This really isn't "unsupported". It's actually failing a sanity check. a File type fd should never have a socket type inode

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1498,56 +1657,74 @@ impl Cage {
...
+                        // file, with each device type returning different results returned
+                        // from the `_read_chr_file` file.
                         Inode::CharDev(char_inode_obj) => {
+                            // reads from character devices by matching the device number (DevNo) of
+                            // the DeviceInode. This function returns the number of bytes read from
+                            // the character device and updates the buffer `buf` with them.
                             self._read_chr_file(&char_inode_obj, buf, count)
                         }
-
+                        // For `Socket` type inode, a panic is returned as socket type files are not
```
rennergade: same as above

rennergade (changes requested):
Some minor changes requested. Looks good!

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: Fixed!

src/safeposix/syscalls/fs_calls.rs
namanlalitnyu: updated

rennergade (approved):
Approved.

PR #298, opened 2024-07-10
Listen sys call comments + tests

davidge20:
## Description

Added comments to listen_syscall and added two tests

Fixes # (issue)

 

Two issues that I would like to share:
1. While calling the listen sys call, the backlog is set to 5 in the function, so the test queues more than 5 connections. The 6th connection should return with an error, given the queue is maximized. However, the 6th queue returns with EINPROGRESS.
2. We create a blocking socket to connect to a "server" and it returns immediately. The blocking socket should not return with success immediately as the server did not accept the connection.

### Type of change

## How Has This Been Tested?

- Test A - `ut_lind_net_listen_more_than_backlog()`
- Test B - `ut_lind_net_listen_nonblock_then_block()`

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -1518,8 +1518,8 @@ impl Cage {
     /// 
     /// it accepts two parameters: 
     /// * `sockfd` - a file descriptor that refers to a socket
-    ///    of type SOCK_STREAM or SOCK_SEQPACKET.
-    /// // ** Do we deal with SOCK_SEQPACKET ?? ** //
+    ///    of type SOCK_STREAM 
+    ///    Note, we do not implement sockets of type SOCK_SEQPACKET
     /// * `backlog` - defines the maximum length to which the
     ///   queue of pending connections for sockfd may grow.  If a
     ///   connection request arrives when the queue is full, the client may
@@ -1562,6 +1562,7 @@ impl Cage {
     //
     // TODO: We are currently ignoring backlog 
     pub fn listen_syscall(&self, fd: i32, _backlog: i32) -> i32 {
+        //BUG:s
         //If fd is out of range of [0,MAXFD], process will panic
         //Otherwise, we obtain a write gaurd to the Option<FileDescriptor> object
         let checkedfd = self.get_filedescriptor(fd).unwrap();
@@ -1582,9 +1583,16 @@ impl Cage {
                             return 0;
                         }
 
-                        //If the given socket is connected to another socket or 
-                        //if a non blocking socket is in progress of connecting
-                        //to another socket
+                        //Possible connection states in which the socket
+                        //can not be set to listening mode:
+                        // * Connected to another socket and can send
+                        // and receive data
+                        // * Connected to another socket and can only send
+                        // data
+                        // * Connected to another socket and can only receive
+                        // data
+                        // * A non-blocking socket is in progress of connecting
+                        // to another socket
                         ConnState::CONNECTED
                         | ConnState::CONNRDONLY
                         | ConnState::CONNWRONLY
@@ -1618,6 +1626,13 @@ impl Cage {
 
                             //If the given socket is not assigned an address,
                             //attempt to bind the socket to an address.
+                            //
+                            //An implicit bind refers to the automatic binding of a socket to an address
+                            //and port by the system, without an explicit call to the bind() function by
+                            //the programmer.
+                            //
+                            //If implicit bind fails, return with the errno if known
+                            //Otherwise, panic!
                             if sockhandle.localaddr.is_none() {
                                 let shd = sockhandle.domain as i32;
                                 let ibindret = self._implicit_bind(&mut *sockhandle, shd);
@@ -1677,9 +1692,11 @@ impl Cage {
                                 return lr;
                             };
 
-                            //set rawfd for select sys call
-                            //** Why is this being done in listen ?? 
-                            // We are also doing it in connect right now **/
+                            // Set the rawfd for select_syscall as we cannot implement the select 
+                            // logics for AF_INET socket right now, so we have to call the select 
+                            // syscall from libc, which takes the rawfd as the argument instead of 
+                            // the fake fd used by lind.
+                            // The raw fd of the socket is the set to be the same as the fd set by the kernal in the libc connect call
                             sockfdobj.rawfd = sockhandle.innersocket.as_ref().unwrap().raw_sys_fd;
 
                             //If listening socket is not in the table of pending 
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -152,12 +152,60 @@ impl Cage {
         0
     }
 
+    /// ## `socket_syscall`
+    ///
+    /// ### Description
+    /// This function creates a new socket, ensuring the requested domain,
+    /// socket type, and protocol are supported by SafePosix.
+    /// It validates the requested communication domain, socket type, and
+    /// protocol, permitting only combinations that are known to be safe and
+    /// secure.
+    ///
+    /// ### Function Arguments
+    /// * `domain`: The communication domain for the socket. Supported values
+    ///   are `PF_INET` (Internet Protocol) and `PF_UNIX` (Unix domain sockets).
+    /// * `socktype`: The socket type. Supported values are `SOCK_STREAM`
+    ///   (stream sockets) and `SOCK_DGRAM` (datagram sockets).
+    /// * `protocol`: The protocol to use for communication. This defaults to
+    ///   TCP for stream sockets (`SOCK_STREAM`) and UDP for datagram sockets
+    ///   (`SOCK_DGRAM`).
+    ///
+    /// ### Returns
+    /// * The new file descriptor representing the socket on success.
+    ///
+    /// ### Errors
+    /// * `EOPNOTSUPP(95)`: If an unsupported combination of domain, socket
+    ///   type, or protocol is requested.
+    /// * `EINVAL(22)`: If an invalid combination of flags is provided.
+    /// ### Panics
+    /// There are no panics in this syscall.
     pub fn socket_syscall(&self, domain: i32, socktype: i32, protocol: i32) -> i32 {
         let real_socktype = socktype & 0x7; //get the type without the extra flags, it's stored in the last 3 bits
-        let nonblocking = (socktype & SOCK_NONBLOCK) != 0;
+        let nonblocking = (socktype & SOCK_NONBLOCK) != 0; // Checks if the socket should be non-blocking.
+                                                           //Check blocking status for storage in the file descriptor, we'll need this for
+                                                           // calls that don't access the kernel
+                                                           // socket, unix sockets, and properly directing kernel calls for recv and accept
         let cloexec = (socktype & SOCK_CLOEXEC) != 0;
+        // Checks if the 'close-on-exec' flag is set. This flag ensures the socket is
+        // automatically closed if the current process executes another program,
+        // preventing unintended inheritance of the socket by the new program.
 
+        // additional flags are not supported
+        // filtering out any socktypes with unexpected flags set.
+        // This is important as we dont want to pass down any flags that are not
+        // supported by SafePOSIX. which may potentially cause issues with the
+        // underlying libc call. or the socket creation process. leading to
+        // unexpected behavior.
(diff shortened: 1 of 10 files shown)


rustc-ice-2024-07-10T16_06_47-19579.txt
```diff
@@ -0,0 +1,36 @@
...
+                               at /rustc/6be96e3865c4e59028fd50396f7a46c3498ce91d/library/alloc/src/boxed.rs:2064:9
+  17:     0xffff7f2971e8 - std::sys::pal::unix::thread::Thread::new::thread_start::h9b0657d0725f835d
+                               at /rustc/6be96e3865c4e59028fd50396f7a46c3498ce91d/library/std/src/sys/pal/unix/thread.rs:108:17
+  18:     0xffff7f0d597c - <unknown>
+  19:     0xffff7f13ba4c - <unknown>
+  20:                0x0 - <unknown>
+
+
+rustc version: 1.81.0-nightly (6be96e386 2024-07-09)
+platform: aarch64-unknown-linux-gnu
```
qianxichen233: could you remove these redundent files? You can either uncommit this commit with `git reset --soft HEAD^` and commit again with these files removed, or just create a new commit with these files removed

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1510,22 +1510,81 @@ impl Cage {
...
+    /// ### Description
+    /// 
+    /// `listen_syscall` listen for connections on a socket
+    /// 
+    /// ### Arguments
+    /// 
+    /// it accepts two parameters: 
+    /// * `sockfd` - a file descriptor that refers to a socket
+    ///    of type SOCK_STREAM or SOCK_SEQPACKET.
+    /// // ** Do we deal with SOCK_SEQPACKET ?? ** //
```
qianxichen233: I don't think we support SOCK_SEQPACKET. You can mention this somewhere in the comment

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1510,22 +1510,81 @@ impl Cage {
...
+    ///   that the socket isn't assigned an address
+    /// * unknown errno value from socket listen sys call from libc
+    /// 
+    /// for more detailed description of all the commands and return values, see 
+    /// [listen(2)](https://linux.die.net/man/2/listen)
+    //
+    // TODO: We are currently ignoring backlog 
     pub fn listen_syscall(&self, fd: i32, _backlog: i32) -> i32 {
+        //If fd is out of range of [0,MAXFD], process will panic
+        //Otherwise, we obtain a write gaurd to the Option<FileDescriptor> object
```
qianxichen233, line 1885: I think you should either mark a BUG here, or just fix it (return EBADF if fd is out of range)

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1510,22 +1510,81 @@ impl Cage {
...
+                    //success
                     match sockhandle.state {
                         ConnState::LISTEN => {
-                            return 0; //Already done!
+                            return 0;
                         }
 
+                        //If the given socket is connected to another socket or 
+                        //if a non blocking socket is in progress of connecting
+                        //to another socket
```
qianxichen233: I think you can add more comments explaining each state below. Like what each state is, and what does that state mean to a socket

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1563,18 +1629,27 @@ impl Cage {
...
+                            //to readily accept connections
                             NET_METADATA.listening_port_set.insert(porttuple.clone());
                             sockhandle.state = ConnState::LISTEN;
 
-                            let listenret = sockhandle.innersocket.as_ref().unwrap().listen(5); //default backlog in repy for whatever reason, we replicate it
+                            //Call listen from libc on the socket
+                            //Set the backlog to 5:
+                            //default backlog in repy for whatever reason, we replicate it
+                            //** Would we ever want to change the backlog??  **/
+                            let listenret = sockhandle.innersocket.as_ref().unwrap().listen(5);
```
qianxichen233: But we do have a backlog argument passed into listen_syscall, why not just use it directly?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1586,30 +1661,44 @@ impl Cage {
...
                                 ));
+                                //Set the socket state to NOTCONNECTED, as
+                                //the socket is not listening
                                 sockhandle.state = ConnState::NOTCONNECTED;
                                 return lr;
                             };
 
-                            //set rawfd for select
+                            //set rawfd for select sys call
+                            //** Why is this being done in listen ??
```
qianxichen233: I guess since there is a code logic before that does something like "if the socket didn't bind to any address, then assign a new address". So the innersocket could be possibly a new one that just created in this syscall, and we certainly want to update its rawfd as well

src/tests/networking_tests.rs
```diff
@@ -2642,6 +2783,7 @@ pub mod net_tests {
     }
 
     #[test]
+    #[ignore]
```
qianxichen233: why this is marked as ignore?

src/tests/networking_tests.rs
```diff
@@ -669,6 +670,146 @@ pub mod net_tests {
...
+        assert_eq!(cage.exit_syscall(EXIT_SUCCESS), EXIT_SUCCESS);
+        lindrustfinalize();
+    }
+
+    //We create a blocking socket to connect to the server and it returns immediately
+    //The blocking socket should not return with success immediately as the server
+    //did not accept the connection
+    #[test]
+    #[ignore]
+    pub fn ut_lind_net_listen_nonblock_then_block() {
```
qianxichen233: I don't quite understand this test, what's the issue here? Like the connect is supposed to block forever since server is not accepting the socket, but it actually returned immediately?

rustc-ice-2024-07-10T16_02_42-17210.txt
```diff
@@ -0,0 +1,36 @@
+thread 'coordinator' panicked at /rustc/6be96e3865c4e59028fd50396f7a46c3498ce91d/compiler/rustc_codegen_ssa/src/back/write.rs:1676:29:
+/rustc/6be96e3865c4e59028fd50396f7a46c3498ce91d/compiler/rustc_codegen_ssa/src/back/write.rs:1676:29: worker thread panicked
```
namanlalitnyu, line 2: I think we should remove these `rustc-ice*` files from these changes.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1546,12 +1609,15 @@ impl Cage {
...
                             if sockhandle.domain == AF_UNIX {
                                 sockhandle.state = ConnState::LISTEN;
                                 return 0;
                             }
 
+                            //If the given socket is not assigned an address,
+                            //attempt to bind the socket to an address.
                             if sockhandle.localaddr.is_none() {
                                 let shd = sockhandle.domain as i32;
                                 let ibindret = self._implicit_bind(&mut *sockhandle, shd);
```
namanlalitnyu, line 1958: Can you explain a bit about this function through a couple of comments for better understanding.

rustc-ice-2024-07-10T16_06_47-19579.txt
davidge20: Perfect, thanks!

Had to use `git reset --mixed HEAD^` so the changes arent commited again automatically and need to be added with git add.

src/safeposix/syscalls/net_calls.rs
davidge20: Good question. @rennergade why was the backlog set to 5 and not the number passed in to listen?

src/tests/networking_tests.rs
davidge20: The test fails given it's not giving the behavior we expect. Right now if one test fails all of the following tests fail. Once I think yash fixes that we don't need to ignore it anymore.

src/tests/networking_tests.rs
davidge20: Aren't INET sockets are blocking by default? Hence, as you mentioned, shouldn't the connect sys call block forever since the server is not accepting the socket? However, it returns with success immediately.

src/tests/networking_tests.rs
namanlalitnyu: Can you try it now with the latest rebase, looks like Yash's Changes are in related to this.

src/tests/networking_tests.rs
qianxichen233: From what I've read, connect is actually not going to block if server already listened but hasn't accept the connection yet. `connect` can return normally if it goes into the server's listening queue successfully.
btw another suggestion is that it might not be a good idea to add the test that would expect `blocking forever` behavior

davidge20: Yeah thats a good point about adding a test that expects blocking forever behavior. 

Based on the following: https://stackoverflow.com/questions/8258107/does-connect-block-for-tcp-socket , 
so even if the TCP socket is in blocking mode, connect will return with success by being put on the server's listening queue cause the OS takes care of that? 

Then whats the point of blocking mode?

qianxichen233: nonblocking mode could affect many other functions like accept, send, recv, etc. And in case of connect, there is still some thing get affected. For example, when you try to connect in nonblocking mode, if the server cannot respond immediately (note that this is different from server refuse to respond, in which case the server would explictly send a message back (TCP RST specifically I think)), connect could block until it receives something from server. But in nonblocking mode, it will return after a few milliseconds (must wait some time to determine if server can respond because of network delay) if the server is not responding with INPROGRESS state. Later when the connection is either denied or accept, you can check it using getsockopt syscall. I think the blocking/nonblocking behavior on `connect` syscall, especially on INET is very complex, mostly because you need to balance network delay and nonblocking delay, and it may involve a lot of TCP stuff

qianxichen233 (approved):

davidge20:
@rupeshkoushik07  requesting your review, thanks!

namanlalitnyu (approved):
LGTM!

rupeshkoushik07 (approved):
LGTM !

src/tests/networking_tests.rs
```diff
@@ -2761,6 +2857,7 @@ pub mod net_tests {
     }
 
     #[test]
+    #[ignore]
```
yashaswi2000: whats the current behaviour?

src/safeposix/syscalls/net_calls.rs
yashaswi2000: bumping this up @rennergade

yashaswi2000 (approved):
LGTM

src/safeposix/syscalls/net_calls.rs
rennergade: Yeah i dont see any reason to not use the passed parameter. Lets make that change. Just a weird legacy thing.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1865,12 +1935,23 @@ impl Cage {
                                 );
                             }
 
-                            // simple if it's a domain socket
+                            //If the given socket is a Unix socket, lind handles
+                            //the connection, return with success
                             if sockhandle.domain == AF_UNIX {
```
rennergade, line 1941: I think if were going to actually add backlog in for INET its worth putting a TODO tag here saying implement backlog for unix

src/tests/networking_tests.rs
rennergade: yeah this one definitely shouldnt fail so we need to look into it

src/tests/networking_tests.rs
```diff
@@ -2,6 +2,7 @@
 pub mod net_tests {
     use super::super::*;
     use crate::interface;
+    use crate::interface::errnos::{syscall_error, Errno};
```
rennergade: why did you need to import this?

rennergade (changes requested):
This is very close and quite thorough. Good job!

A few minor changes, lets add the backlog and make sure that test is working.

src/tests/networking_tests.rs
davidge20: some mistake on my end, removed the ignore statement

src/tests/networking_tests.rs
davidge20: Wasn't necessary, i figured out how to use errno

removed the statement

rennergade (approved):

PR #301, opened 2024-07-12
Description, Comments and tests for `stat_syscall` & `fstat_syscall`

pranav-bhatt:
## Description

This PR adds comments and tests for the `stat_syscall` & `fstat_syscall` system call function.

### Type of change

## How Has This Been Tested?

For the unit tests added, run ``

- Test A - `safeposix-rust/tests/fs_tests.rs - ut_lind_fs_stat_syscall_tests()`,
- Test B - `safeposix-rust/tests/fs_tests.rs - ut_lind_fs_fstat_syscall_tests()`

## Checklist:

### DIFF
--- a/src/interface/types.rs
+++ b/src/interface/types.rs
@@ -30,7 +30,7 @@ pub struct FSData {
 //redefining the StatData struct in this file so that we maintain flow of
 // program derive eq attributes for testing whether the structs equal other
 // statdata structs from stat/fstat
-#[derive(Eq, PartialEq, Default)]
+#[derive(Eq, PartialEq, Default, Debug)]
 #[repr(C)]
 pub struct StatData {
     pub st_dev: u64,
--- a/src/interface/types.rs
+++ b/src/interface/types.rs
@@ -30,7 +30,7 @@ pub struct FSData {
 //redefining the StatData struct in this file so that we maintain flow of
 // program derive eq attributes for testing whether the structs equal other
 // statdata structs from stat/fstat
-#[derive(Eq, PartialEq, Default, Debug)]
+#[derive(Eq, PartialEq, Default)]
 #[repr(C)]
 pub struct StatData {
     pub st_dev: u64,
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1317,6 +1317,7 @@ impl Cage {
     ///
     /// ### Panics
     ///
+    /// * If the file descriptor passed
     /// * If the inode number retrieved from the file descriptor does not exist
     ///   in `FS_METADATA.inodetable`.
     ///
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1325,7 +1325,14 @@ impl Cage {
 
     pub fn fstat_syscall(&self, fd: i32, statbuf: &mut StatData) -> i32 {
         // Attempt to get the file descriptor
-        let checkedfd = self.get_filedescriptor(fd).unwrap();
+        // BUG: This line results in a panic if the file descriptor is invalid. We
+        // should gracefully be handling this error.
+        let checkedfd_result = self.get_filedescriptor(fd);
+        if let Err(_) = checkedfd_result {
+            return syscall_error(Errno::EBADF, "fstat", "Bad File Descriptor");
+        }
+        let checkedfd = checkedfd_result.unwrap();
+
         // Acquire a write lock on the file descriptor to ensure exclusive access.
         let unlocked_fd = checkedfd.read();
 
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1325,12 +1325,13 @@ impl Cage {
 
     pub fn fstat_syscall(&self, fd: i32, statbuf: &mut StatData) -> i32 {
         // Attempt to get the file descriptor
-        // BUG: This line results in a panic if the file descriptor is invalid. We
-        // should gracefully be handling this error.
         let checkedfd_result = self.get_filedescriptor(fd);
+
+        // If the file descriptor is invalid, return an error of code 9 (bad file num).
         if let Err(_) = checkedfd_result {
             return syscall_error(Errno::EBADF, "fstat", "Bad File Descriptor");
         }
+        // If the file descriptor is valid, we get the file descriptor object
         let checkedfd = checkedfd_result.unwrap();
 
         // Acquire a write lock on the file descriptor to ensure exclusive access.
@@ -5307,40 +5308,40 @@ impl Cage {
     }
 
     /// ### Description
-    /// 
+    ///
     /// `shmget_syscall` returns the shared memory segment identifier associated with a particular `key`
     /// If a key doesn't exist, shmget creates a new memory segment and attaches it to the key.
-    /// Traditionally if the value of the key equals `IPC_PRIVATE`, we also create a new memory segment which 
-    /// is not associated with a key during this syscall, 
-    /// but for our implementaion, we return an error and only create a new memory 
+    /// Traditionally if the value of the key equals `IPC_PRIVATE`, we also create a new memory segment which
+    /// is not associated with a key during this syscall,
+    /// but for our implementaion, we return an error and only create a new memory
     /// segment when the IPC_CREAT flag is specified in the`shmflag` argument.
-    /// 
-    /// ### Returns 
-    /// 
+    ///
+    /// ### Returns
+    ///
     /// An 32 bit integer which represens the identifier of the memory segment associated with the key
-    /// 
+    ///
     /// ### Arguments
-    /// 
+    ///
     /// `key` : An i32 value that references a memory segment
     /// `size` : Size of the memory segment to be created if key doesn't exist
     /// `shmflag` : mode flags which indicate whether to create a new key or not
-    ///  The `shmflag` is composed of the following 
+    ///  The `shmflag` is composed of the following
     ///  * IPC_CREAT - specify that the system call creates a new segment
-    ///  * IPC_EXCL - this flag is used with IPC_CREAT to cause this function to fail when IPC_CREAT is also used 
+    ///  * IPC_EXCL - this flag is used with IPC_CREAT to cause this function to fail when IPC_CREAT is also used
     ///               and the key passed has a memory segment associated with it.
-    /// 
-    /// ### Errors 
-    /// 
+    ///
+    /// ### Errors
+    ///
     /// * ENOENT : the key equals the `IPC_PRIVATE` constant
     /// * EEXIST : key exists and yet either `IPC_CREAT` or `IPC_EXCL` are passed as flags
     /// * ENOENT : key did not exist and the `IPC_CREAT` flag was not passed
     /// * EINVAL : the size passed was less than the minimum size of segment or greater than the maximum possible size
-    /// 
+    ///
     /// ### Panics
-    /// 
+    ///
     /// There are no cases where the function directly panics
-    /// 
-    pub fn shmget_syscall(&self, key: i32, size: usize, shmflg: i32) -> i32 { 
+    ///
+    pub fn shmget_syscall(&self, key: i32, size: usize, shmflg: i32) -> i32 {
         //Check if the key passed equals the IPC_PRIVATE flag
         if key == IPC_PRIVATE {
             // Return error since this is not suppported currently
@@ -5392,7 +5393,7 @@ impl Cage {
                 // Insert new id in the hash table entry pointed by the key
                 vacant.insert(shmid);
                 // Mode of the new segment is the 9 least significant bits of the shmflag
-                let mode = (shmflg & 0x1FF) as u16; 
+                let mode = (shmflg & 0x1FF) as u16;
                 // Create a new segment with the key, size, cageid of the calling process
                 let segment = new_shm_segment(
                     key,
@@ -5407,7 +5408,7 @@ impl Cage {
             }
         };
         // Return the shmid
-        shmid 
+        shmid
     }
 
     //------------------SHMAT SYSCALL------------------
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1325,14 +1325,8 @@ impl Cage {
 
     pub fn fstat_syscall(&self, fd: i32, statbuf: &mut StatData) -> i32 {
         // Attempt to get the file descriptor
-        let checkedfd_result = self.get_filedescriptor(fd);
-
-        // If the file descriptor is invalid, return an error of code 9 (bad file num).
-        if let Err(_) = checkedfd_result {
-            return syscall_error(Errno::EBADF, "fstat", "Bad File Descriptor");
-        }
-        // If the file descriptor is valid, we get the file descriptor object
-        let checkedfd = checkedfd_result.unwrap();
+        // BUG: This can panic if there is an invalid file descriptor provided
+        let checkedfd = self.get_filedescriptor(fd).unwrap();
 
         // Acquire a write lock on the file descriptor to ensure exclusive access.
         let unlocked_fd = checkedfd.read();
--- a/src/safeposix/syscalls/fs_calls.rs
+++ b/src/safeposix/syscalls/fs_calls.rs
@@ -1338,6 +1338,16 @@ impl Cage {
             // handle regular files, dirs, and char files based on the
             // information in the inode.
             match filedesc_enum {
+                // Fail faster for sockets
+                Socket(_) => {
+                    return syscall_error(
+                        Errno::EOPNOTSUPP,
+                        "fstat",
+                        "we don't support fstat on sockets yet",
+                    );
+                }
+
+                // if a normal file descriptor is found
                 File(normalfile_filedesc_obj) => {
                     // fetch the inode object of the normal file
                     let inode = FS_METADATA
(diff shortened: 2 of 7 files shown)


src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1148,12 +1148,46 @@ impl Cage {
     }
 
     //------------------------------------STAT SYSCALL------------------------------------
+    /// ### Description
+    ///
+    /// `stat_syscall` retrieves file information for the file specified by
+    /// `path` and populates the provided `statbuf` with this information.
```
yashaswi2000, line 1154: can add whether we need permissions on the file to do this syscall or not etc

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1193,6 +1229,8 @@ impl Cage {
         statbuf.st_blocks = 0;
     }
 
+    // helper function to populate information of socket inode object into the
+    // statbuf
     fn _istat_helper_sock(inodeobj: &SocketInode, statbuf: &mut StatData) {
```
yashaswi2000, line 1237: document the helper function aswell

yashaswi2000 (commented):
Good job with the description, add doc strings to helper functions as well, missing unit tests.

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: So I have a couple of thoughts here:
1. The helper functions themselves are pretty straightforward since they are just populating fields
2. the definitions of each of those fields and the possible values they can take are pretty long and can lead to over-documentation
3. Adding the definitions of each field is probably out of the scope of the helper functions, and could maybe (but again probably not) be handled in the object definition of the inodes itself instead of here.

A better way here would be to link the man page for inodes. That should provide a better experience for folks and keep things clean IMO.

yashaswi2000: that sounds good.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1148,12 +1148,49 @@ impl Cage {
...
     pub fn stat_syscall(&self, path: &str, statbuf: &mut StatData) -> i32 {
+        //convert the path to an absolute path of type `PathBuf`
         let truepath = normpath(convpath(path), self);
 
         //Walk the file tree to get inode from path
         if let Some(inodenum) = metawalk(truepath.as_path()) {
+            // can panic if inode does not exist in inode table
             let inodeobj = FS_METADATA.inodetable.get(&inodenum).unwrap();
 
             //populate those fields in statbuf which depend on things other than the inode
```
Anway-Agte, line 1194: I line explaining what istat_helper does would be good for reference

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1148,12 +1148,49 @@ impl Cage {
...
+    /// For more detailed description of all the commands and return values,
+    /// refer to the stat man page [here](https://man7.org/linux/man-pages/man2/stat.2.html).
 
     pub fn stat_syscall(&self, path: &str, statbuf: &mut StatData) -> i32 {
+        //convert the path to an absolute path of type `PathBuf`
         let truepath = normpath(convpath(path), self);
 
         //Walk the file tree to get inode from path
         if let Some(inodenum) = metawalk(truepath.as_path()) {
+            // can panic if inode does not exist in inode table
```
ve1nard: I believe the case of nonexistent entry (which would have led to a panic) is checked in the conditional statement above. In other words, if there is no entry, metawalk would return None, and you won't get to the unwrap part.

ve1nard (commented):
a minor comment

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: nice catch! kinda redundant that we are checking the table 2 times though, but it does improve code readability

ve1nard (approved):
lgtm!

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1148,12 +1148,48 @@ impl Cage {
...
+    /// (search) permission is required on all of the directories in
+    /// pathname that lead to the file.
+    ///
+    /// ### Arguments
+    ///
+    /// It accepts two parameters:
+    /// * `path` - A string slice that specifies the file path for which status
+    ///   information is to be retrieved.
+    /// * `statbuf` - A mutable reference to a `StatData` struct where the file
+    ///   status will be stored.
```
Anway-Agte, line 1165: Does the status have a fixed range of values that it can take ? if so - we could list them here to know what to expect from this function

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1148,12 +1148,48 @@ impl Cage {
...
+    /// errno is returned to indicate the error.
+    ///
+    /// ### Errors
+    ///
+    /// * `ENOENT` - The file specified by `path` does not exist or the path is
+    ///   invalid.
+    ///
+    /// ### Panics
+    ///
+    /// * If `statbuf` is incorrectly populated, causing potential runtime
```
Anway-Agte: what causes it to be incorrectly populated?

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1242,18 +1290,55 @@ impl Cage {
     }
 
     //------------------------------------FSTAT SYSCALL------------------------------------
+    /// ### Description
```
Anway-Agte, line 1292: A line stating that the function has the same functionality as fstat - the only difference is the file indicator (i.e path if stat and fd ion fstat) would be good I think

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1286,14 +1372,17 @@ impl Cage {
                         "we don't support fstat on sockets yet",
                     );
                 }
+                // Streams don't have inodes, so we'll populate statbuf with dummy info
```
Anway-Agte, line 1379: a line explaining what the dummy info is would be good for clarification

src/tests/fs_tests.rs
```diff
@@ -4036,4 +4037,136 @@ pub mod fs_tests {
...
+
+        let cage = interface::cagetable_getref(1);
+
+        let mut statdata = StatData::default();
+
+        // test out whether an error is output for a non existent file path
+        // (ENOENT[-2])
+        assert_eq!(
+            cage.stat_syscall("non_existent_file_path", &mut statdata),
+            syscall_error(Errno::ENOENT, "stat", "test_failure")
```
Anway-Agte, line 4425: I think only checking with -(Errno::ENOENT as i32) suffices for checking errors - we don't need to call the syscall_error function explicitly

Anway-Agte (changes requested):

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: istat helper already has a function description, and adding it as well would be imo redundant

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: the statdata struct is used in multiple places, not just here, so I felt having it explained here in its entirety would be over documentation

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: yeah this was left here initially since I didnt dive too deep into functions. Will state that this function doesnt panic

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: yeah I had considered this exact thing, but I thought being explicit about everything instead of a slightly vague statement like "same as stat except in xyz" wouldn't really be helpful for end users. Plus this would then have to be replicated everywhere there is commonality

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: I checked, and there is no particular reason for the dummy info being these values, they could be anything, but probably this is something that was taken from the live environments in the past.

src/tests/fs_tests.rs
pranav-bhatt: ENOENT is a positive value, whereas the errors returned are negative. I thought instead of typecasting and negating the result, it would be clearer to understand this way

src/safeposix/syscalls/fs_calls.rs
Anway-Agte: I was thinking in the context of this function

src/safeposix/syscalls/fs_calls.rs
Anway-Agte: i meant in addition to what you have written

src/tests/fs_tests.rs
Anway-Agte: Most test cases do it the negative way, I thought it'd be uniform in how we write test cases

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1242,18 +1290,55 @@ impl Cage {
...
+    /// * If the file descriptor passed is invalid (less than zero or greater
+    ///   than MAX FD which is 1024)
+    /// * If the inode number retrieved from the file descriptor does not exist
+    ///   in `FS_METADATA.inodetable`.
+    ///
+    /// For more detailed description of all the commands and return values,
+    /// refer to the stat man page [here](https://man7.org/linux/man-pages/man2/stat.2.html).
 
     pub fn fstat_syscall(&self, fd: i32, statbuf: &mut StatData) -> i32 {
+        // Attempt to get the file descriptor
```
davidge20, line 1327: we've been adding 'BUG' as a comment here for the reason you mention in the panics section. Adding it here too could help with fixing that later on.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1286,14 +1372,17 @@ impl Cage {
                         "we don't support fstat on sockets yet",
                     );
                 }
+                // Streams don't have inodes, so we'll populate statbuf with dummy info
                 Stream(_) => {
                     self._stat_alt_helper(statbuf, STREAMINODE);
                 }
+                // Pipes don't have inodes, so we'll populate statbuf with dummy info
```
davidge20, line 1383: What does the dummy info signify here?

src/tests/fs_tests.rs
```diff
@@ -4036,4 +4037,136 @@ pub mod fs_tests {
...
+        assert_eq!(cage.bind_syscall(socketfd, &socket), 0);
+
+        // fstat_syscall test here
+        let socketfile_path_fd = cage.open_syscall(socketfile_path, O_RDONLY, S_IRWXA);
+        assert!(socketfile_path_fd > 0);
+        assert_eq!(cage.fstat_syscall(socketfile_path_fd, &mut statdata), 0);
+        assert_eq!(cage.close_syscall(socketfile_path_fd), 0);
+
+        // socket teardown
+        assert_eq!(cage.close_syscall(socketfd), 0);
```
davidge20, line 4529: is the cleanup necessary here or does lindrustfinalize take care of that?

davidge20 (commented):
overall great work, left some minor comments

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: there is no particular reasoning behind using this dummy info, i believe nick mentioned in a previous call that this info was pulled straight from a live environment.

src/tests/fs_tests.rs
pranav-bhatt: AFAIK `lindrustfinalize` just deals with 'domain' socket inode, but im not sure what that means though, so i thought there was no harm handling this part myself

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: ah i see. From what I can tell there aren't really any special ranges that the statbuf can take, but do correct me if im wrong

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: got it, thanks for pointing this out

Anway-Agte (approved):

davidge20 (approved):
LGTM !

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1242,18 +1289,63 @@ impl Cage {
...
+    /// For more detailed description of all the commands and return values,
+    /// refer to the stat man page [here](https://man7.org/linux/man-pages/man2/stat.2.html).
 
     pub fn fstat_syscall(&self, fd: i32, statbuf: &mut StatData) -> i32 {
-        let checkedfd = self.get_filedescriptor(fd).unwrap();
+        // Attempt to get the file descriptor
+        let checkedfd_result = self.get_filedescriptor(fd);
+
+        // If the file descriptor is invalid, return an error of code 9 (bad file num).
+        if let Err(_) = checkedfd_result {
```
yashaswi2000: this differs with the way how our current syscalls work, they usually panic with unwrap.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1286,14 +1379,17 @@ impl Cage {
                         "we don't support fstat on sockets yet",
```
yashaswi2000: can we fail faster here

src/interface/types.rs
```diff
@@ -30,7 +30,7 @@ pub struct FSData {
 //redefining the StatData struct in this file so that we maintain flow of
 // program derive eq attributes for testing whether the structs equal other
 // statdata structs from stat/fstat
-#[derive(Eq, PartialEq, Default)]
+#[derive(Eq, PartialEq, Default, Debug)]
```
yashaswi2000: are we printing statData anywhere?

yashaswi2000 (commented):
minor comments, almost ready to merge

src/interface/types.rs
pranav-bhatt: i found it handy to be able to print the statdata while writing this up, thought it would be useful to have as an ability. Though, I understand since it isnt being actively printed anywhere, this isnt required. Shall I remove this?

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: I see, I made this change cause @davidge20 [over here](https://github.com/Lind-Project/safeposix-rust/pull/301#discussion_r1688547802) said that this behavior was a bug, and I thought I'll just handle it here itself. Would you like me to keep the panic?

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: not sure how to do that since we only get to know this once we read the fd enum. Would you have any ways to do so?

src/safeposix/syscalls/fs_calls.rs
yashaswi2000: yaa lets have consistent behaviour and then we can fix them together.

src/safeposix/syscalls/fs_calls.rs
yashaswi2000: i meant the order of match block, so that error matches go first.

src/interface/types.rs
yashaswi2000: yaa lets remove that

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: got it

yashaswi2000 (approved):
LGTM

src/tests/fs_tests.rs
```diff
@@ -4394,4 +4409,137 @@ pub mod fs_tests {
...
+
+        let cage = interface::cagetable_getref(1);
+
+        let mut statdata = StatData::default();
+
+        // test whether an invalid fd results in a panic
+        let invalid_fd = 0x10000000; // some dummy high fd which should fail fd check
+
+        // here, we catch the panic spat out by fstat into an Err
+        let invalid_fd_panic = catch_unwind(AssertUnwindSafe(|| {
```
rennergade: I think we'd rather not check this right now then add this extra catch_unwind functionality. Fixing the valid fd bug is next on the list to tackle.

src/safeposix/syscalls/fs_calls.rs
```diff
@@ -1193,6 +1231,9 @@ impl Cage {
         statbuf.st_blocks = 0;
     }
 
+    // helper function to populate information of socket inode object into the
```
rennergade, line 1234: We probably dont need to add this above every helper. Maybe just the top one.

rennergade (changes requested):
Looks very good. Two minor changes that I think would work better.

src/safeposix/syscalls/fs_calls.rs
pranav-bhatt: I thought since '//' doesnt count as a docstring, and plus there is a nice description that you see over each helper when you hover over them, this would be nice. Shall i still go ahead and remove?

src/tests/fs_tests.rs
```diff
@@ -6,9 +6,10 @@ pub mod fs_tests {
     use crate::interface;
     use crate::safeposix::syscalls::fs_calls::*;
     use crate::safeposix::{cage::*, dispatcher::*, filesystem};
-    use libc::c_void;
+    use libc::{c_void, O_DIRECTORY};
     use std::fs::OpenOptions;
     use std::os::unix::fs::PermissionsExt;
+    use std::panic::{catch_unwind, AssertUnwindSafe};
```
rennergade: lets not import this anymore as well

src/safeposix/syscalls/fs_calls.rs
rennergade: sure this is fine as is

rennergade (approved):

PR #307, opened 2024-07-16
getpeername, getsockname, gethostname and getifaddrs update

qianxichen233:
## Description

This PR adds some comments and tests for getpeername_syscall, getsockname_syscall, gethostname_syscall and getifaddrs_syscall. A issue was specified in the test: when calling connect without binding to any address, an new address will be automatically assigned to the socket. But we failed to update the corresponding information in the sockhandle, causing the ip and port returned from getsockname_syscall are both 0.

Fixes # (issue)
1. Add fd range check at the beginning of getpeername_syscall and getsockname_syscall

 

### Type of change

## How Has This Been Tested?

- Test A - `ut_lind_net_getpeername_bad_input`
- Test B - `ut_lind_net_getpeername_inet`
- Test C - `ut_lind_net_getpeername_unix`
- Test D - `ut_lind_net_getpeername_udp`
- Test E - `ut_lind_net_getsockname_bad_input`
- Test F - `ut_lind_net_getsockname_empty`
- Test G - `ut_lind_net_getsockname_inet`
- Test H - `ut_lind_net_getsockname_unix`
- Test I - `ut_lind_net_gethostname`
- Test J - `ut_lind_net_getifaddrs`

## Checklist:

### DIFF
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -3367,12 +3367,12 @@ impl Cage {
     /// ### Description
     /// The `getpeername_syscall()` returns the address of the peer connected to
     /// the socket fd, in the buffer pointed to by ret_addr
-
+    ///
     /// ### Function Arguments
     /// The `getpeername_syscall()` receives two arguments:
     /// * `fd` -  The file descriptor of the socket
     /// * `ret_addr` - A buffer of GenSockaddr type to store the return value
-
+    ///
     /// ### Returns
     /// On success, zero is returned. Otherwise, errors or panics are returned
     /// for different scenarios.
@@ -3439,12 +3439,12 @@ impl Cage {
     /// The `getsockname_syscall()` returns the current address to which the
     /// socket fd is bound, in the buffer pointed to by ret_addr. If the socket
     /// hasn't bound to any address, it returns an empty address.
-
+    ///
     /// ### Function Arguments
     /// The `getsockname_syscall()` receives two arguments:
     /// * `fd` -  The file descriptor of the socket
     /// * `ret_addr` - A buffer of GenSockaddr type to store the return value
-
+    ///
     /// ### Returns
     /// On success, zero is returned. Otherwise, errors or panics are returned
     /// for different scenarios.
@@ -3482,7 +3482,7 @@ impl Cage {
                 if sockhandle.domain == AF_UNIX {
                     // in case of AF_UNIX socket
                     if sockhandle.localaddr == None {
-                        // if hasn't bind to any address
+                        // if hasn't bound to any address,
                         // return an empty address
                         let null_path: &[u8] = &[];
                         *ret_addr = interface::GenSockaddr::Unix(interface::new_sockaddr_unix(
@@ -3501,9 +3501,9 @@ impl Cage {
                         // with both ip and port set to 0. But family should be set since it is
                         // something that was already specified when the socket was created
 
-                        // for ipv4, set the address to 0.0.0.0 if the address is not initialized
-                        // yet for ipv6, set the address to 0:0:0:0:0:0:0:0
-                        // (::) if the address is not initialized yet
+                        // for ipv4, set the address to 0.0.0.0 to indicate uninitialized address
+                        // for ipv6, set the address to 0:0:0:0:0:0:0:0
+                        // (::) to indicate uninitialized address
                         let addr = match sockhandle.domain {
                             AF_INET => interface::GenIpaddr::V4(interface::V4Addr::default()),
                             AF_INET6 => interface::GenIpaddr::V6(interface::V6Addr::default()),
@@ -3512,7 +3512,6 @@ impl Cage {
                             }
                         };
                         ret_addr.set_addr(addr);
-                        // sets port to 0
                         ret_addr.set_port(0);
                         // set the family
                         ret_addr.set_family(sockhandle.domain as u16);
@@ -3546,12 +3545,12 @@ impl Cage {
     /// address_ptr, which has length bytes.  If the null-terminated
     /// hostname is too large to fit, then the name is truncated, and no error
     /// is returned
-
+    ///
     /// ### Function Arguments
     /// The `gethostname_syscall()` receives two arguments:
     /// * `address_ptr` -  The buffer to hold the returned host name
     /// * `length` - The length of the buffer
-
+    ///
     /// ### Returns
     /// On success, zero is returned. Otherwise, errors or panics are returned
     /// for different scenarios.
@@ -3564,8 +3563,8 @@ impl Cage {
     ///
     /// more details at https://www.man7.org/linux/man-pages/man2/gethostname.2.html
     pub fn gethostname_syscall(&self, address_ptr: *mut u8, length: isize) -> i32 {
-        // we only return the default host name because we do not allow for the user to
-        // change the host name right now
+        // we only return the default host name (Lind) because we do not allow for the
+        // user to change the host name right now
         if length < 0 {
             return syscall_error(
                 Errno::EINVAL,
@@ -4373,12 +4372,12 @@ impl Cage {
     /// The `getifaddrs_syscall()` function creates a linked list of structures
     /// describing the network interfaces of the local system, and stores the
     /// address of the first item of the list in buf.
-
+    ///
     /// ### Function Arguments
     /// The `getifaddrs_syscall()` receives two arguments:
     /// * `buf` -  The buffer to hold the returned address
     /// * `count` - The length of the buffer
-
+    ///
     /// ### Returns
     /// On success, zero is returned. Otherwise, errors or panics are returned
     /// for different scenarios.
--- a/src/safeposix/syscalls/net_calls.rs
+++ b/src/safeposix/syscalls/net_calls.rs
@@ -3152,7 +3152,8 @@ impl Cage {
                         "the socket is not connected",
                     );
                 }
-                // we just return the remoteaddr stored in sockhandle
+                // remoteaddr stores the value we want so we just return the remoteaddr stored
+                // in sockhandle
                 *ret_addr = sockhandle.remoteaddr.unwrap();
                 return 0;
             } else {
@@ -3176,7 +3177,8 @@ impl Cage {
     /// ## ------------------GETSOCKNAME SYSCALL------------------
     /// ### Description
     /// The `getsockname_syscall()` returns the current address to which the
-    /// socket fd is bound, in the buffer pointed to by ret_addr.
+    /// socket fd is bound, in the buffer pointed to by ret_addr. If the socket
+    /// hasn't bound to any address, it returns an empty address.
 
     /// ### Function Arguments
     /// The `getsockname_syscall()` receives two arguments:
@@ -3215,6 +3217,8 @@ impl Cage {
                 // get the read lock of socket handler
                 let sock_tmp = sockfdobj.handle.clone();
                 let sockhandle = sock_tmp.read();
+                // each socket type has different structure
+                // so we must handle them seperately
                 if sockhandle.domain == AF_UNIX {
                     // in case of AF_UNIX socket
                     if sockhandle.localaddr == None {
@@ -3233,8 +3237,8 @@ impl Cage {
                 } else {
                     // in case of AF_INET/AF_INET6
                     if sockhandle.localaddr == None {
-                        // for ipv4, sets the address to 0.0.0.0 if the address is not initialized
-                        // yet for ipv6, sets the address to 0:0:0:0:0:0:0:0
+                        // for ipv4, set the address to 0.0.0.0 if the address is not initialized
+                        // yet for ipv6, set the address to 0:0:0:0:0:0:0:0
                         // (::) if the address is not initialized yet
                         // setting the family as well based on the domain
                         let addr = match sockhandle.domain {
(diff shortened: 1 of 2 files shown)


src/safeposix/syscalls/net_calls.rs
```diff
@@ -3103,31 +3103,68 @@ impl Cage {
...
                 let sockhandle = sock_tmp.read();
+                // if the socket is not connected, then we should return an error
                 if sockhandle.remoteaddr == None {
                     return syscall_error(
                         Errno::ENOTCONN,
                         "getpeername",
                         "the socket is not connected",
                     );
                 }
+                // we just return the remoteaddr stored in sockhandle
```
Anway-Agte: A comment saying that remoteaddr stores the value we want would be good for clarification

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3173,70 @@ impl Cage {
         }
     }
 
+    /// ## ------------------GETSOCKNAME SYSCALL------------------
+    /// ### Description
+    /// The `getsockname_syscall()` returns the current address to which the
+    /// socket fd is bound, in the buffer pointed to by ret_addr.
```
Anway-Agte: A sentence explaining what to expect if the socket is not bound would be great

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3173,70 @@ impl Cage {
...
         let unlocked_fd = checkedfd.read();
         if let Some(filedesc_enum) = &*unlocked_fd {
             if let Socket(sockfdobj) = filedesc_enum {
+                // must be a socket file descriptor
+
+                // get the read lock of socket handler
                 let sock_tmp = sockfdobj.handle.clone();
                 let sockhandle = sock_tmp.read();
                 if sockhandle.domain == AF_UNIX {
+                    // in case of AF_UNIX socket
```
Anway-Agte, line 3483: Explaining why we need to do stuff differently for each socket type would be good

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3173,70 @@ impl Cage {
...
-                    //if the socket is not none, then return the socket
+                    // if the socket address is not none, then return the socket address
                     *ret_addr = sockhandle.localaddr.unwrap();
                     return 0;
                 } else {
+                    // in case of AF_INET/AF_INET6
                     if sockhandle.localaddr == None {
-                        //sets the address to 0.0.0.0 if the address is not initialized yet
-                        //setting the family as well based on the domain
+                        // for ipv4, sets the address to 0.0.0.0 if the address is not initialized
```
Anway-Agte: "set" the address to ...

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3173,70 @@ impl Cage {
...
+                        // yet for ipv6, sets the address to 0:0:0:0:0:0:0:0
```
Anway-Agte: "set"

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3173,70 @@ impl Cage {
...
+                        // (::) if the address is not initialized yet
```
Anway-Agte: what does the (::) mean?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3173,70 @@ impl Cage {
...
+                        // setting the family as well based on the domain
```
Anway-Agte: What are we doing by setting the family exactly?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3190,9 +3273,32 @@ impl Cage {
         }
     }
 
-    //we only return the default host name because we do not allow for the user to
-    // change the host name right now
+    /// ## ------------------GETHOSTNAME SYSCALL------------------
+    /// ### Description
+    /// The `gethostname_syscall()` returns the null-terminated hostname in the
+    /// character array name, which has length bytes.  If the null-terminated
```
Anway-Agte: It stores it in the address_ptr right ? Where is the character array name coming from ?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3678,9 +3788,32 @@ impl Cage {
...
+    /// ## ------------------GETIFADDRS SYSCALL------------------
+    /// ### Description
+    /// The `getifaddrs_syscall()` function creates a linked list of structures
+    /// describing the network interfaces of the local system, and stores the
+    /// address of the first item of the list in buf.
+
+    /// ### Function Arguments
+    /// The `getifaddrs_syscall()` receives two arguments:
+    /// * `buf` -  The buffer to hold the returned address
+    /// * `count` - The length of the buffer
```
Anway-Agte, line 4379: Does count mean number of elements in the linked list?

Anway-Agte (changes requested):

src/safeposix/syscalls/net_calls.rs
qianxichen233: :: is the shorthand for ipv6 address of 0:0:0:0:0:0:0:0, looks like it gets cargo formatted to the next line

src/safeposix/syscalls/net_calls.rs
qianxichen233: count refers to the size of the buffer as mentioned in the comment. This is something used very frequently in c that when we pass a pointer to the buffer, we always need to pass the allocated size of the buffer, so it can know the capacity of the buffer and will not overflow it. This is less common in rust as rust provides a Vec type that has most of the common needs built in. However, since this is the syscall argument for getifaddrs, we are just following it.

src/safeposix/syscalls/net_calls.rs
qianxichen233: this is something I copied from man page, but I think it refers to address_ptr. I'll update it later

src/safeposix/syscalls/net_calls.rs
qianxichen233: when the socket hasn't bind to any address, we will return an empty address with both ip and port field set to 0, since we know neither of the value. But family is something we already know as the family is specified when we create the socket using `socket_syscall`. So even the empty address should have the same family as the socket.

namanlalitnyu: Adding this in the form of comments would be good.

namanlalitnyu (commented):
Just one minor comment.

namanlalitnyu (approved):

rupeshkoushik07 (approved):
LGTM!

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3103,31 +3103,69 @@ impl Cage {
         }
     }
 
+    /// ## ------------------GETPEERNAME SYSCALL------------------
+    /// ### Description
+    /// The `getpeername_syscall()` returns the address of the peer connected to
+    /// the socket fd, in the buffer pointed to by ret_addr
+
```
JustinCappos: Does Rustdoc behave in a sane way with blank lines interspersed?  I don't know...

Maybe make the blank lines have /// as well.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3174,76 @@ impl Cage {
...
+                // get the read lock of socket handler
                 let sock_tmp = sockfdobj.handle.clone();
                 let sockhandle = sock_tmp.read();
+                // each socket type has different structure
+                // so we must handle them seperately
                 if sockhandle.domain == AF_UNIX {
+                    // in case of AF_UNIX socket
                     if sockhandle.localaddr == None {
+                        // if hasn't bind to any address
```
JustinCappos:
```suggestion
                        // if hasn't bound to any address,
```

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3136,29 +3174,76 @@ impl Cage {
...
                     if sockhandle.localaddr == None {
-                        //sets the address to 0.0.0.0 if the address is not initialized yet
-                        //setting the family as well based on the domain
+                        // if the socket hasn't bound to any address, we'd return an empty address
+                        // with both ip and port set to 0. But family should be set since it is
+                        // something that was already specified when the socket was created
+
+                        // for ipv4, set the address to 0.0.0.0 if the address is not initialized
+                        // yet for ipv6, set the address to 0:0:0:0:0:0:0:0
+                        // (::) if the address is not initialized yet
```
JustinCappos: You might explain the difference between "not bound" and "not initialized"

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3167,21 +3252,26 @@ impl Cage {
                             }
                         };
                         ret_addr.set_addr(addr);
+                        // sets port to 0
```
JustinCappos: Why have this comment?  Probably should explain why you're doing it or just omit it.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -3190,9 +3280,32 @@ impl Cage {
...
+    ///
+    /// ### Errors
+    /// * EINVAL - length is negative
+    ///
+    /// ### Panics
+    /// No Panic is expected from this syscall.
+    ///
+    /// more details at https://www.man7.org/linux/man-pages/man2/gethostname.2.html
     pub fn gethostname_syscall(&self, address_ptr: *mut u8, length: isize) -> i32 {
+        // we only return the default host name because we do not allow for the user to
```
JustinCappos: What is default?  Why?  Does this make sense?

src/tests/networking_tests.rs
```diff
@@ -617,22 +798,103 @@ pub mod net_tests {
...
+        // let mut retsocket =
+        // interface::GenSockaddr::V6(interface::SockaddrV6::default());
+        // assert_eq!(cage.getsockname_syscall(sockfd, &mut retsocket), 0);
+        // // port should be 0
+        // assert_eq!(retsocket.port(), 0);
+        // // address should be ::
+        // assert_eq!(
+        //     retsocket.addr(),
+        //     interface::GenIpaddr::V6(interface::V6Addr::default())
+        // );
```
JustinCappos: making a separate test with #ignore is preferred.

src/tests/networking_tests.rs
```diff
@@ -641,17 +903,139 @@ pub mod net_tests {
...
+            let mut retsocket = interface::GenSockaddr::V4(interface::SockaddrV4::default());
+            assert_eq!(cage2.getsockname_syscall(clientsockfd1, &mut retsocket), 0);
+            // BUG: when calling connect without binding to any address, an new address will
+            // be automatically assigned to the socket, and its ip address is supposed to be
+            // 127.0.0.1, and port is not supposed to be 0. But we failed this test
+
+            // assert_ne!(retsocket.port(), 0);
+            // assert_eq!(retsocket.addr(), interface::GenIpaddr::V4(interface::V4Addr {
+            //     s_addr: u32::from_ne_bytes([127, 0, 0, 1]),
+            // }));
```
JustinCappos: So, let's separate out this test in as an #ignore test

src/tests/networking_tests.rs
```diff
@@ -641,17 +903,139 @@ pub mod net_tests {
...
+
+            cage2.exit_syscall(EXIT_SUCCESS);
+        });
+
+        assert_eq!(cage.listen_syscall(sockfd, 4), 0);
+        let mut sockgarbage = interface::GenSockaddr::V4(interface::SockaddrV4::default());
+        barrier.wait();
+        let fd = cage.accept_syscall(sockfd as i32, &mut sockgarbage);
+        assert!(fd > 0);
```
JustinCappos: This is a little pedantic, but why > 0?  I think you want fd >= 0, fd !=-1, fd > -1, etc.

src/tests/networking_tests.rs
```diff
@@ -641,17 +903,139 @@ pub mod net_tests {
...
+            let mut retsocket = interface::GenSockaddr::Unix(interface::new_sockaddr_unix(0, &[]));
+            assert_eq!(cage2.getsockname_syscall(clientsockfd1, &mut retsocket), 0);
+            let unixinfo = match retsocket {
+                interface::GenSockaddr::Unix(value) => value,
+                _ => unreachable!(),
+            };
+            assert_eq!(unixinfo.sun_family, AF_UNIX as u16);
+            assert_ne!(unixinfo.sun_path[0], 0);
+
+            cage2.exit_syscall(EXIT_SUCCESS);
```
JustinCappos: could probably benefit from a few sprinkled comments about what is happening

JustinCappos (changes requested):
A few suggestions throughout...

src/tests/networking_tests.rs
qianxichen233: This is something I copied from other test and did not notice. But I actually feel fd > 0 makes some sense, since 0 is supposed to be occupied by stdin, and we did not do things like dup or close before it, so I guess the fd should not be equal to 0 as well?

src/safeposix/syscalls/net_calls.rs
qianxichen233: default hostname is "Lind"
The hostname is not configurable due to us not implementing the privileged syscall sethostname() as we haven't found any common applications that require it

src/tests/networking_tests.rs
JustinCappos: What about fd 1 and 2 though?  They should be occupied too.  Why is fd 0 special?

qianxichen233: yeah, I guess I'll remove this check, it is not the focus of this test after all

ve1nard (approved):
lgtm!

JustinCappos:
@qianxichen233 If this is ready for my review again, please ping me so I know.  (And follow up on slack in a day or so, if I don't respond.)

JustinCappos (approved):
looks good now.  Thanks for the fixes!

Anway-Agte (approved):

Issue #306, opened 2024-07-15
Refactor Testsuite check to be inside lind_kill_from_id


rennergade:
We have to check if were running the test suite so we don't throw common signals such as SIGCHLD on cage exit etc. Right now we do that outside of lind_kill_from_id but it would be better practice to do it inside the function. @qianxichen233 also pointed out there are places were missing this such as generating EPIPE in write etc.

Issue #263, opened 2024-06-13
Unwrapping on an Err() causes a program to panic


ve1nard:
## Description

If a file descriptor provided to `get_filedescriptor()` is out of the allowed range: 0 to 1024(excluded), the method returns `Err()`, and unwrapping on the returned `Err()` value later on causes a program to panic.

### Cases:

Any syscall, like `fcntl` or `ioctl`, that involves file descriptors needs to access a file descriptor table using `get_filedescriptor()`. The pattern is always the same and involves the following line:

`let checkedfd = self.get_filedescriptor(fd).unwrap();`

### Why this behavior?

By definition, an `unwrap` method for the `Result` type should panic when provided with an `Err()` result.

### How is this tested?

Unit tests calling `fcntl` syscall with out-of-range file descriptors, e.g. negative values or values greater than or equal to 1024, failed with *called Result::unwrap() on an Err*.

### References

https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/error-handling.html#unwrapping-explained

Anway-Agte:
Should a match condition be written to handle the Err() value before unwrapping it?

PR #314, opened 2024-07-18
Send, sendto, recv, recvfrom syscall comments and tests

davidge20:
## Description

Commented send, sendto, recv, recvfrom syscalls.
Added basic tests to flag errors in send and ran exhaustive tests on unix sockets for sends/receives

Fixes # (issue)

 

### Type of change

## How Has This Been Tested?

- Test A - `ut_lind_net_recvfrom_unix()
- Test B - 'ut_lind_net_send_close_server_socket_early()'
- Test C - 'ut_lind_net_send_no_peer_address()'

## Checklist:


src/safeposix/syscalls/net_calls.rs
```diff
@@ -1242,6 +1242,108 @@ impl Cage {
...
+    ///
+    /// ** EAGAIN - (Internet domain datagram sockets) The socket referred to by
+    ///   sockfd had not previously been bound to an address and, upon
+    ///   attempting to bind it to an ephemeral port, it was determined that all
+    ///   port numbers in the ephemeral port range are currently in use.  See
+    ///   the discussion of /proc/sys/net/ipv4/ip_local_port_range in ip(7).
+    ///
+    /// * EALREADY - Another Fast Open is in progress.
+    ///
+    /// ** EBADF - sockfd is not a valid open file descriptor.
```
qianxichen233: why some of these have two star characters

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1251,19 +1353,28 @@ impl Cage {
...
             match filedesc_enum {
+                //In this case, the file descriptor refers to a socket
                 Socket(ref mut sockfdobj) => {
+                    //Grab a write guard to the socket handle
                     let sock_tmp = sockfdobj.handle.clone();
                     let mut sockhandle = sock_tmp.write();
 
-                    // check if this is a domain socket
+                    //If the socket's domain is UNIX, return with error as UNIX
+                    //sockets are connection based.
```
qianxichen233: I think AF_UNIX socket could be udp (connectionless)

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1272,21 +1383,19 @@ impl Cage {
...
+                    //The destaddr's address family must match that of the
+                    //socket's address from 'fd'. Otherwise, a message can't be sent
                     if dest_addr.get_family() != sockhandle.domain as u16 {
                         return syscall_error(
                             Errno::EINVAL,
                             "sendto",
                             "An address with an invalid family for the given domain was specified",
                         );
                     }
-                    if (flags & !MSG_NOSIGNAL) != 0 {
```
qianxichen233, line 1282: why this is removed

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1399,23 +1639,51 @@ impl Cage {
...
+                                        //sendpipe is available in unix socket info
+                                        //it is needed to send the message found in buf
+                                        //to the connected socket
                                         Some(sendpipe) => {
                                             sendpipe.write_to_pipe(buf, buflen, nonblocking) as i32
                                         }
+                                        //sendpipe is not available in unix socket info
+                                        //transmission of data is not possible so return with error
                                         None => {
                                             return syscall_error(Errno::EAGAIN, "writev", "there is no data available right now, try again later");
```
qianxichen233: this error looks weird to me. EAGAIN usually relates to nonblocking flag, but here it has nothing to do with nonblocking. And furthermore, in which situation would we reach here? Would it be expected to reach here?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1399,23 +1639,51 @@ impl Cage {
...
+                                        // The default action for SIGPIPE is to terminate the
+                                        // process without a core dump. This simplifies error
+                                        // handling in programs that
+                                        // are meant to run as part of a shell pipeline: reading
+                                        // input, transforming it, and then writing it to another
+                                        // process. SIGPIPE allows the program to skip error
+                                        // handling and blindly write data until it’s killed
+                                        //
+                                        // Trigger SIGPIPE
                                         interface::lind_kill_from_id(self.cageid, SIGPIPE);
```
qianxichen233, line 1683: add a bug tag here due to this [issue](https://github.com/Lind-Project/safeposix-rust/issues/306)

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1425,69 +1693,103 @@ impl Cage {
...
+                                                "The descriptor is not connected",
                                             );
                                         }
-                                        Err(()) => panic!(
-                                            "Unknown errno value from socket sendto returned!"
-                                        ),
                                     };
-                                } else {
-                                    return retval;
+                                    //** Why is it necessary to drop here?? */
```
qianxichen233: in sendto_syscall, we need to acquire the fd/sockhandle with write/read lock again. If we do not release the lock here, deadlock would happen

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1571,8 +1887,13 @@ impl Cage {
...
     ) -> i32 {
-        // maybe select reported a INPROGRESS tcp socket as readable, so re-check the
-        // state here
+        //In the case that the socket is nonblocking and the connection can not
+        //be completed immediately, the connection state of the socket will
+        //be set to INPROGRESS.
+        //It is possible that select_syscall or poll_syscall had reported
+        //the INPROGRESS TCP socket as readable. If so, we can adjust the
+        //state of the socket to CONNECTED.
+        //** In what case would this happen?? */
```
qianxichen233: INPROGRESS state may happen when you try to connect with nonblocking flag, which I think is a half-broken feature right now. Logics related to INPROGRESS state in select is also a mess currently

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1597,50 +1921,81 @@ impl Cage {
...
             loop {
+                //Grab the receive pipe from the socket to read the data
+                //into the remaining space in the buffer **
                 let sockinfo = &sockhandle.unix_info.as_ref().unwrap();
                 let receivepipe = sockinfo.receivepipe.as_ref().unwrap();
                 retval = receivepipe.read_from_pipe(bufleft, buflenleft, nonblocking) as i32;
+                //In the case of an error from reading from the receive pipe
                 if retval < 0 {
+                    //** This step seems weird
+                    // Even if the call fails we return with the previous peeked data? */
```
qianxichen233: The idea of recv is that as long as it already reads something, it can return. The read length does not need to be the same as the passed buffer length, and that's the meaning of the recv return value. But if we did not read anything, we must block until something gets read

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1272,21 +1383,19 @@ impl Cage {
...
                     }
-                    if (flags & !MSG_NOSIGNAL) != 0 {
-                        return syscall_error(
-                            Errno::EOPNOTSUPP,
-                            "sendto",
-                            "The flags are not understood!",
-                        );
-                    }
 
+                    //If sendto_syscall is used on a connection-mode socket, then
```
ve1nard, line 1396: how is this check different from the one below it?

src/safeposix/syscalls/net_calls.rs
davidge20: ** Indicates the error may be returned from RustPOSIX

src/safeposix/syscalls/net_calls.rs
davidge20: Yeah, I should have specified that in my discussion with nick he explained that it happens so rarely and that there aren't really any use cases yet so we don't implement it in lind

src/safeposix/syscalls/net_calls.rs
davidge20: Nick mentioned that at one point it was needed, but that is not the case anymore after we started using pipes

src/safeposix/syscalls/net_calls.rs
davidge20: Good catch. Given we only reach there if the sendpipe is not available in unix socket info, it seems like as long as all goes well in connect/accept we don't expect to reach there. Seems like a safety net. 

Switching the error to ENOTCONN and switching the message seems to be more accurate.

ve1nard (commented):
one small question

src/safeposix/syscalls/net_calls.rs
davidge20: awesome! thanks

src/safeposix/syscalls/net_calls.rs
qianxichen233: if this is a place we do not expect to reach, I think change this to panic would make more sense

src/safeposix/syscalls/net_calls.rs
davidge20: Good catch. -update on 7/23 -> UDP sockets may be in a CONNECTED state as well

ve1nard (approved):
lgtm

src/safeposix/syscalls/net_calls.rs
yashaswi2000: should probably mention it somewhere

src/safeposix/syscalls/net_calls.rs
yashaswi2000: i am confused, if its a connection mode socket, and destaddress is null, we wouldnt reach this line in execusion right?

src/safeposix/syscalls/net_calls.rs
yashaswi2000: +1 to this, we should panic in such a case

src/safeposix/syscalls/net_calls.rs
yashaswi2000: +1

yashaswi2000 (commented):
added some questions for clarification

src/safeposix/syscalls/net_calls.rs
davidge20: If the socket is in some CONNECTED mode, such as CONNECTED, CONNWRONLY, CONNRDONLY, then it continues, ignoring the destaddress that is null

qianxichen233 (approved):

yashaswi2000 (approved):
LGTM

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1242,6 +1242,108 @@ impl Cage {
...
+    ///   Write permission is denied on the destination socket file, or search
+    ///   permission is denied for one of the directories the path prefix.  (See
+    ///   path_resolution(7).)  (For UDP sockets) An attempt was made to send to
+    ///   a network/broadcast address as though it was a unicast address.
+    ///
+    /// ** EAGAIN or EWOULDBLOCK - The socket is marked nonblocking and the
+    ///   requested operation would block.  POSIX.1-2001 allows either error to
+    ///   be returned for this case, and does not require these constants to
+    ///   have the same value, so a portable application should check for both
+    ///   possibilities.
```
JustinCappos: But what do we do?   This is what a reader may want to know.

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1242,6 +1242,108 @@ impl Cage {
...
+    /// * EMSGSIZE - The socket type requires that message be sent atomically,
+    ///   and the size of the message to be sent made this impossible.
+    ///
+    /// * ENOBUFS - The output queue for a network interface was full.  This
+    ///   generally indicates that the interface has stopped sending, but may be
+    ///   caused by transient congestion. (Normally, this does not occur in
+    ///   Linux.  Packets are just silently dropped when a device queue
+    ///   overflows.)
+    ///
+    /// * ENOMEM - No memory available.
```
JustinCappos, line 1323: I don't think we return many of these in our code...

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1358,39 +1490,149 @@ impl Cage {
...
+    ///
+    /// ** EAGAIN - (Internet domain datagram sockets) The socket referred to by
+    ///   sockfd had not previously been bound to an address and, upon
+    ///   attempting to bind it to an ephemeral port, it was determined that all
+    ///   port numbers in the ephemeral port range are currently in use.  See
+    ///   the discussion of /proc/sys/net/ipv4/ip_local_port_range in ip(7).
+    ///
+    /// * EALREADY - Another Fast Open is in progress.
+    ///
+    /// ** EBADF - sockfd is not a valid open file descriptor.
```
JustinCappos: Why the difference in * vs **?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1358,39 +1490,149 @@ impl Cage {
...
+    ///
+    /// * invalid or out-of-bounds file descriptor, calling unwrap() on it will
+    ///   cause a panic.
+    /// * Unknown errno value from sendto returned, will cause panic.
+    ///
+    /// for more detailed description of all the commands and return values, see
+    /// [send(2)](https://linux.die.net/man/2/send)
     pub fn send_syscall(&self, fd: i32, buf: *const u8, buflen: usize, flags: i32) -> i32 {
+        //BUG:
+        //If fd is out of range of [0,MAXFD], process will panic
```
JustinCappos, line 1595: This is not good.  Should return an error instead of panicing

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1358,39 +1490,149 @@ impl Cage {
...
+        //Otherwise, we obtain a write gaurd to the Option<FileDescriptor> object
```
JustinCappos: throughout the document, change gaurd to guard

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1399,23 +1641,52 @@ impl Cage {
...
+                                        //it is needed to send the message found in buf
+                                        //to the connected socket
                                         Some(sendpipe) => {
                                             sendpipe.write_to_pipe(buf, buflen, nonblocking) as i32
                                         }
+                                        //sendpipe is not available in unix socket info
+                                        //transmission of data is not possible so return with error
                                         None => {
-                                            return syscall_error(Errno::EAGAIN, "writev", "there is no data available right now, try again later");
+                                            panic!("sendpipe is not available in unix socket info in send syscall");
```
JustinCappos: Weird to change this.  Why panic?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1715,22 +2087,34 @@ impl Cage {
...
+                        //value, panic
                         Err(()) => panic!("Unknown errno value from socket recvfrom returned!"),
                     };
                 }
-                break; // we're okay to move on
+                break; //upon a successful read from the receive pipe, break
+                       // from the loop
             }
         }
+        //sum the total number of bytes from the last peak plus the additional
```
JustinCappos: peak or peek?

src/safeposix/syscalls/net_calls.rs
```diff
@@ -1869,18 +2432,16 @@ impl Cage {
...
+    /// * invalid or out-of-bounds file descriptor, calling unwrap() on it will
     ///   cause a panic.
     /// * unknown errno value from socket bind sys call from libc in the case
     ///   that the socket isn't assigned an address
     /// * unknown errno value from socket listen sys call from libc
     ///
     /// for more detailed description of all the commands and return values, see
     /// [listen(2)](https://linux.die.net/man/2/listen)
-    //
-    // TODO: We are currently ignoring backlog
```
JustinCappos, line 1881: What does this mean?

src/tests/networking_tests.rs
```diff
@@ -1659,6 +1660,373 @@ pub mod net_tests {
         lindrustfinalize();
     }
 
+    #[test]
+    //UNIX Sockets
+    pub fn ut_lind_net_recvfrom_unix() {
```
JustinCappos, line 2130: Should this be set up as separate tests?

JustinCappos (changes requested):

src/safeposix/syscalls/net_calls.rs
rennergade: [This is a bug throughout](https://github.com/Lind-Project/safeposix-rust/issues/263) and is going to be updated globally. Should have someone fixing this next week.

src/safeposix/syscalls/net_calls.rs
davidge20: Many of the errors still could be returned by the libc called so Nick and I agreed to leave them there but notate the ones that we could return.

src/safeposix/syscalls/net_calls.rs
davidge20: At the bottom of the errors section, I wrote " ** Indicates the error may be returned from RustPOSIX." Maybe moving it to the top would be clearer?

src/safeposix/syscalls/net_calls.rs
davidge20: Looking at it again, I suppose it is possible that send is called on a unix socket that isn't connected, so we wouldn't want to panic. Changing to an error, thanks!

src/safeposix/syscalls/net_calls.rs
davidge20: Before we were ignoring the 'backlog' parameter of the listen_syscall, but now we are passing it to the libc call so the comment can be removed.

src/tests/networking_tests.rs
davidge20: What do you mean?

rennergade:
@JustinCappos can you give David feedback so we can finalize this?

src/safeposix/syscalls/net_calls.rs
JustinCappos: If an error cannot be returned, why is it listed?   Isn't this comment a description of this interface?

src/tests/networking_tests.rs
rennergade: Im fine with this as is.

src/safeposix/syscalls/net_calls.rs
rennergade: via Justin - "If a code can't be returned by the interface, why mention it?  Isn't this documentation describing the RustPOSIX API?  Why explain errors that cannot be returned by it?"

rennergade: Realistically all of these can _possibly_ be returned when were in INET mode because it calls send/recv on a raw socket. So instead of removing these or doing weird things with the * notation I think just making it clear that thats the behavior would be best.

rennergade (approved):

PR #323, opened 2024-08-23
Update write vectored 

rupeshkoushik07:
## Description

- Adds comments and tests to writev_syscall
- Adds stream functionality to writev_syscall
- Update write_vectored_to_pipe 

Fixes # (issue)

 

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

tested using cargo test --lib 

## Checklist:

### DIFF
--- a/src/interface/file.rs
+++ b/src/interface/file.rs
@@ -212,17 +212,14 @@ impl EmulatedFile {
             // checks if the file object (fobj) exists.
             let mut file = f.lock();
             // Seek to the specified offset from the beginning of the file
-            file.seek(SeekFrom::Start(offset as u64))?; // moves the file pointer to the desired starting position (offset) from the
-                                                        // beginning of the file.
+            file.seek(SeekFrom::Start(offset as u64))?; 
+            // moves the file pointer to the desired starting position (offset) from the
+            // beginning of the file.
 
             // Use write_vectored for efficient writing from multiple buffers
-            total_bytes_written = file.write_vectored(bufs)?; //It performs a
-                                                              // vectored write
-                                                              // operation,
-                                                              // which means it
-                                                              // writes data to
-                                                              // the file from
-                                                              // multiple buffers
+            total_bytes_written = file.write_vectored(bufs)?; 
+            //It performs a vectored write operation,which means it
+            // writes data to the file from multiple buffers
         }
         // Update recorded filesize if we've written past the previous filesize
         if offset + total_bytes_written > self.filesize {
--- a/src/interface/file.rs
+++ b/src/interface/file.rs
@@ -206,26 +206,18 @@ impl EmulatedFile {
         bufs: &[IoSlice<'_>],
         offset: usize,
     ) -> std::io::Result<usize> {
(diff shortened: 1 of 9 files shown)


src/interface/file.rs
```diff
@@ -201,6 +201,34 @@ impl EmulatedFile {
...
+        if let Some(f) = &self.fobj {
+            // checks if the file object (fobj) exists.
+            let mut file = f.lock();
+            // Seek to the specified offset from the beginning of the file
+            file.seek(SeekFrom::Start(offset as u64))?; 
+            // moves the file pointer to the desired starting position (offset) from the
+            // beginning of the file.
+
+            // Use write_vectored for efficient writing from multiple buffers
+            total_bytes_written = file.write_vectored(bufs)?;
```
rennergade: You should be able to use "write_vectored_at" here instead of seeking and using write_vectored

src/interface/misc.rs
```diff
@@ -109,6 +111,83 @@ pub fn cagetable_clear() {
...
+
+//This function logs data from a slice to the specified file descriptor.
+// It handles logging to standard output (fd 1) and standard error (fd 2).
+// For other file descriptors, it uses a low-level approach to write the data.
+
+pub fn log_from_slice(fd: i32, data: &[u8]) -> Result<i32, String> {
+    match from_utf8(data) {
+        Ok(s) => {
+            match fd {
+                1 => {
```
rennergade: at this point we only log things to stdout so you shouldn't differentiate here.

src/tests/fs_tests.rs
```diff
@@ -4698,4 +4699,81 @@ pub mod fs_tests {
...
+
+        // Create a pipe
+        let mut pipe_fds = PipeArray::default();
+        assert_eq!(cage.pipe_syscall(&mut pipe_fds), 0);
+        let read_fd = pipe_fds.readfd;
+        let write_fd = pipe_fds.writefd;
+
+        // Prepare the data to be written using an iovec structure
+
+        let data = b"Hello, pipe!";
```
rennergade: Lets make this a bit more complex of an iovec and test some other situations

rennergade (changes requested):
This is mostly great. I think the testing should be a bit more robust. Would like a file writev test added as well.

rennergade (approved):

rennergade:
I think this looks good but its failing tests, I think due to you adding write_vectored_at to lib.rs. I know you have to do that but its causing some issue with the runner. @yashaswi2000 can you also give this a review and look at why the test is failing?

yashaswi2000:
the build itself is failing, @rupeshkoushik07 are you able to run cargo test on your local?
"add `#![feature(unix_file_vectored_at)]` to the crate attributes to enable" so i think you need to add this for the library to use vectored write.

rupeshkoushik07:
@yashaswi2000 Even after adding `#![feature(unix_file_vectored_at)]` to `lib.rs`, I still encounter the same error locally.

yashaswi2000 (approved):
LGTM

